PR #1177, opened 2015-01-28
CMS-4983 Unnamed - Fix presentation of "unnamed" content in wizard

edloidas:
Updated the TabItem's label setter and setting logic.
Updated the styles for the wizard header.


edloidas:
This Pull Request is dependent on the #1175 . So make sure to merge that one first.

PR #1181, opened 2015-01-30
CMS-5006 Modal Window - Add standard cancel button

Kitamo:


Kitamo:
My fault, I forgot to remove that pull request because CMS-5007 already included required changes.
I will manually fix those files.

PR #1190, opened 2015-02-02
CMS-5044 Content Validation - Visualization

edloidas:
Removed the site icon.
Placed the validity icon instead.
Added red text on the TabItem, if content is invalid.


edloidas:
Depends on #1186

PR #1310, opened 2015-03-13
XP-135 Grid - keyboard operations stop working after sortting

Kitamo:
-keys were unbinded for sort modal dialog when closing because it contains treegrid which close handler was triggering keys unbinding after modal dialog close handler


srs:
Still not working. When I press "sort", then ESC to exit sort dialog the key's stop working.

Kitamo:
Alright, will check.

Kitamo:
Just checked again in Chrome and Firefox - works perfectly.
@srs  could you please recheck and tell me if there any specific preconditions?

PR #1318, opened 2015-03-18
XP-148 Impossible to 'Publish' content when it was moved to another loca...

CorbenDallas:
...tion

- Problem was that during move, only parent node's version was updated, which means that children nodes that got moved looked 'unchanged' for application. Adjusted doMove() method to call StoreNodeCommand with updateMetadataOnly flag set to false for each moved node, not only parent one. It results in overwriting version field of each moved node.


runarmyklebust:
See comments in task XP-148

PR #1326, opened 2015-03-22
XP-160 Module Selector - problems with empty fields causeing "" values r...

Kitamo:
...ather than null

-updated string based value types to use null as value instead of ""


runarmyklebust:
Not to be pulled into 5.1, wait for 6.0

srs:
Could you update this pull request to be applied on top of 6.0 branch instead of master?

PR #1376, opened 2015-04-01
XP-159 Browser cache of Content Types is never invalidated

CorbenDallas:
- Added  publish of ContentTypeUpdatedEvent and  ContentTypeDeletedEvent to  ContentTypeRegistryImpl
 - ContentTypeServiceImpl is used to inject the publisher bean


CorbenDallas:
Closed due to different solution applied

PR #1377, opened 2015-04-01
XP-237 Add more unit tests for data package in API

Kitamo:
-added more unit tests for data package 'com.enonic.xp.data' so we cover more than 85% line coverage.


srs:
Please use org.junit.Assert instead of junit.framework.Assert. junit.framework.Assert is deprecated and should not be used.

Kitamo:
Done.
https://github.com/enonic/xp/pull/1388

PR #1380, opened 2015-04-02
XP-239 Add more unit tests for aggregation package in API

edloidas:
Added test for the branch package.


srs:
Please use org.junit.Assert instead of junit.framework.Assert. junit.framework.Assert is deprecated and should not be used.

PR #1392, opened 2015-04-06
XP-159 Browser cache of Content Types is never invalidated - one event approach

CorbenDallas:


CorbenDallas:
Closed due to different solution applied

PR #1463, opened 2015-04-23
XP-362 TinyMCE - Make editor field grow unlimited

CorbenDallas:
- [Fixed] - removed  autoresize_max_height parameter from init function


alansemenov:
Works fine

PR #1579, opened 2015-05-20
XP-475 NewContentDialogue - Add panel that helps user select content …

Kitamo:
…type

-[Added] 'Most Popular block' on top of content types to choose (according to design). New block takes lots of space so set it to be not displayed when height is less than 540px.
-[Updated] Refactored NewContentDialog

### DIFF
--- a/modules/admin-ui/src/main/resources/web/admin/apps/content-manager/js/app/create/MostPopularItem.ts
+++ b/modules/admin-ui/src/main/resources/web/admin/apps/content-manager/js/app/create/MostPopularItem.ts
@@ -1,0 +1,20 @@
+module app.create {
+
+    import ContentTypeSummary = api.schema.content.ContentTypeSummary;
+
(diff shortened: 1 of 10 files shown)


Kitamo:
I like your solution more :)

Kitamo:
Regarding mostPopularBlockMaxItemsAllowedproperty  - I've created it exactly for case when some event will have to reduce(or increase) number of most popular items to show, for example when not enough height (block 2 in your comments). Thus making number of most popular items shown dynamic.

alansemenov:
I see. Well, since we don't need this functionality at the moment, let's remove it for now and add it when/if we need it. For now we'll just keep 2 as a constant.

Kitamo:
Regarding 1) in your comments - not sure it is possible, I've been thinking of putting in initially, but there is <ul> element for content types that allows only <li> inside.

Kitamo:
Done.

PR #1580, opened 2015-05-20
XP-521 TinyMCE - More adjustments to floating toolbar

Kitamo:
-Updated TinyMCE toolbar border and shadow to be same as for step-navigation-panel when scrolling


alansemenov:
```CSS
      .box-shadow(0 3px 6px 0 rgba(150, 150, 150, 0.7));
```

Since we're now using this shadow color (```rgba(150, 150, 150, 0.7)```) in at least two different less classes, I suggest making it a variable in color.less to make sure it's consistent. Call it *@admin-toolbar-shadow*, for example.

Kitamo:
https://github.com/enonic/xp/pull/1581

PR #1585, opened 2015-05-21
XP-523 TinyMCE - Align tooltip design

Kitamo:
Merged design of TinyMCE tooltip and our tooltip:
-Used tooltip from tiny as base for new design
-Removed dropshadow
-Kept white on black
-Kept rounded corners
-Used same font as rest of XP

### DIFF
--- a/modules/admin-ui/src/main/resources/web/admin/common/styles/api/form/inputtype/text/tinymce.less
+++ b/modules/admin-ui/src/main/resources/web/admin/common/styles/api/form/inputtype/text/tinymce.less
@@ -214,3 +214,12 @@
     font-family: tinymce;
   }
 }
+
+.mce-tooltip {
+  .mce-tooltip-inner {
+    font-family: "Open Sans", sans-serif;
+    -webkit-box-shadow: none;
+    -moz-box-shadow: none;
+    box-shadow: none;
+  }
+}
--- a/modules/admin-ui/src/main/resources/web/admin/common/styles/api/ui/tooltip.less
+++ b/modules/admin-ui/src/main/resources/web/admin/common/styles/api/ui/tooltip.less
@@ -1,14 +1,17 @@
 .tooltip {
   .reset();
-  .opacity(1);
+  .opacity(.8);
 
   position: absolute;
   background: @admin-tooltip-bg;
-  padding: 5px !important; // to override bootstraps .tooltip.top etc.
+  padding: 5px 8px 4px 8px !important; // to override bootstraps .tooltip.top etc.
   z-index: @z-index-tooltip;
   font-size: 11px;
-  color: @admin-dark-gray;
+  color: @admin-white;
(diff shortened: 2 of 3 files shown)


alansemenov:
The tooltips in TinyMCE look fine and according to the task, but it doesn't look like styling of other tooltips has changed (see images).
image
image

Kitamo:
The tooltip you mention in previous comment is native browser's tooltip shown when hovering over element with 'title' attribute. There is currently no way to override it's styling, there are only workarounds.

I've tried 2 options:
1) Using our own api.ui.Tooltip.ts and removing 'title' attribute - Setting it to all required elements brings performance overhead as new events to listen appear and calculation logic is used to set tooltip's position.
2) CSS workarounds - we replace 'title' attribute with some custom 'data-title' attribute and set [data-title]:hover:after {}  with it's own styling (see http://stackoverflow.com/questions/2011142/how-to-change-the-style-of-title-attribute-inside-the-anchor-tag), but that brings lots of other issues to consider such as positioning of tooltip. 

So do we need more options or should accept one of those I've mentioned?

alansemenov:
I found one solution on stackoverflow which I think we could easily use. Tried it locally and it seems to work quite nicely.

1. Add this function to content-manager/main.ts:

```Typescript
function initToolTip() {
    var ID = "tooltip", CLS_ON = "tooltip_ON", FOLLOW = true,
        DATA = "_tooltip", OFFSET_X = 0, OFFSET_Y = 20,
        showAt = function (e) {
            var ntop = e.pageY + OFFSET_Y, nleft = e.pageX + OFFSET_X;
            wemjq("#" + ID).html(wemjq(e.target).data(DATA)).css({
                position: "absolute", top: ntop, left: nleft
            }).show();
        };
    wemjq(document).on("mouseenter", "*[title]", function (e) {
        wemjq(this).data(DATA, wemjq(this).attr("title"));
        wemjq(this).removeAttr("title").addClass(CLS_ON);
        wemjq("<div id='" + ID + "' />").appendTo("body");
        showAt(e);
    });
    wemjq(document).on("mouseleave", "." + CLS_ON, function (e) {
        wemjq(this).attr("title", wemjq(this).data(DATA)).removeClass(CLS_ON);
        wemjq("#" + ID).remove();
    });
    if (FOLLOW) { wemjq(document).on("mousemove", "." + CLS_ON, showAt); }
}
```

1. Call this new function from startApplication in the same class

2. In tooltip.less apply styling to the new tooltip element (via id):
```css
.tooltip, **#tooltip** {
...
}
```

3. The only problem I noticed is when tooltip text is not escaped (like, "<Unnamed Text editor>") then the tooltip will be blank. But this should be easy to fix.

alansemenov:
screen shot 2015-05-26 at 13 06 53
screen shot 2015-05-26 at 13 06 23
screen shot 2015-05-26 at 13 05 49

Kitamo:
One more problem that it will work for content manager only.

PR #1588, opened 2015-05-21
XP-522 TinyMCE - Remove gradient from "pressed" paste as text button

Kitamo:
-Updated according to design.
(Had to attach css selector to attribute because after pressing button on toolbar all it's classes are being reset so unable to set custom class to distinguish 'paste' button)


alansemenov:
It's #4997da in his example... @admin-blue is #4294de and @admin-input-blue is #98c9f2, I think it's too light... But we could add a new color constant to color.less, like @admin-toolbar-toggle-button, for example and set it to #4997da.

PR #1597, opened 2015-05-25
XP-522 TinyMCE - Remove gradient from "pressed" paste as text button

Kitamo:
-Updated according to design.
(Had to attach css selector to attribute because after pressing button on toolbar all it's classes are being reset so unable to set custom class to distinguish 'paste' button)


alansemenov:
*If just moving &.mce-active to mce-btn then 'bullet list' and 'numbered list' are also have blue background when pressed.*

That is exactly what we need. Even though the original task was only about the paste button, we must make sure that all toolbar buttons are using the same styling throughout all modes.

PR #1607, opened 2015-05-27
XP-311 Opening content wizard has severely degraded performance since…

edloidas:
… 5.0.1

Made the FormView notification silent.


edloidas:
Read the comment to the task, don't merge it right away.

PR #1626, opened 2015-06-01
XP-544 Focal point - Apply mask when image is being edited

pmi:
- refactored Shader.ts to be applicable to regular Elements
- reused Shader.ts
- moved shader and highlighter styles to common module


pmi:
Have to close this pull and create a new one to update it because master has been merged in 6.0

PR #1627, opened 2015-06-02
XP-615 Publish dialog, incorrect count for common dependants and chil…

sshklianko:
…dren

-[Modified] publish dialog to re-fetch dialog data each time selection item is removed in order to get correct number and lists of children and dependants
-[Modified] some method signatures to avoid .ts compilation warnings


sshklianko:
Changes of this branch were pulled and verified as part of other publish task branch that was built on top of this one.

PR #1689, opened 2015-06-09
XP-703 Load layout descriptors from different location

vbradnitski:


srs:
Will this work? You are mixing app/parts and just parts in two commits. We should always check into app/ directory.

PR #1750, opened 2015-06-18
XP-801 Publishing Wizard - Misc fixes of the dependencies list

sshklianko:
- Made dependencies list to be shown as one per line
- Show full content path instead of display name
- Sort alphabetically by full path
- No empty space between the "Show/hide dependents and child items" link and its contents
- Hide "Show/hide dependents and child items" if there are no dependencies
- Cursor must change to "hand" when hovered over "Show/hide dependents and child items" and "Include child items" so that it's obvious that they are clickable
- Rename "Show/hide dependents and child items" to "Show/hide dependencies and child items to be published"
- Move the main section inside scrollable area - when the list of dependencies is being scroller the main section should scroll as well


alansemenov:
Looking good. Two small things and we're done:

1. Space between the main section and the "Show/hide dependencies" link should be the same as between the main section and "Your changes are ready for publishing". Now it's a bit too tight.

2. When long path is wrapped to second line, wrapped part should have indent so that it's aligned with the text above (see img below)
image

PR #1763, opened 2015-06-23
XP-567 Branch: Remove html_part and introduce index value processing

GlennRicaud:


runarmyklebust:
Maybe this should be merged into XP-838 also? Since the upgrade-script have to be done for this to be safe? Lets discuss

PR #1765, opened 2015-06-23
XP-839 SortContentDialog:  layout broken when dialog with contents  i…

sshklianko:
…s scrollable

- Fixed bugs
- Refactoring was made


alansemenov:
Nice :)

PR #1770, opened 2015-06-24
XP-831 ContentWizardPage: 'Save before close' dialog appears when no …

vbradnitski:
…any data typed and a 'Close' button pressed

### DIFF
--- a/modules/admin-ui/src/main/resources/web/admin/common/js/data/PropertySet.ts
+++ b/modules/admin-ui/src/main/resources/web/admin/common/js/data/PropertySet.ts
@@ -211,6 +211,12 @@ module api.data {
             return array;
         }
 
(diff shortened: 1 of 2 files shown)


vbradnitski:
1. There is how it worked before - we removed properties in first loop. But that is what I changed. We can't remove properties in forEach loop - it will make a wrong behavior and not all properties will be enumerated. 
2. Agree about moving empty check to Property's method.

alansemenov:
Ok, let's skip #2 then

alansemenov:
By the way, maybe it makes sense to add "removeProperties" method (taking array as an argument) to PropertySet class. I'm sure we remove multiple properties in many places in the code.

vbradnitski:
agree, added.

PR #1771, opened 2015-06-24
XP-862 Parse permissions from export

GlennRicaud:


runarmyklebust:
Sry, a misunderstanding here Glenn, some of it where already done in the XP-838. Took some of the pullreq by hand into the XP-838-branch.

PR #1772, opened 2015-06-24
XP-865 Add version-file to system-export

GlennRicaud:


runarmyklebust:
This should also be done in the XP-838 branch.

PR #1773, opened 2015-06-24
XP-863 ContentWizard ,DateTime type: wrong validation message apperar…

sshklianko:
…s, when incorrect value typed

- Added  wrapper class for additional validation messages produced by some inputs.
- Added handling of that additional validations

### DIFF
--- a/modules/admin-ui/src/main/resources/web/admin/common/js/form/AdditionalValidationRecord.ts
+++ b/modules/admin-ui/src/main/resources/web/admin/common/js/form/AdditionalValidationRecord.ts
@@ -1,0 +1,67 @@
+module api.form {
+
+    export class AdditionalValidationRecord {
+
+        private message: string;
+        private overwriteDefault: boolean = false;
+
+        constructor(builder: Builder) {
+            this.message = builder.getMessage();
+            this.overwriteDefault = builder.isOverwriteDefault();
+        }
+
+        public static create(): Builder {
+            return new Builder();
+        }
+
+        getMessage(): string {
+            return this.message;
+        }
+
+        isOverwriteDefault(): boolean {
+            return this.overwriteDefault;
+        }
+
+        equals(that: AdditionalValidationRecord): boolean {
+
+            if (this.message != that.message) {
+                return false;
+            }
+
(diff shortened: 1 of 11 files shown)


alansemenov:
AdditionalValidationRecord.ts
1. I don't think we need/use getMessage and isOverwriteDefault methods in the builder class

Date.ts/DateTime.ts
2. We should initialise dateParsingFailed property with "false" value in the constructor, just like we do for validUserInput
3. I suggest moving code of onSelectedDateChanged/onSelectedDateTimeChanged methods to a new protected method in the base class - call it onValueChanged, pass value and valueType as arguments.
4. Do we really need both validUserInput AND dateParsingFailed? If these two always have opposite values, then we could probably use only validUserInput and call hasValidUserInput() inside getSpecialValidation()
5. In getSpecialValidation we have almost identical code in both classes. I suggest moving this code to the base class as well (and then you can change inputOccurrences back to private, I guess):
````Typescript
        getSpecialValidation(): api.form.AdditionalValidationRecord {
            var result = api.form.AdditionalValidationRecord.create().setOverwriteDefault(false).build();

            this.inputOccurrences.getOccurrenceViews().forEach((occurrenceView: api.form.inputtype.support.InputOccurrenceView) => {
                var picker;

                if (api.ObjectHelper.iFrameSafeInstanceOf(occurrenceView.getInputElement(), api.ui.time.DatePicker)) {
                    picker = <api.ui.time.DatePicker>(occurrenceView.getInputElement());
                }
                else if (api.ObjectHelper.iFrameSafeInstanceOf(occurrenceView.getInputElement(), api.ui.time.DateTimePicker)) {
                    picker = <api.ui.time.DateTimePicker>(occurrenceView.getInputElement())
                }

                if (picker && picker.isDateParsingFailed()) {
                    result = api.form.AdditionalValidationRecord.create().
                        setOverwriteDefault(true).
                        setMessage("Incorrect value entered").
                        build();
                    return;
                }
            });

            return result;
        }
```

alansemenov:
And by the way - don't we need the same type of validation for TimePicker as well?

PR #1801, opened 2015-07-01
XP-897 Image cropping - Fix height resizer

Kitamo:
-Updated viewport to drag bottom border with rest of page


alansemenov:
It's a really small thing but why have deduction and comparison with 0 in isCropAreaSmallerThanZoomSlider if you could simply do 
```return height < this.zoomSliderHeight;```
(it would even reflect the method name) :)

PR #1809, opened 2015-07-01
XP-918 Implement ExportCommand in Java

sshklianko:
- Implemented


sshklianko:
I have to admit that script uses same boolean parameter "importWithIds" as import command - rename to exportWithIds would be logical.

PR #1834, opened 2015-07-06
XP-863 ContentWizard ,DateTime type: wrong validation message apperar…

sshklianko:
…s, when incorrect value typed

- Added  wrapper class for additional validation messages produced by some inputs.
- Added handling of that additional validations


sshklianko:
Note, that time.ts onTimeChanged() callback does not use base class' onValueChanged() method due to different Value object initialization

PR #1849, opened 2015-07-07
XP-955 Publishing Wizard - Instantly display the selected items

sshklianko:
- I made publish dialog to use up to 15 initially selected items to render kind of publish preview. While items are resolved - publish button is disabled and subheader message shows appropriate label.


alansemenov:
1. I don't think we need the limit of 15 items - we already have them loaded and selected, it's dependencies that take time to load. As of now, if there are more than 15 items selected then those after the 15th will simply not be shown. So let's remove the limit and if we experience any issues with performance then we'll register a separate issue for that.
2. "publish-dialog-subheader" can be passed into constructor of this.subheaderMessage

PR #1850, opened 2015-07-07
XP-954 Publishing Wizard - Show all dependencies by default

sshklianko:
- Made dependencies list to be shown by default with appropriate label.
- Removed show/hide toggle


alansemenov:
Use header tag (h6) instead of the label tag - it should have the same styling as "Your changes are ready for publishing"

PR #1852, opened 2015-07-08
XP-954 Publishing Wizard - Show all dependencies by default

sshklianko:
- Made dependencies list to be shown by default with appropriate label
  - Removed hide/show dependencies toggle

### DIFF
--- a/modules/admin-ui/src/main/resources/web/admin/apps/content-manager/js/app/publish/ContentPublishDialog.ts
+++ b/modules/admin-ui/src/main/resources/web/admin/apps/content-manager/js/app/publish/ContentPublishDialog.ts
@@ -33,7 +33,7 @@ module app.publish {
 
         private includeChildItemsCheck: api.ui.Checkbox;
 
-        private subheaderMessage = new api.dom.H6El();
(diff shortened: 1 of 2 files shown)


alansemenov:
1. Set margin-bottom: 5px for the header to increase space between it and the list

2. I now get error "Cannot read property 'getId' of undefined" when I expand "Large tree node 2", select all items and click "Publish"
image

3. Pass css class names to constructors of subheaderMessage and dependenciesLabel

4. I think we have too much code related to dependenciesLabel... For example, why do we need renderDependenciesLabel() method that adds the label to the container if we toggle visibility anyway? We should either hide/show it by adding/removing it from the container, or by using setVisible, but not both.

5. Move the header to dependenciesItemsView. It made sense to have it outside when it was a link, but now it's a header and it should never be visible if the list is not visible (and the other way around). It should the previous issue as well. Basically, try to get rid of as much code related to dependenciesLabel as possible (ideally it shouldn't even be a property of the class).

6. If there are still any references to "dependenciesLabel" after #5, rename "dependenciesLabel" to "dependenciesHeader" everywhere

Please make sure that new PR is mergeable (this one is not)

PR #1857, opened 2015-07-09
XP-951 Image Cropping - Fix zoom behaviour

Kitamo:
-Updated edit mode to zoom image only when inside cropped area
-Implemented fix for FF as scroll  in edit mode was not dispatching from Mask to masked element


alansemenov:
Everything works fine, but codewise we should avoid creating class properties unless it's absolutely needed and in case of browser check it is not. So, change it to
```Typescript
if(this.isBrowserFirefox()) { //scrolling manually ff as dispatch event not working
     this.triggerScroll(event);
}
```

Kitamo:
That was my first version, but I wanted to prevent checking user's browser at each mouse wheel events as it is not going to change occasionally, so should I update it or leave it as it is?

alansemenov:
It's just parsing of string with a regexp, there won't be any performance issues.

Kitamo:
https://github.com/enonic/xp/pull/1863

PR #1861, opened 2015-07-09
XP-1006 Publish Wizard: the checkbox "Include child items " present i…

sshklianko:
…n the dialog when content has no any child

- Made dialog to hide include children checkbox, if no children available.
- If no children available, button-row will have margin-top property reduced


alansemenov:
Array class has neat methods called *some* and *every* for checking if any or all items in the array match a condition. So, in this case it's as simple as this:
```Typescript
        private atLeastOneInitialItemHasChild(): boolean {
            return this.selectedContents.some((obj: ContentSummary) => {
                return obj.hasChildren();
            });
        }
```

PR #1872, opened 2015-07-10
XP-1030 Publish Wizard(modal dialog): dialog-subheader is incorrect, …

sshklianko:
…when all contents are already published


alansemenov:
* Method getSelectedItemsCount no longer seems to be used

* Methods getChildrenEligibleForPublishCount, getDependantsEligibleForPublishCount and getSelectedItemsEligibleForPublishCount are not used on their own, they are only used as a part of calculation of total items to be published. Moreover, each of these methods use the same loop. So, when in countItemsToPublishAndUpdateCounterElements you twice calculate total number of items to be published, you actually loop through the same array 6 times. 
So, instead of these three methods make one method that calculates total number of items to be published, call it "getTotalCountToPublish", have it calculate sum in one loop, and use this new method in countItemsToPublishAndUpdateCounterElements (but use local variable, don't call the method twice).

PR #1873, opened 2015-07-10
XP-1010 when module changed in the ModuleConfigurator, content types …

Kitamo:
…still available from previous module

-updated most popular items list generation to pick content types not from all content types but from types allowed to be created in current place

### DIFF
--- a/modules/admin-ui/src/main/resources/web/admin/apps/content-manager/js/app/create/NewContentDialog.ts
+++ b/modules/admin-ui/src/main/resources/web/admin/apps/content-manager/js/app/create/NewContentDialog.ts
@@ -296,7 +296,7 @@ module app.create {
                          parentSite: Site) => {
 
                     this.listItems = this.createListOfContentTypeItems(contentTypes, parentSite);
-                    this.mostPopularItems = this.createMostPopularItemList(contentTypes, directChilds.getContents());
+                    this.mostPopularItems = this.createMostPopularItemList(this.listItems.map((el) => el.getContentType()), directChilds.getContents());
 
                     this.resetNewContentDialogContent();
                     this.toggleMostPopularBlockShown();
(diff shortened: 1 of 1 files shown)


alansemenov:
```Typescript
        private createMostPopularItemList(allowedContentTypes: ContentTypeSummary[],
                                          directChildContents: api.content.ContentSummary[]): MostPopularItem[] {
            var mostPopularItems: MostPopularItem[] = [],
                filteredList: api.content.ContentSummary[] = directChildContents.filter((content: api.content.ContentSummary) => {
                    return !content.getType().isMedia() && !content.getType().isDescendantOfMedia() && Boolean(this.findElementByFieldValue(allowedContentTypes, "id", content.getType().toString()));
                }),
                aggregatedList: ContentTypeInfo[] = this.getAggregatedItemList(filteredList);

            for (var i = 0; i < aggregatedList.length && i < MostPopularItemsBlock.DEFAULT_MAX_ITEMS; i++) {
                var contentType: ContentTypeSummary = this.findElementByFieldValue(allowedContentTypes, "name", aggregatedList[i].contentType);
                mostPopularItems.push(new MostPopularItem(contentType, aggregatedList[i].count));
            }

            return mostPopularItems;
        }
```

The condition that we use to filter the list is getting unreadable. Since content type is the only thing we check there, let's change the code to make sure that allowedContentTypes argument doesn't include media type and its descendants. Create a new method (getAllowedContentTypes) that you will pass as an argument to createMostPopularItemList where you will generate the list. Also, add a new method (isAllowedContentType) returning true/false where you pass allowedContentTypes and content.
Eventually, filtered list will be generated something like this:
```Typescript
filteredList: api.content.ContentSummary[] = directChildContents.filter((content: api.content.ContentSummary) => {
                    return this.isAllowedContentType(allowedContentTypes, content);
                })
```

Kitamo:
allowedContentTypes doesn't have not allowed content types, it is already have been filtered before;
So I guess I just update:

filteredList: api.content.ContentSummary[] = directChildContents.filter((content: api.content.ContentSummary) => {
                    return this.isAllowedContentType(allowedContentTypes, content);
                })

?

alansemenov:
Sure, as long as the check for isMedia and isDescendantOfMedia is no longer required.

PR #1913, opened 2015-07-23
XP-1026 Replace Resource.from usage by ResourceService.getResource

sshklianko:
- Replaced .from() calls with service calls. Refactored code where necessary.
- Adjusted tests
- See comments


sshklianko:
Following things consider revising:
ScriptServiceImplTest - two tests commented due to I was unable to find source of their failure
ErrorPageBuilder - this onne still uses old Resouce.from() call because I could not find nice way to inject ResourceService into it
For these two things Id prefer to have new tasks created as current task has already taken a lot of time

PR #1916, opened 2015-07-24
XP-1026 Replace Resource.from usage by ResourceService.getResource

sshklianko:
- Replaced .from() calls with service calls. Refactored code where necessary.
- Adjusted tests
- See comments


sshklianko:
Note: ErrorPageBuilder still uses old Resouce.from() call because I could not find nice way to inject ResourceService into it.

PR #1920, opened 2015-07-28
XP-1110 Publish button - Add Publish button to the screen when in edi…

sshklianko:
…t mode in mobile resolution

- Added event for publishing contents
- Added publish button for mobile to ContentWizardPanel


alansemenov:
Stepan, please create a new branch XP-1102 and create a new PR to merge into XP-1102 from XP-1110. Same for other subtasks of XP-1102. When all subtasks of XP-1102 are done, we'll merge into master.

PR #1993, opened 2015-08-12
XP-1213 Cache control headers missing from images in 5.3.0

sshklianko:
- Added cache control header for ImageHandleResource


runarmyklebust:
We will not do this now, this has to be done with an automatic cache invalidation. Postponed to 6.1

PR #2032, opened 2015-08-17
XP-1245 Create buildPropertyTree in ContentService

sshklianko:
- Moved buildPropertyTree method to MixinService and ContentService
- Moved JsonToPropertyTreeTranslator


sshklianko:
Note, that original task description mentioned only ContentService to contain buildPropertyTree method, whereas same method was required for mixin service.

PR #2044, opened 2015-08-18
XP-1245 Create buildPropertyTree in ContentService

sshklianko:
- Moved buildPropertyTree method to MixinService and ContentService
- Moved JsonToPropertyTreeTranslator
- Adjusted tests


srs:
Commented on the issue in youtrack. Please look at it.

PR #2061, opened 2015-08-21
XP-1092 Widget Panel - Add support for detail panel in mobile mode

sshklianko:
- Added separate panel for mobile preview with widgets panel


alansemenov:
* I think it's confusing that detail panel and widgets slide from different sides, so please fix it so that widgets slide from left to right as well.

MobileContentTreeGridActions
* In isPrincipalPresent and hasPermission you don't need to use "result" variable. Method "some" returns boolean by itself, so it's as simple as
```Typescript
        private isPrincipalPresent(principalKey: api.security.PrincipalKey,
                                   accessEntriesToCheck: AccessControlEntry[]): boolean {
            return accessEntriesToCheck.some((entry: AccessControlEntry) => {
                return entry.getPrincipalKey().equals(principalKey);
            });
        }
```

MobileContentBrowseToolbar
* "fold" doesn't have to be a class property (we should try to avoid it unless absolutely necessary). Change initFoldButton to return FoldButton and pass it to "addActions" method. 

```Typescript
constructor(actions: app.browse.action.MobileContentTreeGridActions) {
            super("toolbar");

            var foldButton = this.initFoldButton();

            this.addActions(actions.getAllActions(), foldButton);

            this.initEditButton(actions.EDIT_CONTENT);
        }
````
* and I believe addActions should be private?

PR #2062, opened 2015-08-21
XP-1311 Wrong behavior of deleting of content: when wizard opened and…

vbradnitski:
… there are unsaved changes


alansemenov:
You forgot to remove "debugger" in CloseAction.ts

vbradnitski:
removed.

PR #2087, opened 2015-08-26
XP-1334 ContentWizard, adding new page-template, impossible select a …

edloidas:
…type in the grid, using a checkbox

Fixed the problem with selection update (the input was cleared and grid was updated before the multiselection applied).


alansemenov:
Please extract the code below into a separate new method (`applySelection()`?) and call it in both `selectRowOrApplySelection` and `setupListeners` (or maybe you can reuse `selectRowOrApplySelection` as a listener in `setupListeners`?)

```Typescript
this.clearSelection(true);
this.comboBoxDropdown.applyMultipleSelection();
this.input.setValue("");
this.hideDropdown();
```

edloidas:
Used the `selectRowOrApplySelection` as a handler for the `applySelectionsButton.onClicked` event.

PR #2116, opened 2015-09-01
XP-1394 Page Components view - Wrong positioning of context menu afte…

Kitamo:
…r scrolling


alansemenov:
It's too complicated...  What we need to do is simply pass second argument into select method in PageComponentsView's clickListener. I just tried it locally and it works fine.
```Typescript
treeNode.getData().select();
```
->
```Typescript
treeNode.getData().select(null, api.liveedit.ItemViewContextMenuPosition.TOP);
```

alansemenov:
1. Change  oveflowsBottom as follows:
```Typescript
private oveflowsBottom(y:number, notClicked?: boolean): boolean {
            var yPos = y + this.getEl().getHeight() + this.arrow.getHeight() + 1;
            
            return yPos > (notClicked ? document.body.scrollHeight : (wemjq(window).scrollTop() + window.innerHeight));
        }
````
2. Remove the last two new methods

Kitamo:
Done: https://github.com/enonic/xp/pull/2119

PR #2120, opened 2015-09-01
XP-1379 Image Editor - Change handling for different image sizes

pmi:


alansemenov:
Pavel, can you clean up `debugger` clauses in ImageEditor? There are three of them. Also, debug mode should probably be turned off.

PR #2139, opened 2015-09-04
XP-1484 Move FilesHelper from API to core-image module

Kitamo:


srs:
Please put it beside ImageServiceImpl (same package - com.enonic.xp.core.impl.image). Package com.enonic.xp.util is only used for API so it will not be valid to put it in that package.

PR #2140, opened 2015-09-04
XP-1483 Use ByteSource for result from ImageService.readImage

vbradnitski:


srs:
Could you update this pull request so it's mergable into master?

PR #2144, opened 2015-09-07
XP-1476 Page Component View - Hide main context menu when selecting a…

Kitamo:
… node in the view

- changes directly related to issue are in ItemView.ts
- updated PageComponentView to correctly handle change of selected item view (unexpected scrolling presented before)


alansemenov:
```Typescript

            if(clickPosition) {
                this.showContextMenu(clickPosition, menuPosition);
            }
````
This condition is wrong. As you can see, `clickPosition` argument in `select` method is optional, but we should still show context menu when `menuPosition` is set. This happens, for example, when user adds a new component to the page, or when he selects "Select parent" from the context menu.

PR #2145, opened 2015-09-07
XP-1488 Content Manager - Tuning design of Search Panel

sshklianko:
- Indent result count: did not change indent as currently it looks ok and differs from attached images
 - Indent facet headers: did not change indent as currently it looks ok and differs from attached images
 - Reduce size of facet fonts + add more padding: reduced size of facet headers font from 15px to 14px, did not reduce size of facet font as 12px seem to be correct, added 1px top and bottom padding for facets
 - Make facet lists "grey" when not selected, and black when selected: added gray color for unchecked facets
 - Reduce font size for "clear": reduced size from 16px to 14px
 - Adjust alignment between facet font and checkbox (should be vertically centered): set 2px to top attribute and 1px to left attribute of label:before selector

### DIFF
--- a/modules/admin-ui/src/main/resources/web/admin/common/styles/api/app/browse/browse-filter-panel.less
+++ b/modules/admin-ui/src/main/resources/web/admin/common/styles/api/app/browse/browse-filter-panel.less
@@ -14,6 +14,33 @@
     &:hover {
       color: @admin-button-blue3;
     }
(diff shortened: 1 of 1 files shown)


alansemenov:
The first two items: did you test under mobile resolution as well?

sshklianko:
Yes, I did test for mobile resolution

PR #2147, opened 2015-09-07
XP-1433 Implement cookie support in PortalResponse

sshklianko:


srs:
Please make a branch for parent task XP-187 and make a pull agaist this (not master).

PR #2163, opened 2015-09-09
XP-1477 Page Component View - Open context menu directly from the view

Kitamo:


alansemenov:
*         
```Typescript
if(clickPosition) {
  this.showContextMenu(clickPosition, menuPosition);
}
```
This condition is wrong. As you can see, clickPosition argument in select method is optional, but we should still show context menu when menuPosition is set. This happens, for example, when user adds a new component to the page, or when he selects "Select parent" from the context menu.

* When the burger icon is hovered over, it should change color to white. Same when the burger icon is clicked on (and the menu opens), it should change color to white and stay white as long as the menu is open (even if mouse cursor is no longer over the icon). This can probably be solved by "focused" state of the icon (if it doesn't lose focus when the menu opens).

PR #2164, opened 2015-09-09
XP-1483 Use ByteSource for result from ImageService.readImage

vbradnitski:


srs:
Sorry, could you pull this once more? It's not mergable.

PR #2177, opened 2015-09-11
XP-1519 Admin UI - Improve loading of background image

edloidas:
Added LazyImage to launcher background.


alansemenov:
1. Can we keep "/admin/common/images/background-1920.jpg" in `lazy-image.less`? It's a better approach than having it hardcoded in a typescript class.
2. I don't think we need `className` argument in `LazyImage` class' constructor, I doubt it will ever be used. Plus, to use it now one should remember to add a space in front of a class, otherwise it will be like "lazy-imagemyclass".
3. `setSrc` should be private, I believe

edloidas:
1. Moved background image url from TS to `launcer.less`.
2. My fault. Didn't saw the missing space symbol. Anyhow, I removed it.
3. The `setSrc()` is used from outside, also to set a new image to replace the current.

PR #2187, opened 2015-09-14
XP-1543 Live Edit Toggle button - Replace current "icon" with "screen"

edloidas:
Replaced LIVE button with screen icon.
Fixed cog icon styles.


alansemenov:
Can you fix the PR so that it's mergeable?

PR #2193, opened 2015-09-15
XP-1553 Live Edit - When clicking directly on large components, page …

pmi:
…repositions leaving context menu outside viewport


alansemenov:
Pavel, can you fix it so that it's mergeable?

PR #2237, opened 2015-09-22
XP-1576 Support login using multiple user stores in lib-auth

Kitamo:


Kitamo:
Tests will be failing until XP-1575 merged.

PR #2245, opened 2015-09-23
XP-1540 Detail Panel - blind spot for detail panel at 1920px

sshklianko:
- Added details panel toggle button for large screens
- Made details panel for large screens to slide smoothly
- Removed padding for widget selector

### DIFF
--- a/modules/admin-ui/src/main/resources/web/admin/apps/content-manager/js/app/browse/ContentBrowsePanel.ts
+++ b/modules/admin-ui/src/main/resources/web/admin/apps/content-manager/js/app/browse/ContentBrowsePanel.ts
@@ -20,6 +20,7 @@ module app.browse {
     import ContentId = api.content.ContentId;
     import DetailsPanel = app.view.detail.DetailsPanel;
     import DetailsPanelToggleButton = app.view.detail.DetailsPanelToggleButton;
+    import LargeDetailsPanelToggleButton = app.view.detail.LargeDetailsPanelToggleButton;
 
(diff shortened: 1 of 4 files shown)


alansemenov:
* We need to refactor the ``onClicked`` handler... Something like this:
```Typescript

            this.onClicked((event) => {
                this.detailsPanel.addClass("left-bordered");
                if (this.toggleClass("expanded").hasClass("expanded")) {
                    this.splitPanelWithGridAndDetails.showSecondPanel();
                    this.splitPanelWithGridAndDetails.hideSplitter();
                } else {
                    this.splitPanelWithGridAndDetails.foldSecondPanel();
                }

                setTimeout(() => {
                    this.detailsPanel.removeClass("left-bordered");
                    if (this.hasClass("expanded")) {
                        this.splitPanelWithGridAndDetails.showSplitter();
                    }
                    this.splitPanelWithGridAndDetails.distribute();
                }, 500);
```
* Showing and hiding splitter in the ``onClick`` handler is a bit confusing (when "expanded" class is on): first we call ``showSecondPanel`` method (which calls ``showSplitter`` inside), then we immediately call ``hideSplitter`` and then after 500ms we call ``showSplitter`` again. Can we avoid this? Maybe add "``boolean showSplitter = true``" argument to ``showSecondPanel`` and pass false to it when calling from the ``onClick`` handler?

* Code inside ``hideSecondPanel`` and ``foldSecondPanel`` is almost identical, except for the ``this.secondPanel.hide()`` call. We should either have one method with a boolean argument or call one method from another (``hideSecondPanel`` will call ``foldSecondPanel`` and then ``this.secondPanel.hide()``)

PR #2258, opened 2015-09-24
XP-1634 Enable refreshing of grid in case of event-problems

pmi:
- added refresh icon
- fixed the filter panel icon position to be inline with other icons
- moved grid mask inside the grid to show it when reloading
- also maintain expanded/selected state while reloading grid


alansemenov:
1. Remove ``debugger`` clauses from Grid.ts
2. On hover change color of the refresh button to black, not blue

pmi:
Ok, updated

PR #2280, opened 2015-09-29
XP-1526 Behavior of validation for content in the Wizard and Grid, ar…

Kitamo:
…e different. Red-Icon for invalid content missed in the BrowsePanel

-Introduced value validator that checks if value is null or not when occurences checked.


Kitamo:
Possible problem solution.
NO MERGE UNTIL APPROVED

PR #2289, opened 2015-09-30
XP-1666 Info panel - Item state

vbradnitski:


alansemenov:
* The div with item state should have higher height and have a border/drop-shadow around it, as shown in the image attached to the task

* You use the same (more or less) piece of code in 4 different places in ContentBrowsePanel. Please extract it to a method where you pass contentId and status as arguments.
```Typescript
if (el.getContentId().equals(this.detailsPanel.getItem().getModel().getContentId())) {
  this.detailsPanel.setContentStatus(el.getCompareStatus());
}
```

vbradnitski:
pull has been updated.

PR #2347, opened 2015-10-09
XP-1797 Content grid triggers request for deleted items

pmi:


alansemenov:
Small refactoring:
* Let's replace layoutEmpty and layoutNonEmpty with one method with a boolean parameter, the code is 100% identical (use toggleClass on the last line)
* In ItemStatisticsHeader.setItem the last two blocks should be joined under the same condition and the code inside should be extracted into a separate method:
```Typescript
if (item.getPath()) {
  this.appendToHeader(path, item.getPath(), "parent-path");
  this.appendToHeader(pathName, item.getPathName(), "path-name");
}
```
* Same method:
```Typescript
if (item.getDisplayName()) {
                    this.headerTitleEl.getEl().
                        setInnerHtml(item.getDisplayName(), true).setAttribute("title", item.getDisplayName());
                } else {
                    this.headerTitleEl.getEl().setInnerHtml("").setAttribute("title", "");
                }
````
should be replaced with one-liner (it's ok to pass escapeHtml=true to setInnerHtml for empty value as well).

pmi:
Updated !

PR #2358, opened 2015-10-14
XP-1826 Attachment type for the same attachment differs depending if …

sshklianko:
…you are uploading it from new content dialog or from content wizard

- Made content type to be set during media update the same way as during creation


srs:
Could you please make this pull request against master branch?

sshklianko:
Here is the new one: https://github.com/enonic/xp/pull/2374

PR #2362, opened 2015-10-15
XP-1768 Detail panel - Change transition behaviour

sshklianko:
- Added possibility for second panel of split panel to slide right
- Removed/adjusted some redundant calls for details panels


srs:
Could you please make this pull request against master branch?

sshklianko:
Here is a new one: https://github.com/enonic/xp/pull/2372

PR #2366, opened 2015-10-15
XP-1855 LiveEdit - Error when opening Superhero for edit in XP admin

pmi:
- added check for duplicate jQuery.ready(document) events in main.ts
- fixed NPE in PageModel.getPage
- fixed NPE in ElementHelper.getData
- removed unnecessary setParentElement from PageView


alansemenov:
We cannot use event.target instead of event.originalEvent.target in InsertablesPanel.isOverIFrame - DnD in LiveEdit is not working because of this (when you try to drag a new part to page). Instead we should cast to JQueryEventObject in handleDrag:
```Typescript
var over = this.isOverIFrame(<JQueryEventObject>event);
```

pmi:
I didn't like the idea of casting it back and forth, so I defined args types in draggables declaration and that seem to did the trick for TS.

PR #2375, opened 2015-10-18
XP-1824 Keep content's attachment name while replacing it

sshklianko:


alansemenov:
Change the ``createUploader`` method  to accept ``property`` as the one and only argument and move the check for ``propertyAlreadyHasAttachment`` inside that method. You don't really need ``fileName`` in ``layoutProperty`` for anything else (and you have another variable called ``fileName`` inside ``onFileUploaded`` listener which is confusing).

PR #2401, opened 2015-10-21
XP-1836 IE11 - Make admin open after login

Kitamo:
Was fixed:
1. Alan's fixes mentioned in youtrack;
2. Multiple errors on entering any of apps (CM, users, apps); Cause - ai.app.Application status was not updated correctly on loading, thus AppLauncher.doLoadApplication was triggered twice and iframe src was
attempted to set twice what was properly handled by Chrome and FF but made IE try reload app again. Fix - updated Application show() and hide() functions.
3. Errors in Element.loadExistingChildren(); Cause - HTML's Node.children property is fully supported by FF and Chrome, partially by IE - children property will be returned from Element, but undefined
for SVGelement, Document and DocumentFragment. Fix - Element and ElementHelper functions updated to take into account that Node.children can be undefined 
4. ID generation was broken, all id's were generate like api.dom.Element-X; Cause - ClassHelper.distanceTo() was using non-standard at least until ECMAScript 6 propery 'name'; this property is already supported by 
FF and Chrome but not yet by IE. Fix - ClassHelper already had crossbrowser functions for finding object's name, used those;

Not fixed (not everything ofcourse, just what I've seen resolving issues above):
-Error when opening NewContentDialog
-LiveEdit: try insert part or layout and choose it's content from dropwdown - spinner appears but nothing happens, no errors in console; just seems some event not handled or not throw
-to be continued...


alansemenov:
* Create getChildren method in ElementHelper class which will return child nodes independent on browser:
```Typescript

        public getChildren(): Node[] {

            return this.el.children ||
this.el.childNodes.filter((childNode: Node) => {
  return (childNode.nodeType == Node.ELEMENT_NODE);
});
            
```

* Use this new method everywhere where you need to work with array of child nodes: 
ElementHelper.ts:
```Typescript

        private getChildrenCount(): number {
            return this.getChildren().length;
        }

        private getChild(index: number): Node { 
          return this.getChildren()[index];
        }
```
Element.ts:
```Typescript

        public loadExistingChildren(): Element {

            var children = this.getHTMLElement().getChildren();
                for (var i = 0; i < children.length; i++) {
                    var childAsElement = Element.fromHtmlElement(<HTMLElement>children[i], true, this);
                    this.children.push(childAsElement);
                }
            }
            return this;
        }
```

I'm sure I missed something but you got the idea :)

Kitamo:
Oh yes, I thought I've done this :)

Kitamo:
this.el.childNodes.filter - won't work as NodeList is not array.

Kitamo:
getChildren(): Node[] {

            return this.el.children ||
                   Array.prototype.slice.call(this.el.childNodes).filter((childNode: Node) => {
                       return (childNode.nodeType == Node.ELEMENT_NODE);
                   });
        }

alansemenov:
Yep, neat idea :)

PR #2429, opened 2015-10-26
XP-1919 Implement widget endpoint for portal

sshklianko:
- Added WidgetHandler and WidgetHandlerWorker classes with appropriate tests
- Added RenderMode.ADMIN
- Moved abstract PortalHandlerWorker to Portal-Api
Note, I did not implement XxxUrlParams, XxxUrlFunction and XxxUrlHandler classes as they seem to be out of scope of this task.


srs:
Do not share ControllerHandlerWorker in API. Please just reimplement the code you need in WidgetHandler in admin-impl. ControllerHandlerWorker should be in portal-impl (as it used to).

PR #2434, opened 2015-10-27
XP-1961 LiveEdit panel must be shown by default when opening content …

sshklianko:
…for edit
Added check for existence of page controller that will enable live editor if true

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-manager/js/app/wizard/ContentWizardPanel.ts
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-manager/js/app/wizard/ContentWizardPanel.ts
@@ -273,6 +273,10 @@ module app.wizard {
                         }
                     });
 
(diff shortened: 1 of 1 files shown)


alansemenov:
I suggest we modify contentNotRenderable method and use it to make sure we're consistent with other places in the class where we check that live edit is renderable. I actually think that we should rename the method to "isContentRenderable" and change its body to the opposite, since we more often check that content is renderable than not.
```Typescript

private isContentRenderable(): boolean {
    return this.liveEditModel && this.liveEditModel.getPageModel() && this.liveEditModel.getPageModel().hasController();
}

constructor() {
   ...
   this.onShown((event: api.dom.ElementShownEvent) => {
       ...
       if (this.isContentRenderable()) {
          this.wizardActions.getShowSplitEditAction().execute();
       }
       responsiveItem.update();
   });
   ...
}
```

Please test with different content types after the changes.

PR #2438, opened 2015-10-27
XP-1668 Info panel - User access

vbradnitski:


alansemenov:
UserAccessWidgetItemView:
* ``layoutList()`` is too big and should be refactored. Move generation of ``userAccessList`` array (the ``for`` loop) into a separate method ``getUserAccessList``
* Rename ``unicalKeys`` to ``uniqueKeys``

AccessControlEntryView:
* ``getAccessValueFromEntry`` doesn't need ``else``'s - it's enough with ``return`` in each ``if``
* Assign ace.getAllowedPermissions() to a local variable
* Modify isCanWrite, isCanPublish and isFullAccess methods to call the methods for more limiting permissions, this will make the code much more readable and understandable than it is now. For example, 
```Typescript
private static canWrite(allowed: Permission[]): boolean {
  return this.canRead(allowed) &&
     allowed.indexOf(Permission.CREATE) >= 0 &&
     allowed.indexOf(Permission.MODIFY) >= 0 &&
     allowed.indexOf(Permission.DELETE) >= 0;
}

private static canPublish(allowed: Permission[]): boolean {
  return this.canWrite(allowed) &&
     allowed.indexOf(Permission.PUBLISH) >= 0;
}
```
etc

Checking of ```allowed.length``` can be done either directly in ``getAccessValueFromEntry`` or in a separate method:
```Typescript
private static canOnlyWrite(allowed: Permission[]): boolean {
  return this.canRead(allowed) && allowed.length == 1;
}
```

PR #2471, opened 2015-11-03
XP-1922 Implement createUser in lib-auth

sshklianko:


srs:
You do not need to use @Component. This is used for publishing OSGi services and a ScriptBean is not an OSGi service.

sshklianko:
See https://github.com/enonic/xp/pull/2472
Removed @Component from other handlers as well

PR #2477, opened 2015-11-04
XP-2060 Fix details.less

sshklianko:
- Renamed file to details-panel.less
- Adjusted formatting of details.less entries
- Removed WidgetViewToggleButton.ts as its not used at all


alansemenov:
@sshklianko  can you please fix the PR so that it's mergeable?

PR #2485, opened 2015-11-04
Add note about starting the application

selbekk:
Took me 15 minutes to figure out how to start the application.

This patch adds a quick note about how to start Enonic locally after building it.


srs:
Thanks.

PR #2512, opened 2015-11-09

sshklianko:
- Adjusted DeleteContentCommand to return info for all deleted nodes with status NEW, just like it does for nodes with other statuses.
- Adjusted tests.
- Made delete dialog to fire deleted event for each deleted node which enables closing of wizard editors for those nodes.


sshklianko:
See https://github.com/enonic/xp/pull/2549

PR #2519, opened 2015-11-10
XP-2118 Info panel - Contents of sections are duplicated after publis…

sshklianko:
…hing

- Made content browse panel to trigger details panel update only for active item
- Fixed 'XP-2035Detail panel - Contents disappear after modifying an item' which was not merged into 6.3


alansemenov:
Looks like we repeat the same code three times - any chance of moving the condition inside ```updateDetailsPanel``` method?
```Typescript
var infoPanelItem: ContentSummaryAndCompareStatus = ActiveDetailsPanelsManager.getActiveDetailsPanel().getItem();
if (infoPanelItem && (infoPanelItem.getId() == el.getId())) {
  this.updateDetailsPanel(el);
}
```

sshklianko:
Here is the adjusted one: https://github.com/enonic/xp/pull/2521

PR #2534, opened 2015-11-12
XP-2112 Not possible to delete published content when application is …

sshklianko:
…missing

- During publish of Pending delete item that belongs to an application that was removed - ApplicationNotFoundException occurred during resolving icon url. Added method to IconUrlResolver that returns null if its not possible to resolve content icon.


sshklianko:
See https://github.com/enonic/xp/pull/2543

PR #2542, opened 2015-11-13
XP-2148 Cannot edit links or view source in HtmlAreas in site config.

sshklianko:
- Added additional check to Modal Dialog that checks if tinymce element was clicked while dialog is open
- Added counter for number of open dialogs so that we don't hide mask when any of the open dialogs closed
- Added remove() call to HtmlArea.ModalDialog in order to remove HtmlArea dialogs on close - it creates new dialog each time on open anyway, so we do not pollute dom tree with redundant dialog elements (other option would be to init and use single dialog once for each html area option that requires dialog - see openXxxDialog() methods in HtmlArea)


alansemenov:
Please rename "clickedTinyMceElementThatCanBeIgnored" to something less specific, like "isIgnoredElementClicked"

alansemenov:
Also, please add a simple 1px border for a modal dialog that is following SiteConfigurator dialog so that HTML area dialogs are more discrete. Selector will look something like this:
```CSS
.site-configurator-dialog + .modal-dialog {
  border: 1px solid;
}
```

sshklianko:
Here is the new one: https://github.com/enonic/xp/pull/2545

PR #2549, opened 2015-11-13

sshklianko:
- Added method to ContentBrowsePanel that will fire DeletedEvent for all child nodes of deleted node that were loaded into the grid. As long as we may open editor only for  such nodes - problem is solved.


sshklianko:
See https://github.com/enonic/xp/pull/2564/files

PR #2550, opened 2015-11-13
XP-2141 User Access section - Title with rights of Everyone is not al…

vbradnitski:
…ways shown


alansemenov:
Can you fix this PR so that it's mergeable?

PR #2557, opened 2015-11-17
XP-2162 Wrong behavior in LiveEdit: part inserted with wrong index

vbradnitski:


alansemenov:
The idea is that the style prefix is hardcoded in only two places in the code: the main less file and api.StyleHelper class, so we should avoid hardcoding it anywhere else.
Use
```Typescript
var componentIndex = wemjq(">.drag-helper, >." + api.StyleHelper.getCls("item-view'), regionView.getHTMLElement()).index(ui.item);
```
instead

PR #2564, opened 2015-11-18

sshklianko:
- Added method to ContentBrowsePanel that will fire DeletedEvent for all child nodes of deleted node that were loaded into the grid.
- For cases when node was not loaded into the grid (when you reload browser page with editor tab open and then delete parent of edited content), added additional check for all open tabs to identify ones that are open for non-existent content. That check fires with delay to ensure that deleted nodes that were loaded into grid will trigger tab close before this check.

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-manager/js/app/ContentAppPanel.ts
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-manager/js/app/ContentAppPanel.ts
@@ -16,6 +16,9 @@ module app {
 
         private mask: api.ui.mask.LoadMask;
 
+        private closeRedundantTabsHandler: () => void = api.util.AppHelper.debounce(this.closeTabsWithNonExistentContents.bind(this), 500,
+            false);
+
         constructor(appBar: api.app.bar.AppBar, path?: api.rest.Path) {
 
             super({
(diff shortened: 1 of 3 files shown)


alansemenov:
ContentAppPanel
1. In closeTabsWithNonExistentContents don't make any requests if ids is an empty array (exit the method before new api.content.GetContentSummaryByIds...).
2. Rename closeRedundantTabsHandler to "closeObsoleteTabsHandler"
3. Rename closeTabsWithNonExistentContents to "closeObsoleteTabs"
4. In getOpenEditingTabIds use "filter" method on getNavigationItems(), then map filtered array to the result
5. In closeTabsWithNonExistentContents after you got contentSummaries, call "filter" method on the "ids" array and then forEach:
new api.content.GetContentSummaryByIds(ids).get().then((contentSummaries: ContentSummary[]) => {
var existingIds: string[] = contentSummaries.map((content: ContentSummary) => {
  return content.getId();
});
ids.filter((id: ContentId) => {
  return existingIds.indexOf(id.toString()) == -1; 
}).forEach((id: ContentId) => {
...
}

ContentBrowsePanel.notifyChildrenDeleted
1. I don't understand why you used condition node.isExpanded(), to be honest. It shouldn't matter if the folder is expanded or not. When I delete a folder that is not collapsed, tabs of its children will not be closed.
2. if (node.hasChildren()) is not needed
3. Make getChildren() a local variable of the method and use it in the for loop.

sshklianko:
Here is PR adjusted in accordance with your comments: https://github.com/enonic/xp/pull/2569

PR #2569, opened 2015-11-19

sshklianko:
- Added method to ContentBrowsePanel that will fire DeletedEvent for all child nodes of deleted node that were loaded into the grid.
- For cases when node was not loaded yet into the grid (when you reload browser page with editor tab open), added additional check for all open tabs to identify tabs that are open for non-existent content. That check fires with delay to ensure that deleted nodes that were loaded into grid will triiger tab close before this check.


sshklianko:
Missed Alex' comment...

PR #2653, opened 2015-12-11
XP-2392 FoldButton - Fix positioning and styling of the dropdown

Kitamo:


alansemenov:
One small issue - items inside the dropdown should also have "cursor: pointer", just like items in the toolbar.

And please fix the PR so that it's mergeable.

PR #2677, opened 2015-12-17
XP-2442 Selection not shown for multiple RadioButtons inside item-set

edloidas:
Added unique name for the radio group representation.


alansemenov:
/Users/ase/dev/xp/modules/admin/admin-ui/src/main/resources/web/admin/common/js/dom/ElementRegistry.ts(44,20): error TS2322: Type 'string | number' is not assignable to type 'number'.
  Type 'string' is not assignable to type 'number'.

PR #2730, opened 2016-01-11
XP-2521 Page Components view - Show text from a text component as its…

Kitamo:
… title


alansemenov:
It doesn't work for me. Try this:
1) Open a page for edit
2) Open the Page Components view
3) Right-click a node and add a new text component
4) Change text in this new component

The node still says "Text" even though I changed the text.
image

Kitamo:
Updated

PR #2761, opened 2016-01-19
XP-2622 Modal dialog - Mask doesn't disappear after clicking outside …

Kitamo:
…of the dialog

-Issue origin: OpenEditPermissonsDialog event was triggered second time when pressing enter key after opening dialog due to focus still remained on 'Edit Permissions' button, thus opening handler called twice and two masks added on page.

-Fix: Setting focus on access control combobox or inherit permissions checkbox (same is done for other modal dialogs)


alansemenov:
Need a general fix for the issue that will work for all dialogs. See comments to https://youtrack.enonic.net/issue/XP-2622

PR #2768, opened 2016-01-19
XP-2608 Page Editor - Auto-hide the wizard panel when a new text comp…

Kitamo:
…onent is dragged to the page

-Moved call to minimizing wizard panel to onComponentAdded() because it is called for all ComponentViews, meanwhile onItemViewSelected() is called for all ComponentViews except TextComponentView what brought this issue


alansemenov:
Based on the code, you changed it so that the panel would collapse when a new component is added, and not when it's selected. This is wrong. Everything should work as before - the panel should collapse whenever a component is selected (and that also means whenever a new component is added, because we select the new component after it's added). What we should find you is why onItemViewSelected listener is triggered for all component types except for a text component.

PR #2831, opened 2016-02-08
XP-2664 HTML Editor - Support image scaling when rendering in the portal

pmi:
- created interface FormItemContainer
- added it to Form, FieldSet, FormItemSet
- update wizard and preview on content updated event if it has html area with updated content id inside only
- also set timestamp on every image inside html area each time
- resize html area on shown in case its contents have been updated


alansemenov:
@pmi Can you change this PR to target branch XP-2646? All changes related to image cropping in HTML area are in that branch for now.

pmi:
Retargeting to XP-2646

PR #2920, opened 2016-03-05
XP-2888 Publishing Dialog - Enable opening invalid content items

Kitamo:


alansemenov:
1. No need to repeat "PublishContent", we know that all items in the ContentPublishDialog are items to be published. So rename "isInvalidContentPublishItem" to "isInvalidContent" and "setupClickListenerForPublishItem" to "addOnClickedListener".
2. Fix the grey on-hover background to take 100% of the width

PR #3125, opened 2016-04-11
XP-3124 Open ContentSelector doesn't close when another ContentSelect…

edloidas:
…or is opened

Replaced `click` body event with `focusout`.


alansemenov:
Can you please target the PR to 6.6? Master is only for critical bugfixes until we release 6.5

PR #3131, opened 2016-04-11
XP-3131 SiteConfigurator dilaog should be closed, when 'edit' button …

edloidas:
…for the background pressed and wizard for image opened

Close dialog on image edit request.

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/common/js/content/site/inputtype/siteconfigurator/SiteConfiguratorDialog.ts
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/common/js/content/site/inputtype/siteconfigurator/SiteConfiguratorDialog.ts
@@ -78,9 +78,10 @@ module api.content.site.inputtype.siteconfigurator {
         }
 
         private handleDialogClose(formView: FormView) {
+            let imageSelector;
             formView.getChildren().forEach((element: api.dom.Element) => {
                 if (api.ObjectHelper.iFrameSafeInstanceOf(element, InputView)) {
-                    const imageSelector = (<InputView> element).getInputTypeView().getElement();
+                    imageSelector = (<InputView> element).getInputTypeView().getElement();
                     if (api.ObjectHelper.iFrameSafeInstanceOf(imageSelector, ImageSelector)) {
                         (<ImageSelector> imageSelector).onEditContentRequest(this.close.bind(this));
                     }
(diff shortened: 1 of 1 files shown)


modules/admin/admin-ui/src/main/resources/web/admin/common/js/content/site/inputtype/siteconfigurator/SiteConfiguratorDialog.ts
```diff
@@ -74,6 +77,17 @@ module api.content.site.inputtype.siteconfigurator {
             });
         }
 
+        private handleDialogClose(formView: FormView) {
+            formView.getChildren().forEach((element: api.dom.Element) => {
+                if (api.ObjectHelper.iFrameSafeInstanceOf(element, InputView)) {
+                    const imageSelector = (<InputView> element).getInputTypeView().getElement();
```
alansemenov: @mta I don't see any reason to use "const" instead of "var" or "let" here. Do you?

edloidas:
@alansemenov I do. Since, it's 2016 and ES2016 fully implemented and supported (and TS will transpile `let`/`const` anyway), I belive we should comply with the standards.
We should use either `let` or `const`. I think, `const` should be our default choise.
`const` should be used to make non-reassignable variables, not constant (immutable) values. `imageSelector` won't be reassigned in the future.
You can take a look at the discussion on [StackExchange](http://programmers.stackexchange.com/questions/278652/how-much-should-i-be-using-let-vs-const-in-es6), or read the [article](https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75) by Eric Elliot.

*But* code is a communication, so if others see `const` confusing, I may use `let` instead. Either way, we should decide, when to use `let` or `const` in our team.

alansemenov:
I know about const and I don't have anything against using it but only when it's justified according to its purpose which is to store a fixed value that will be used further in the code. But that is not the case in this specific example. This looks very confusing when we define a new constant on each iteration inside a loop when we in fact need only one variable that should be defined outside of the loop and used as a temporary storage to add an event handler to an element.

```Typescript
private handleDialogClose(formView: FormView) {
    let imageSelector;
    formView.getChildren().forEach((element: api.dom.Element) => {
        if (api.ObjectHelper.iFrameSafeInstanceOf(element, InputView)) {
            imageSelector = (<InputView> element).getInputTypeView().getElement();
            if (api.ObjectHelper.iFrameSafeInstanceOf(imageSelector, ImageSelector)) {
                (<ImageSelector> imageSelector).onEditContentRequest(this.close.bind(this));
            }
        }
    });
}
```

edloidas:
Ok, I get it. Though I thought the problem was the `const` over `let`/`var` in the first place.
As for the defining a varialbe, I used `imageSelector` to optimize the access and make the code shorter, so I placed it inside the context it was used, not hoisting it up, to the top of the main function. Besides, I don't think it will gain us real significant performance boost, if any at all.
But, I must agree, the code looks more organized.

alansemenov:
I only looked at it from the readability point of view. It simply looks confusing when we define a const with the same name on each iteration inside a loop. If we extract all the code inside forEach into a separate method then it's perfectly fine to define imageSelector via const in that method.

edloidas:
I can agree with that. Pushed a separate commit to save the discussion.

PR #3152, opened 2016-04-13
XP-3216 ContentBrowseFilterPanel: Don't trigger refresh event in Cont…

sshklianko:
…entBrowsePanel if Filter Panel not shown


Kitamo:
If you not triggering refresh events on content deleted/created when search panel is closed you should trigger it when it's open.

sshklianko:
There are many places where refresh() is called, if you wanted a panel to be refreshed on show - you should have written it like that in the improvement description

PR #3167, opened 2016-04-27
XP-3233 Content JS Library - Modify - Exception when call modify on a…

GlennRicaud:
…n image


RF0:
How are we supposed to call modify on an image now. With this change I am now for a modify operation getting 

> com.enonic.xp.data.ValueTypeException: Value of type [java.lang.String] cannot be converted to [PropertySet]

var result = contentLib.createMedia()..
contentLib.modify(
                {
                    key: result._id,
                    editor: function (c) {
                        c.data.copyright = "Creative Commons CC0";
                        return c;
                    }
                }
            );

GlennRicaud:
You have to update the variable xpVersion in the gradle.properties of your application.
Because the correction is also in a lib that is included by your application

PR #3170, opened 2016-04-27
XP-3217 Page Editor - Disable anchor links in the root of a part

sshklianko:
- For parts that start with link: enabled pointer-events; set display property to 'block' value in order to enable correct selection and behavior
- In PartComponentView constructor I replaced href attribute value in order to disable navigation
Otherwise, parts with links in the root just don't behave correctly


modules/admin/admin-ui/src/main/resources/web/admin/common/js/liveedit/part/PartComponentView.ts
```diff
@@ -36,6 +38,12 @@ module api.liveedit.part {
             this.parseContentViews(this);
         }
 
+        resetHrefForRootLink(builder: PartComponentViewBuilder) {
```
alansemenov: Shouldn't this be a private method?

PR #3198, opened 2016-05-03
XP-3265 Implement MacroDescriptor in TypeScript

sshklianko:


srs:
Will this actually work? I do not think "import" of files would work in common since it's compiled using the "old"-approach.

PR #3255, opened 2016-05-16
XP-3362 Content status not updated in the wizard, when content was so…

Kitamo:
…rted

-Added sorted handler for content wizard panel


alansemenov:
@Kitamo Can you fix the PR so that it's mergeable?

alansemenov:
Here too, rename "isThisContentWasSorted" to "wasSorted"

Kitamo:
https://github.com/enonic/xp/pull/3267

PR #3258, opened 2016-05-16
XP-3364 Imossible to delete content from the wizard, when content was…

Kitamo:
… moved to another location

-Added moved handler for content wizard panel


alansemenov:
@Kitamo Rename "isThisWasMoved" to "wasMoved" and fix the PR so that it's mergeable

Kitamo:
https://github.com/enonic/xp/pull/3268

PR #3268, opened 2016-05-18
XP-3364 Impossible to delete content from the wizard after it was mov…

Kitamo:
…ed to another location


alansemenov:
Seems like this one became unmergeable again...

Kitamo:
https://github.com/enonic/xp/pull/3279

PR #3289, opened 2016-05-23
XP-3380 Content in Page Editor not updated after adding a new component

vbradnitski:


alansemenov:
``skipValidation`` flag is used only in ContentWizardPanel, so I don't think we should pass it all the way through all the descendant classes. What we should do instead is make ``skipValidation`` a local property of the ContentWizardPanel class and set it when needed, like this:
```Typescript
export class ContentWizardPanel extends api.app.wizard.WizardPanel<Content> {
...
private skipValidation: boolean = false;
...

    saveChanges(skipValidation: boolean = false): wemQ.Promise<Content> {
        this.skipValidation = skipValidation;
        ...
        return super.saveChanges();
    } 

    layoutPersistedItem(persistedContent: Content): wemQ.Promise<void> {
    ...
       if (viewedContent.equals(persistedContent) || this.skipValidation) {
          ...
       }
    ...
    }
    ...
}
```

In the end there should be two classes in the changeset - ContentWizardPanel and LiveFormPanel. Please make sure that PR is mergeable (now it's not).

PR #3342, opened 2016-06-02
XP-3517 Remove text extraction data for some media types

sshklianko:


jsi:
I am merging this to include it in RC-2.  Runar still have to test this.

PR #3347, opened 2016-06-03
XP-3486 Refactor LiveEditModel.initPageModel

sshklianko:
- Extracted initialization code of LiveEditModel into separate class LiveEditModelInitializer
- Refactored initialization methods


sshklianko:
I did not move LiveEditModelInitializer into separate file as its logic is relevant only for LiveEditModel class

PR #3358, opened 2016-06-06
XP-3545 Reset facets before applying filter by dependencies

sshklianko:


alansemenov:
* Let's move the check for ```hasFilterSet``` and ```isDependenciesFilterActive``` inside ```handleEmptyFilterInput```:

```TypeScript
    private searchFacets() {
        this.handleEmptyFilterInput();
        this.searchDataAndHandleResponse(this.createContentQuery());
    }

    private isDependenciesFilterActive(): boolean {
        return this.dependenciesSection.isActive();
    }

    private refreshFacets() {
        this.handleEmptyFilterInput(true);
        this.refreshDataAndHandleResponse(this.createContentQuery());
    }

    private handleEmptyFilterInput(isRefresh?: boolean) {
        if (this.hasFilterSet() || this.isDependenciesFilterActive()) {
            return;
        }       
        if (isRefresh) {

            this.resetFacets(true, true).then(() => {
                new ContentBrowseRefreshEvent().fire();
            }).catch((reason: any) => {
                api.DefaultErrorHandler.handle(reason);
            }).done();

        } else { // it's SearchEvent, usual reset with grid reload
            this.reset();
        }
    }
````
* Either remove method ```isDependenciesFilterActive``` (since the check for ```this.dependenciesSection.isActive()``` will be used only inside ```handleEmptyFilterInput```) or make it private

PR #3360, opened 2016-06-06
XP-3545 Reset facets before applying filter by dependencies

sshklianko:


sshklianko:
I  removed search and remove methods and moved their logic into doSearch and doRefresh so that we dno't need that mislieading return statement

alansemenov:
Nice!

PR #3376, opened 2016-06-07
XP-3536 Content Browse Panel - Enable showing full selection on demand

edloidas:
Added sticky toolbar.
Added selection display limit.


alansemenov:
Merged in 6.7

PR #3393, opened 2016-06-10
XP-3584 Publish toolbar button and Publish menu have wrong state in some cases

edloidas:
Changed the publish enable logic for multiple selection, according to the logic for a single item: disable publish only each and every item has online status.


alansemenov:
@edloidas Can you make this one mergeable, please?

PR #3422, opened 2016-06-16
XP-3230 contentSelector / imageSelector - Bad handling of deleted items

sshklianko:
- Added displayMissingSelectedOptions and removeMissingSelectedOptions to comboBox in order to enable handling of missing contents
- Made Content and Image selectors to remove missing properties on appropriate event
- Added service to ContentResource with appropriate front-end classes that check existence of passed content ids. Service is used by ComboBox.
- Made comboBox to generate event for removal of non-existing content references and render placeholder if content is not available (due to permissions. for example)
- Adjusted xXxSelectedOptionsView classes to generate empty view for missing content's placeholder
- Adjusted image selector toolbar to handle such placeholders
- Made Selectors to notify user via notification popup that missing references were removed

### DIFF
--- a/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/json/content/ContentsExistJson.java
+++ b/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/json/content/ContentsExistJson.java
@@ -1,0 +1,55 @@
+package com.enonic.xp.admin.impl.json.content;
+
+import java.util.List;
+
+import org.codehaus.jparsec.util.Lists;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import com.enonic.xp.content.ContentId;
+
+public class ContentsExistJson
+{
+    private final List<ContentExistJson> contentsExistJson = Lists.arrayList();
+
+    public ContentsExistJson()
+    {
+    }
+
+    public void add( final ContentId contentId, final boolean exists )
+    {
+        contentsExistJson.add( new ContentExistJson( contentId.toString(), exists ) );
+    }
+
+    @SuppressWarnings("UnusedDeclaration")
+    public List<ContentExistJson> getContentsExistJson()
+    {
+        return contentsExistJson;
+    }
+}
+
+class ContentExistJson
+{
+
+    private final String contentId;
+
+    private final boolean exists;
+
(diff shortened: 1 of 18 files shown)


alansemenov:
Great job, @sshklianko! My only suggestion is to simplify the ```selectExistingAndHandleMissing``` method in Combobox.ts as I think it suffers a bit from overkill of checks for existence of options (and lots of calls to .splitValue and looping through arrays). 

See my suggestion below. 

Another option is to build a mixed array of Option objects (both existing and non-existing) and then pass it to a method selectOptions which will simply loop through it and select all options without any checks.

```Typescript
        private selectExistingAndHandleMissing(value: string) {
            var optionIds = this.splitValues(value),
                existingOptionIds = optionIds,
                missingOptionIds = this.getMissingOptionsIds(value),
                nonExistingIds: string[] = [];

            if (missingOptionIds.length == 0) {
                this.selectExistingOptions(existingOptionIds);
                return;
            }

            existingOptionIds = optionIds.filter(function(value){
                return missingOptionIds.indexOf(value) == -1;
            });

            this.selectExistingOptions(existingOptionIds);

            new api.content.ContentsExistRequest(missingOptionIds).sendAndParse().then((result: api.content.ContentsExistResult) => {
                missingOptionIds.forEach((val) => {
                    var contentExists = result.contentExists(val);
                    if (this.displayMissingSelectedOptions && (contentExists || !this.removeMissingSelectedOptions)) {
                        var option = (<BaseSelectedOptionsView<OPTION_DISPLAY_VALUE>> this.selectedOptionsView).makeEmptyOption(val);
                        this.selectOption(option, true);
                    }
                    if (!contentExists) {
                        nonExistingIds.push(val);
                    }
                });

                if (this.removeMissingSelectedOptions) {
                    this.notifyContentMissing(nonExistingIds);
                }
            });

        }

        private selectExistingOptions(values: string[]) {
            values.forEach((val) => {
                var option = this.getOptionByValue(val);
                this.selectOption(option, true);
            });
        }
```

sshklianko:
@alansemenov , see https://github.com/enonic/xp/pull/3428
I adjusted that method in accordance with your recomendation but not completely - in a send() callback I iterate over ALL ids instead of missing ones in order to preserve an order in which items appear in a selector.
Also, I made a separate method selectOptions() that does not check existence of passed ids.

PR #3495, opened 2016-07-11
XP-3660 Faster opening of content from the grid

pmi:
WARNING: It's huge and I'd like it to be reviewed by 2 people at least !

- changed Element.render to be async ( rendered event is thrown after that )
- refactored ALL wizards to inherit as much as possible from ancestors
- refactored ALL wizards to load all necessary data themselves ( or create if needed for new )
- refactored ALL wizards to render forms after data loaded
- refactored ALL app panels to create wizards immediately and let them do the rest


alansemenov:
@pmi Can you fix the PR so that it's mergeable? It's time to merge changes into master.

PR #3496, opened 2016-07-12
XP-3741 Implement alternate password for SU user

vbradnitski:


srs:
Removing this. We already have a fix.

PR #3514, opened 2016-07-20
XP-3775 User App: wizard tabs for User and Group not contains a displ…

sshklianko:
…ay name

- Due to adjustments in rendering and loading resources for wizard panels, principal panel label appeared to be unavailable at the moment wizard tab was rendering, hence, I moved setting of the wizard label value into onRender() handler
- Simplified some code inside


alansemenov:
@sshklianko Methods ```resolveTabMenuItemForUserItemBeingEdited``` and ```resolveTabMenuItemForUserItemBeingViewed``` look 100% identical and I don't think we need both. Remove one and rename the other one to ```resolveTabMenuItem```.

sshklianko:
Done: https://github.com/enonic/xp/pull/3516

PR #3517, opened 2016-07-20
XP-3777 ContentWizard bad styles for the display-name input, when dou…

sshklianko:
…ble click on the content, that already opened

- Made AutoSize input to update it's size only if it's visible - otherwise, regular tab switch causes it to do unnecessary recalculation, which results in incorrect width values set


alansemenov:
@sshklianko Refactor it please to exit the method if the input is not visible instead of including entire body inside the "if...":
```Typescript
private updateSize() {
   if (!this.isVisible()) {
      return;
   }
   ...
}
```

sshklianko:
Done: https://github.com/enonic/xp/pull/3518

PR #3541, opened 2016-08-01
XP-3854 Image selctor content: options in the RichComboBoxComboBox ar…

ashklianko:
…e not clickable

-Setting optionView listeners only when it is rendered otherwise some objects like checkbox or icon are not yet initialized and attempt to add some hanlder on them fails with javascript NPE (ubndefined)


alansemenov:
@ashklianko We need to refactor the code inside ``optionView.onRendered(()...)``:
* extract each listener (onClicked, onKeyDown, onChecked, onLoaded) to a separate method, like it's done with onFocus listener
* create a separate listener method for onRendered where you assign all the abovementioned listeners

ashklianko:
ok

ashklianko:
https://github.com/enonic/xp/pull/3549

PR #3561, opened 2016-08-04
XP-3861 Content Wizard - Masking of panels

pmi:
- updated content wizard to fire 'rendered' event after live edit is fully rendered
- tuned ContentWizardDataLoader to do more parallel requests to boost loading time
- updated 'added' event firing in Element, so that it is fired only if (after) parent is added to DOM
- updated Mask to resize on masked element resize


pmi:
Updated branch with master and created new PR

PR #3649, opened 2016-08-29
XP-3753 When content has been published, name shown in the notificati…

vbradnitski:
…on message, but expexted is display name


modules/admin/admin-ui/src/main/resources/web/admin/common/js/content/form/inputtype/upload/MediaUploader.ts
```diff
@@ -136,7 +136,13 @@ module api.content.form.inputtype.upload {
                     this.uploaderWrapper.addClass("empty");
                     property.setValue(this.newInitialValue());
 
-                    api.notify.showFeedback('\"' + result.getDeleted()[0].getName() + '\" deleted');
+                    if (result.getDeleted() > 0) {
+                        if (result.getDeleted() == 1) {
```
alansemenov: Some refactoring here, maybe?
```Typescript
var itemStr = (result.getDeleted() == 1) ? " item was" : " items were";
api.notify.showSuccess(result.getDeleted() + itemStr + ' deleted');

PR #3651, opened 2016-08-29
XP-4017 Publish button blinks when switching content selection via us…

sshklianko:
…er without publish rights

- Moved reset buttons calls into  SendAndParse() callbacks so that there is no delay between reset and buttons state update that looks to user as blink
- Refactored code in order to simplify it and do less back-end calls
- Made unpublish button to be disabled if user has no publish rights

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/browse/action/ContentTreeGridActions.ts
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/browse/action/ContentTreeGridActions.ts
@@ -117,35 +117,6 @@ export class ContentTreeGridActions implements TreeGridActions<ContentSummaryAnd
         this.UNPUBLISH_CONTENT.setVisible(false);
     }
 
-    private resetDefaultActionsSingleItemSelected(contentBrowseItem: ContentBrowseItem) {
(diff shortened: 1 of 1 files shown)


modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/browse/action/ContentTreeGridActions.ts
```diff
@@ -175,15 +146,17 @@ export class ContentTreeGridActions implements TreeGridActions<ContentSummaryAnd
         this.SHOW_NEW_CONTENT_DIALOG_ACTION.setEnabled(false);
         this.EDIT_CONTENT.setEnabled(this.anyEditable(contentSummaries));
         this.DELETE_CONTENT.setEnabled(this.anyDeletable(contentSummaries));
-        this.DUPLICATE_CONTENT.setEnabled(false);
+        this.DUPLICATE_CONTENT.setEnabled(contentSummaries.length == 1 ? true : false);
```
alansemenov: Small refactoring: it's enough to do ```this.DUPLICATE_CONTENT.setEnabled(contentSummaries.length == 1);``` without the ```"? true : false"``` part.

Same with SORT_CONTENT.

sshklianko:
Here is new PR https://github.com/enonic/xp/pull/3654

PR #3711, opened 2016-09-14
XP-4083 Site Wizard- page-controller selctor still displayed, when co…

ashklianko:
…ntroller has been selected

-Issue origin: after calling saveChanges() it's then() method is supposed to be called before updateHandler() which is triggered by socket update event; Order in which then() in  saveChanges() and updateHandler are triggered is unpredicatable, so disabling updateHandler for ContentWizardPanel when save was triggered until response is received and then() is fulfilled.
-updateHandler() updated to not do anything if event was triggered by ContentWizardPanel save itself


alansemenov:
Retarget this PR to branch 6.8 please

PR #3712, opened 2016-09-14
XP-4026 Importing a lot of content kills content-mananger

vbradnitski:


alansemenov:
This should be targeted at branch 6.8

PR #3722, opened 2016-09-16
XP-3998 Replace three details panels with one

sshklianko:
- Separated slidable panels implementation from details view that contains widget
- Refactored details panel to separate sliding panels functionality from widgets view functionality
- Renamed DetailsPanel into DetailsView- now details view is responsible only for rendering widgets, not hiding/showing itself


alansemenov:
This should be targeted to branch 6.8

sshklianko:
Done. See https://github.com/enonic/xp/pull/3725

PR #3723, opened 2016-09-16
XP-4108 Internet Exploder @Gjensidige enters into compatibility mode …

ashklianko:
…/ crashes on login screen of Enonic XP probably because of missing meta element

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/admin/tools/applications/applications.html
+++ b/modules/admin/admin-ui/src/main/resources/admin/tools/applications/applications.html
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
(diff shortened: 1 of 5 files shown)


RF0:
I have tested distro with this pull request in IE at Gjensidige, and now the pages display properly (although veeery slowly in IE)

One quick note: Best practice is to add the <meta charset="UTF-8" in top of head (see links below) as it may affect performance otherwise (see links below), so maybe switch places on the <meta charset and <meta http-equiv elements?

https://blogs.msdn.microsoft.com/ieinternals/2011/07/18/best-practice-get-your-head-in-order/
https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Introduction_to_HTML5

PR #3748, opened 2016-09-23
XP-4147 Head requests with http-client library fail if Content-Length header is set

GlennRicaud:


modules/lib/lib-http-client/src/test/java/com/enonic/xp/lib/http/HttpRequestHandlerTest.java
```diff
@@ -100,6 +100,20 @@ public void testSimplePostRequest()
     }
 
     @Test
+    public void testSimpleHeadRequest()
+        throws Exception
+    {
+        server.enqueue( addResponse( "GET request" ) );
```
aro: HEAD request?

GlennRicaud: No. Because the content length returned by a HEAD request should be the same as a GET request.

Check mate.

PR #3749, opened 2016-09-26
XP-4153 Make option set to manage multiselection property

sshklianko:
- Added validation of multiselection property
- Made options view to manage selected items property array
- Added disabled checkbox image that is applied to checkboxes when max items are already checked


alansemenov:
@sshklianko I'm pulling this one but please remember to replace the ``multiselection`` node with ``minimum/maximum`` attributes on the ``options`` node.

PR #3752, opened 2016-09-26
XP-3981 Applications App - Applications should be sorted by display name 

edloidas:
Added placeNode method to the ApplicationTreeGrid.
Implemented listKeys (to fetch only application keys) request.


edloidas:
Depends on #3751.

modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/application/ApplicationResource.java
```diff
@@ -128,6 +129,29 @@ public ListApplicationJson list( @QueryParam("query") final String query )
     }
 
     @GET
+    @Path("listKeys")
+    public List<ApplicationKeyJson> listKeys( @QueryParam("query") final String query )
+    {
+        Applications applications = this.applicationService.getInstalledApplications();
+        if ( StringUtils.isNotBlank( query ) )
+        {
```
aro: This block can be extracted in a separate filterApplications method, is already used in 3 places in this class.

modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/application/ApplicationResource.java
```diff
@@ -128,6 +129,29 @@ public ListApplicationJson list( @QueryParam("query") final String query )
     }
 
     @GET
+    @Path("listKeys")
+    public List<ApplicationKeyJson> listKeys( @QueryParam("query") final String query )
```
aro: No need to create the ApplicationKeyJson class for just one string field. Use List<String> instead. The method is named listKeys, so it is not expected to return more details, just the keys as string. And it is anyway untyped since it is sent as json.

PR #3767, opened 2016-09-29
XP-4179 ContentWizard - Status of content not updated after rollback to version with a different status

vbradnitski:


alansemenov:
@runarmyklebust Merge this one after XP-4150

PR #3768, opened 2016-09-29
XP-3982 Site Wizard - option items in the application selector should…

edloidas:
… be sorted by display name

Moved applications sorting to the backend.


edloidas:
@aro Can you verify the changes?

modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/application/ApplicationResource.java
```diff
@@ -97,7 +97,8 @@ public ListApplicationJson list( @QueryParam("query") final String query )
     {
         Applications applications = this.applicationService.getInstalledApplications();
```
aro: If the sorting is done after the filtering there will be less items to sort.

edloidas: I agree. The sources are now updated.

PR #3770, opened 2016-09-30
XP-4204 Response status code set from error.js is ignored

aro:


GlennRicaud:
It does not work for me.

populateStatus( builder, value.getMember( "status" ) );
should be set after
setRedirect( builder, value.getMember( "redirect" ) );

GlennRicaud:
Okay my remark can be discussed another time (a status would override the redirect).
Yep the fix is correct

PR #3791, opened 2016-10-06
XP-4233 PageTemplate-wizard - PageDescriptorDropdown not present on t…

vbradnitski:
…he page

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/common/js/liveedit/LiveEditModel.ts
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/common/js/liveedit/LiveEditModel.ts
@@ -80,13 +80,11 @@ module api.liveedit {
                 hasDefaultPageTemplate = this.pageModel.hasDefaultPageTemplate(),
                 hasApplications = this.siteModel.getApplicationKeys().length > 0;
 
-            var isRenderable = hasApplications || hasController || hasDefaultPageTemplate,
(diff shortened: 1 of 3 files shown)


modules/admin/admin-ui/src/main/resources/web/admin/common/js/liveedit/LiveEditModel.ts
```diff
@@ -80,13 +80,11 @@ module api.liveedit {
...
                 hasApplications = this.siteModel.getApplicationKeys().length > 0;
 
-            var isRenderable = hasApplications || hasController || hasDefaultPageTemplate,
-                isRenderableSite = isSite && isRenderable,
-                isRenderableTemplate = isTemplate && hasController;
-            
-            return isRenderableSite || isRenderableTemplate || hasDefaultPageTemplate;
+            var isRenderable = hasApplications || hasController || hasDefaultPageTemplate;
+
+            return isRenderable && (isTemplate || isSite);
```
alansemenov: With this condition the page editor will not render preview for a page with a template. For example, open any content under Superhero/posts folder.

alansemenov: We should also show the same "Select a controller below to set up a customized page" message above the page descriptor when creating a template as we show when creating a site.

PR #3833, opened 2016-10-18
XP-4280 HTML Area - "Source" button is not available in the inline mode

ashklianko:


alansemenov:
Nothing wrong here, but we need to fix a couple of things.

* If "*source*" tool is explicitly excluded from the toolbar by a user, it will be added by this new code anyway. We need to avoid this. Add a new private property "`customToolConfig`" to HTMLAreaBuilder and assign value to it inside "`setTools`" method. This way we can at any point check whether "*code*" was not explicitly excluded (see below)

```javascript
private isToolExcluded(tool: string): boolean {
  if (!this.customToolConfig || !this.customToolConfig["exclude"]) {
    return false;
  }
  return this.customToolConfig["exclude"].indexOf(tool) > -1;
}
```

* We should avoid being specific about inline mode in the method's name. Create a new private method "`includeTool`" (or "`addTool`") with a string argument. Call this method from includeTools method, and also from `createEditor` like this:

```javascript
if (this.inline && !this.isToolExcluded("code")) {
  this.includeTool("code");
}
```

ashklianko:
Done

PR #3834, opened 2016-10-18
XP-4282 OptionSet validation - Don't validate the form after selecting an option

sshklianko:


sshklianko:
Merge this into 4101 branch  after merging 4152

modules/admin/admin-ui/src/main/resources/web/admin/common/js/form/set/optionset/FormOptionSetOptionView.ts
```diff
@@ -202,15 +207,18 @@ module api.form {
...
                 }
-                this.notifySelectionChanged();
+                this.notifySelectionChanged(this);
             });
 
             var checkboxEnabledStatusHandler: () => void = () => {
-                var canCheckMoreOptions = !button.isChecked() && this.cantSelectMoreOptions();
-                button.setDisabled(canCheckMoreOptions);
-                button.toggleClass("disabled", canCheckMoreOptions);
+                var buttonShouldBeDisabled = !button.isChecked() && this.cantSelectMoreOptions();
```
alansemenov: 1. We need to rename the cantSelectMoreOptions method, current name is confusing. Change it to ``isSelectionLimitReached``.

2. Let's add isDisabled method to the Checkbox class:

        isDisabled(): Boolean {
            return this.checkbox.getEl().isDisabled();
        }

3. Add an optional "cls" argument to "setDisabled" method inside the Checkbox class:

        setDisabled(value: boolean, cls?: string): Checkbox {
            this.checkbox.getEl().setDisabled(value);
            if (cls) {
                this.toggleClass(cls, value)
            }
            return this;
        }

4. Extract the code that enables/disables a checkbox into a separate private method (for example "setCheckBoxDisabled") and call it both on initial rendering and inside the checkboxEnabledStatusHandler. 

```Typescript
private setCheckBoxDisabled(checkBox: api.ui.Checkbox) {
  var checkBoxShouldBeDisabled = !checkBox.isChecked() && this.isSelectionLimitReached();

  if (checkBox.isDisabled() != checkBoxShouldBeDisabled) {
    checkBox.setDisabled(checkBoxShouldBeDisabled, "disabled");
  }
}
```

alansemenov:
Found a bug, too.
1. Create a new content with option set (the one you have in Features)
2. In the multiselect optionset (in addition to already selected Option2) select Option3, and then Option4. When you select Option4,  inputs inside Option3 will be validated but they shouldn't be.

PR #3836, opened 2016-10-19
XP-4143 Page Editor - Close text edit mode on a click outside

ashklianko:
-ItemView's mouse click handler ('handleClick()')'  may be called rather before or after TextComponentView's blur event (non predicatable), so to not process first click on any ItemViews have to introduce static 'isNextClickDisabled' variable that is set to 'true' in onblur handler of TextComponentView; but blur event maybe triggered by clicking on elements different from ItemView, thus using timeout function after setting it to true and setting it to false again after some short period of time to let ItemView's click handler to be called in case blur was triggered by it, in case blur was triggered by some other event let ItemView click

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/common/js/liveedit/ItemView.ts
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/common/js/liveedit/ItemView.ts
@@ -87,6 +87,10 @@ module api.liveedit {
 
         private itemViewIdProducer: ItemViewIdProducer;
 
+        private static isHighlightingAllowed:boolean = true;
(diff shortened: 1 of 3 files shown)


alansemenov:
Let's try to move isHighlightingEnabled from ItemView to PageView (each component always has a page) - it feels more logical to enable/disable highlighting for the entire page rather than making it a static property of ItemView. Maybe you can even get away with not allowing clicks/highlighting if the page is currently in text edit mode and then you won't need any additional properties.

ashklianko:
https://github.com/enonic/xp/pull/3845

PR #3854, opened 2016-10-24
XP-4297 Add option set multiselection validation to back-end

sshklianko:
- Refactored FormItems: made Input FormOptionSet and FormItemSet to extend same parent with common fields
- Refactored from items validation to put all occurrence and option selection validation into one OccurrenceValidation class - so that Form is not iterated a few times. Also, common GenericFormItemParent enabled common interface for testing occurrences.
- Added validation for FormOptionSet
- Slightly adjusted tests. Tests are still to be added for option set


aro:
- Remove GenericFormItem. The new FormOptionSet should just extend from FormItem. Adding another level in the class hierarchy is a risk for introducing new bugs at this point, and might break some calling code from previous versions. Duplicate some code if necessary in FormOptionSet, but keep changes in existing api classes with as less changes as possible.

- Why remove MinimumOccurrencesValidator and MaximumOccurrencesValidator, and move it all to OccurrenceValidator? It's more difficult for me to see what are the changes now. 
I don't understand the method validateFormOptionSet in OccurrenceValidator. And "_selection" is a magic string.

- PropertyArray is used internally in PropertySet, but we should try not to expose it (returning from a method) unless it's really necessary.

- Why is there an empty class FormOptionSetValidatorTest, is this task finished?
- There is also a TODO comment in OccurrenceValidator, why not do it now?

sshklianko:
New PR: https://github.com/enonic/xp/pull/3866

PR #3866, opened 2016-10-25
XP-4297 Add option set multiselection validation to back-end

sshklianko:
- Added validation for FormOptionSet
- Refactored from items validation to put all occurrence and option selection validation into one OccurrenceValidation class - so that Form is not iterated a few times.


aro:
There are no unit tests at all for the new validation logic?
- There are no tests that cover FormOptionSet in OccurrenceValidatorTest
- We need also tests for FormOptionSet in FormDefaultValuesProcessorImplTest.
- Why is there a change to FormOptionSetOptionView.ts in a back-end task?

sshklianko:
@aro , there are quite many tests to be made for option set - we need a separate task for that.
FormOptionSetOptionView.ts fixes a really small problem with one line, so that we don;t need a separate task for that.

sshklianko:
Re-pulled without .ts class: https://github.com/enonic/xp/pull/3870

PR #3890, opened 2016-10-28
XP-4305 Add unit tests for SiteConfigurator class

vbradnitski:


alansemenov:
Very good start, @vbradnitski !

Some issues that need to be fixed:

* Most importantly, the tests should be structured better. Basically now you have just one group (describe) for layout, and then the rest of the specs in the root. We should try to not have any tests in the root at all. The more nested structure of specs we have, the more readable test runs are. Also, some of the specs require layout to be finished - so it would make sense to move them inside the "describe" that tests the layout.

Look at the specs for FormOptionSet - it's a relatively simple class, but it has a proper structuring of specs. 

I would go for something like this:
```
describe('api...SiteConfigurator') {
  describe('constructor') {

    it ('properly creates an instance')
    }

    it ('properly initialises context property')
    }

    it ('properly initialises formContext property')
    }
  }

  describe('what happens during layout') {

    it('layout of the parent class is called') {
    }

    it('siteConfigProvider is initialized) {
    }

    describe('combobox is created') {

      describe('what happens when an option is selected in the combobox') {

        it('selected app must be saved to property set') {
        }
      }

      describe('what happens when an option is moved in the combobox') {

        it('moved app must be saved to property set') {
        }
      }
    }
  }

  describe('public methods') {

        it('getValueType()') {
        }

        it('resetBaseValues()') {
        }
   }

}
```

* Some important tests are missing. For example, you should have a separate group of specs checking if the class itself is created without errors, basically testing the constructor: a) create an instance, test it via .toBeDefined(), b) test that context property is initialised correctly, c) test that formContent property is initialised correctly. Remember that the goal is to have 100% coverage of the class.

* All public methods must be tested (put them into a separate group - except for those that are already tested via other specs in the class)

* Test in "combobox created correctly" doesn't make much sense. You are not really testing that combobox was created - if configurator["comboBox"] is undefined then your test will crash.

There are two ways to test that combobox was properly created, both involve spying on the "createComboBox" method:

1) In beforeEach you create a "fake" combobox and make your spy return this combobox when createComboBox is called. Disadvantage here is that code inside the actual method will never be called, so you cannot use this to test the behaviour of the combobox, but this approach is convenient if you want to test some code that is doing something *with* the combobox:

```
  var comboBox;
        beforeEach((done) => {
            comboBox = new SiteConfiguratorComboBox(...arguments here...);
            spyOn(configurator, "createComboBox").and.returnValue(comboBox);
            spyOn(configurator, "appendChild");
            
            configurator.layout(...);
        });


        it("combobox was added to the view", function (done) {
            expect(configurator.appendChild).toHaveBeenCalledWith(comboBox);
            done();
        });
```

2) You create spy on the same method, but instead of returning a fake combobox, you get the actual one from the spy object. This approach is convenient if you want to test the behaviour of the actual combobox, which has listeners applied to it etc.

```
        var selectedOption, createComboBoxSpy, comboBox;

        beforeEach((done) => {
            createComboBoxSpy = spyOn(configurator, "createComboBox").and.callThrough();

            spyOn(configurator, "appendChild");

            configurator.layout(...);
        });

        it("combobox was created", function (done) {
            expect(createComboBoxSpy).toHaveBeenCalled();
            done();
        });

        it("combobox has correct type", function (done) {
            comboBox = createComboBoxSpy.calls.mostRecent().returnValue;
            expect(api.ClassHelper.getClassName(comboBox)).toEqual("SiteConfiguratorComboBox");
            done();
        });

        it("combobox was added to the view", function (done) {
            comboBox = createComboBoxSpy.calls.mostRecent().returnValue;

            expect(configurator.appendChild).toHaveBeenCalledWith(comboBox);
            done();
        });

        it("combobox was rendered", function (done) {
            comboBox = createComboBoxSpy.calls.mostRecent().returnValue;

            expect(comboBox.rendered).toBeTruthy();
            done();
        });
```

* In general, we should avoid accessing object properties directly via ["propertyName"] - this is unsafe.

* Your specs result in 404 calls to getSiteApplications REST service which should not happen. This can be solved by stubbing `doWhenLoaded` method on RichCombobox to instantly resolve the promise.

alansemenov:
One last thing:
* Don't add specs to the main _spec.ts in common, otherwise we'll end up with hundreds of lines in that file. It should be structured so that _spec.ts file must be present in every folder of the path to the spec file. See how it's done for FormOptionSet specs now.

vbradnitski:
PR has been updated.

vbradnitski:
updated.

alansemenov:
Merged in 6.9

PR #3891, opened 2016-10-28
XP-4271 Unpublish dialog - Add progress bar

edloidas:
Created parent class `ProgressBarDialog` (for publish, unpublish and delete dialogs).
Renamed global class to `is-processing`.
Refactored and renamed `ContentPublishMenuManager` to `ContentPublishMenuButton`.
`ContentPublishMenuButton` now uses inheritance, not composition.
Refactored `ContentPublishDialog` and `ContentUnpublishDialog`.
Updated styles.
Removed unused JSON classes.
Updated backend for the unpublish request/command.


edloidas:
@aro Server unpublish command ( `UnpublishContentCommand` ) is not async. Though the listeners and request are updated, the `ProgressReporter` will track the progress only at the beginning and at the end (`0%` and `100%`). `UnpublishContentCommand` needs further implementation, just like the `PublishContentCommand`.

PR #3899, opened 2016-10-31
XP-4337 Add specs for FormOptionSetView

pmi:
- also fixed NPE in FormSetOccurrenceView.validate() after update() was called


alansemenov:
Slowly but steadily :)
The general comment is that even though we do aim to cover 100% of the class, we should not do this by simply invoking some big method which will call other methods and thus get a full coverage. Our main goal is that by reading the spec runs it should be immediately clear what the class is doing, what main actions of the class are doing under the hood and what cases our tests cover. Specs must be as atomic as possible.

Since FormOptionView is a big class that needs a lot of tests, I suggest you break XP-4337 into subtasks and write specs in batches. It's difficult to both write and review so many tests.

Here are my comments to the PR:

* Your ``describe`` for constructor is not really testing the constructor. It shouldn't have to call {{.layout}} in order to test constructor of the class. What it should test inside the ``describe`` is this:
1) that instance was properly created (use ``expect(instance).toBeDefined()``)
2) that parentDataSet is initialized
3) that formOptionSet is initialized
4) that "even" class is applied (one case)
5) that "odd" class is applied (another case)

* Tests inside ``layout`` and ``validate`` describes are basically testing the same thing. But there are no tests checking the actual code inside ``.layout``. It's fine to have separate group of tests for validation, but then it makes sense to not invoke ``.validate`` method when you are testing layout, so that it doesn't go inside the ``validate``. What we need to test is what happens immediately in the method/action that you are testing. It should be clear what kind of tests we need to write simply by looking at the method's code. This is a working example for the layout method which covers everything that happens in the method:
```
describe('layout()', function () {
            var formOptionSetOccurrencesConstructor, formOptionSetOccurrences;

            beforeEach(function (done) {
                formOptionSetOccurrencesConstructor = api.form.FormOptionSetOccurrences;

                spyOn(api.form, 'FormOptionSetOccurrences').and.callFake(function(config) {
                    formOptionSetOccurrences = new formOptionSetOccurrencesConstructor(config);
                    spyOn(formOptionSetOccurrences, "layout").and.returnValue(wemQ<void>(null));
                    return formOptionSetOccurrences;
                });

                spyOn(optionSetView, "validate");
                done();
            });

            describe('default behaviour', function () {
                var addButtonSpy, collapseButtonSpy;
                
                beforeEach(function (done) {

                    addButtonSpy = spyOn(optionSetView, "makeAddButton").and.callThrough();
                    collapseButtonSpy = spyOn(optionSetView, "makeCollapseButton").and.callThrough();
                    
                    optionSetView.layout().then(function () {
                        done();
                    });
                });

                it('should create a container for occurrence views and append it to DOM', function () {
                    expect(optionSetView.getEl().getElementsByClassName("occurrence-views-container").length).toEqual(1);
                });

                it('should create form option set occurrences', function () {
                    expect(api.form.FormOptionSetOccurrences).toHaveBeenCalled();
                });

                it('should perform layout of the option set occurrences with validation', function () {
                    expect(formOptionSetOccurrences.layout).toHaveBeenCalledWith(true);
                });

                it('should create add button for occurrences', function () {
                    expect(addButtonSpy).toHaveBeenCalled();
                });

                it('should create collapse button for occurrences', function () {
                    expect(collapseButtonSpy).toHaveBeenCalled();
                });

                it('should run validation', function () {
                    expect(optionSetView.validate).toHaveBeenCalled();
                });

            });

            describe('when layout is called without validation', function () {

                beforeEach(function (done) {
                    optionSetView.layout(false).then(function () {
                        done();
                    });
                });

                it('should NOT run validation', function () {
                    expect(optionSetView.validate).not.toHaveBeenCalled();
                });
            });
        });
```

* Don't access methods via ``object["methodName"]``, use ``spyOn(object, "methodName")``

* You shouldn't have to call the layout method every time you test validation. You need to run layout once, and then invoke ``validate`` method in ``beforeEach`` and have lots of its that test what happens after validate is called. We also have to include much more cases (each will have its own ``beforeEach`` with a call to ``validate`` with specific arguments, plus one main ``beforeEach`` with general setup - like for ``layout``):
1) default behaviour (``.validate`` with no arguments): here we should cover all of the common cases like breaking min/max occurrences and call to renderValidationErrors
2) case when we pass ``viewToSkipValidation`` into the method: here we should make sure that .validate is called for all occurrenceViews except for the one we passed
3) case when we pass ``silent=false``: here we should test that ``notifyValidityChanged`` is called

describe("what happens during validation") {
  beforeEach() {
     ..create common spies for the validate method...
     optionSetView.layout() {
       done();
     }
  }

  describe("common behaviour") {
    beforeEach() {
       ..create common spies for the validate method...
       optionSetView.validate();
       done();
     }

     it(...) {}
     it(...) {}
  }
  
  describe("when validation is not silent") {
    beforeEach() {
       ..create spies for this specific case...
       optionSetView.validate(false);
       done();
     }
     // Don't repeat specs from common scenario here!
     it("should notify about validation") {}
    }

  describe("when one view should not be validated") {
    beforeEach() {
       ...have a loop here that will set a spy on .validate method for each view in 
this.formItemOccurrences...
       optionSetView.validate(true, this.formItemOccurrences[2]);
       done();
     }
     // Don't repeat specs from common scenario here!
     it("should validate all views expect for the one specified") {}
     it("should not validate the specified view") {}
    }

  
}

* While checking the code, I noticed a lot of duplicate code in ``validate`` and ``handleFormOptionSetOccurrenceViewValidityChanged`` - see if you can refactor that

* Please avoid having specs for several different cases inside one "it", like "should render validation errors but not notify validity changed listeners" - these are two different things

* We need specs for event handlers of ``formItemOccurrences`` (method ``subscribeFormOptionSetOccurrencesOnEvents``) - you can get the object in the same way I did above while testing layout and then fire events on it.

* We need specs for clicks on ``addButton`` and ``collapseButton`` - you can get the buttons by setting ``.and.callThrough`` spies on ``makeAddButton`` and ``makeCollapseButton`` methods before calling layout, and then accessing spy.calls.mostRecent().returnValue

PR #3918, opened 2016-11-03
XP-4356 PageEditor won't get updated after a fragment has been remove…

sshklianko:
…d in the browse panel

- Added handling of content deleted event for FragmentComponentView
- Made PageComponentsView to maintain array of invalid items
- Added method to highlight invalid rows in PageComponentsTreeGrid + style for it
- Made  FragmentComponentView to notify about load error


alansemenov:
* When opening content with deleted fragment, it will highlight the missing fragment with red background in the editor, but not in the Page Components view (see image)
image

* Change the warning from "Fragment ... cannot be found" to "Fragment... could not be found" so that it's consistent with the message shown inside the placeholder.

* When fragment is deleted while the content is open in the Wizard, the deleted fragment will be highlighted, but we should also show the text "Fragment content could not be found" inside. Below is content with two fragments - one was initially missing (and it shows the text) and another was deleted after opening the content (and it does not show the text).

image

* Invalid node inside Page Components View should not turn to acid red upon selection. Just keep the same background as when it's not selected.

image


* Names of the new methods in PageComponentsView.ts are confusing:

1) ``markAsValid`` is not marking anything, it removes an item from invalid items, so it should be something like ``removeFromInvalidItems`` or ``excludefromInvalidItems``

2) same with ``markAsInvalid`` - it should be called something like ``addToInvalidItems``

3) ``updateInvalidRows`` - this is where we actually do visual stuff with invalid rows, so it should be called ``highlightInvalidItems``. I also think that this method should be called from inside the two methods mentioned above, since it always follows their call.

PR #3935, opened 2016-11-07
XP-4405 Inspect panel - Radio buttons inside component config are mis…

alansemenov:
…placed [Windows]


alansemenov:
@pmi Please review and merge

PR #3940, opened 2016-11-08
XP-4409 Clear publish.from on duplicate

runarmyklebust:
Fixed and added tests


GlennRicaud:
Simple and efficient. 
And a constant search of improvement by adding test for missing cases.
Perfect as usual

PR #3942, opened 2016-11-08
XP-3204 Html-area field values not unescaped on IndexDocument creation

GlennRicaud:


runarmyklebust:
This will wait to 6.9

PR #3951, opened 2016-11-09
XP-4410 OptionSet - Cosmetic improvements

alansemenov:
* Changed radiobuttons to look as in the prototype
* Fixed missing bottom border of an itemset inside an optionset

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/common/styles/api/form/form-option-set.less
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/common/styles/api/form/form-option-set.less
@@ -183,6 +183,10 @@
         display: none;
         opacity: 0.5;
 
+        .help-text-toggler {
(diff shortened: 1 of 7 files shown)


modules/admin/admin-ui/src/main/resources/web/admin/common/styles/api/form/form-option-set.less
```diff
@@ -221,10 +225,30 @@
...
+              content: "";
+              display: block;
+              position: absolute;
+              top: 2px;
+              left: -15px;
+              height: 14px;
+              width: 14px;
+              background: white;
+              border: 1px solid lightgray;
+              box-shadow: inset 0px 0px 0px 2px white;
```
sshklianko: Lets replace border-radius styling with a call to mixin class .box-shadow() from mixins.less

Same for border-radius

PR #3957, opened 2016-11-10
XP-4436 Add missing specs for Property class

vbradnitski:


alansemenov:
* You have an error in the test of "reset" method:
```
            it("reset property set to initial value", () => {
                let propertySet = new PropertySet(),
                    propertySetSpy = spyOn(property, "reset");
                ;

                property.setValue(new Value(propertySet, ValueTypes.DATA));

                property.reset();
                expect(propertySetSpy).toHaveBeenCalled();
            });
````
It should be ```propertySetSpy = spyOn(propertySet, "reset");```

* You call .getLong method in the tests for getDouble
* You call .getLocalDateTime in the tests for getLocalTime

All of the above is instantly obvious if you open coverage statistics: 
image

vbradnitski:
updated.

PR #3967, opened 2016-11-14
XP-4413 Site Config - Validation crashes on invalid value with regexp

sshklianko:
- Made site form to display validation errors on show
- Adjusted validateOccurrences() of BaseInputTypeNotManagingAdd to check for input validity before incrementing valid inputs count
- Adjusted TextInput to notify about its change not only when input is valid - the same way other inputs do
- Fixed problem with uninitialized errors container of ValidationRecordingViewer
- Adjusted InputValidatorBuilder to accept Form instead of ContentType, so that SiteConfig and other forms may be passed for validation directly
- Added site form inputs validation in ValidateContentDataCommand
- Added test for broken input in SiteConfig


alansemenov:
Works as it should

PR #4010, opened 2016-11-22
XP-4463 Content Grid - Highlight content items with read-only access

ashklianko:
-using 'expand' param to have full Content with permissions returned instead of ContentSummary
-added 'readOnly' parameter for ContentSummaryAndCompareStatus to mark content that user can't modify
-using  slickgrid's 'getItemMetadata(row)' method to add readonly css classes
-Setting title attribute with 'Read-only' value on node row via jquery because slickgrid does not provide such a possibilty; used grid's datachange event to trigger title attr update (had to call title attr update also in collapse node because it is not used to trigger datachange event)


ashklianko:
Forgot  to mention one thing: title attribute is added to readonly rows but our  tooltip from main.ts doesn't seem to work correctly. I beleive that is to be a separate issue.

PR #4051, opened 2016-11-29
XP-4457 Lib Content - Add support for accessing contenttype information

aro:


aro:
Yep, better getTypes. Updated with function renamed.

PR #4063, opened 2016-12-01
XP-4560 Time-based publishing - Publish Wizard - Schedule publishing

sshklianko:
- Added publish schedule dialog
- Adjusted publish request - added from and to fields
- Adjusted publish dialog to interact with schedule dialog
- Added Json class for Schedule object on back-end. Made PublishContentJson to use it.


GlennRicaud:
Very nice!

PR #4069, opened 2016-12-01
XP-4545 Site config permission check

GlennRicaud:


modules/admin/admin-ui/src/main/resources/web/admin/common/js/ui/selector/combobox/RichComboBox.ts
```diff
@@ -88,6 +88,15 @@ module api.ui.selector.combobox {
             }
         }
 
+        setReadOnly(readOnly: boolean) {
+            super.setReadOnly(readOnly);
+
+            this.comboBox.setReadOnly(readOnly);
+            this.selectedOptionsView.setEditable(!readOnly);
+
```
alansemenov: Here we call .setEditable directly on this.selectedOptionsView, but that has already been done by the combobox itself on the previous line?

pmi: Agree with that

modules/admin/admin-ui/src/main/resources/web/admin/common/js/content/site/inputtype/siteconfigurator/SiteConfigurator.ts
```diff
@@ -57,6 +69,7 @@ module api.content.site.inputtype.siteconfigurator {
             this.siteConfigProvider.onAfterPropertyChanged(() => this.ignorePropertyChange = false);
 
             this.comboBox = this.createComboBox(input, this.siteConfigProvider);
+            this.comboBox.setReadOnly(this.readOnly);
```
pmi: Made even better solution that ensures that setReadOnly is called only 1 time after readOnly request is finished

PR #4071, opened 2016-12-02
XP-4571 Time-based publishing - Content Studio - Publish status

sshklianko:
- Adjusted ContentResource.compare method to return publish status as well - as this method is used by treegrid so that we don't need to do subsequent call to getPublishStatuses. Changed appropriate json classes on back-end and front-end
- Adjusted ContentSummaryAndCompareStatus.ts to store publishResult field
- Adjusted statusFormatter of ContentRowFormatter so that it displays 'pending' and 'expired' publish statuses along with compare status in TreeGird
- Adjusted ContentWizardToolbarPublishControls to manage publishStatus along with compareStatus. Adjusted WizardPanel to set publish status into it.
- Added GetPublishStatusesRequest.ts and appropriate json and wrapper classes
- Adjusted styles

### DIFF
--- a/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/json/content/CompareContentResultJson.java
+++ b/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/json/content/CompareContentResultJson.java
@@ -8,10 +8,13 @@ public class CompareContentResultJson
 
     private final String id;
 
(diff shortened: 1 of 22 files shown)


GlennRicaud:
I think we can rework a bit ContentResource and the CompareContentResultsJson, but I will create another task for this. Seems to work correctly (I saw a mistake but not linked to this task).
Nice!

PR #4098, opened 2016-12-05
XP-4614 Source Button is not displayed on the MCE-toolbar  in the inl…

ashklianko:
…ine mode

-Showing source code button in inline mode by default


alansemenov:
@pmi No, it shouldn't be available for users outside those three roles. The problem is that now the source button is not available even if the user has one of these roles.

ashklianko:
https://github.com/enonic/xp/pull/4119

PR #4109, opened 2016-12-07
XP-4563 fix content.js so that it is rendered

sshklianko:
- Changed the way option set stores selection - as recently we've made option  set's PropertySet to store only properties of selected items - there is no need for usage of selection array. The fact that option exists in PropertySet means that it is selected, hence, removed its uses from back-end, though on front-end it's still used for user operations.
- Adjusted FormDefaultValuesProcessorImpl to add default options to option set's PropertySet
- Adjusted cleanFormRedundantData() method of ContentWizardPanel to remove selection array property
- Adjusted OccurrenceValidator
- Added and updated test


alansemenov:
> Changed the way option set stores selection 

@ashklianko Does it mean that you changed the storage format for the option-set? We already have customers who have built their content upon existing format, so we can't change the way we store selection, it's too late for that. Otherwise we will need an upgrade script and I don't see a reason for that.

PR #4115, opened 2016-12-07
XP-4644 Image caching - Store cached images using a hash

aro:


srs:
Looks good. @GlennRicaud will check it out.

PR #4122, opened 2016-12-08
XP-4653 In Content Studio, Browse view triggers parts twice

sshklianko:
- Removed extra check (that doubles same request) made by jquery that checks if portal may render preview - firstly, this is a job for isRenderable() check which stands above, secondly, XP-3126 is not reproduced anymore


pmi:
https://youtrack.enonic.net/issue/XP-3126 comes back with these changes so we have to keep the 404 check by either making a head request or updating isRenderable() method to return false

PR #4129, opened 2016-12-09
XP-4582 Setup TSLint

edloidas:
Added prepush Git hook, that will run TSLint on diff with `merge-base` from origin/master.
Updated `npm` dependencies.
Replaced `lodash` dependency with a specific function from it. 
Fixed TS compilation errors.
Added a separate class for HTTP status codes.
Fixed linting errors.


edloidas:
Please, don't merge it on Friday :)

srs:
I do not like githooks for this. It should be enforced during build instead. Everyone should build and test code before checking in anyway.

edloidas:
@srs I had a reason for Git Hooks and described it in YouTrack. But it's is possible do it with CI, so I will update the branch. 
Also, I would like you to read and approve the other changes (node version, to be specific). Please, read the YouTrack task and leave the comment here.

PR #4144, opened 2016-12-13
XP-4696 Wizard Step Form - X icon is not correctly aligned

alansemenov:
* Fixed alignment of the X icon via mixin

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/common/styles/variables/mixins.less
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/common/styles/variables/mixins.less
@@ -746,14 +746,15 @@
   }
 }
 
-.close-mixin() {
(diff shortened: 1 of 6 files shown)


modules/admin/admin-ui/src/main/resources/web/admin/common/styles/variables/mixins.less
```diff
@@ -746,14 +746,15 @@
...
-  cursor: pointer;
-  font-size: 12px;
-  display: block;
-  text-decoration: none;
+.icon-close-mixin() {
+  .icon-close {
+    .focusable();
+    position: absolute;
+    top: 8px;
+    right: 0px;
```
pmi: Maybe we should make it 8px as well as top ?

PR #4172, opened 2016-12-15
XP-4705 Alert dialog always appears, when new image was uploaded in t…

sshklianko:
…he wizard but Save was not pressed

- Added synchronization between persisted content extradata and mixins data - while rendering form we intentionally add fields from mixins and populate viewed content with it's data, but persisted content was not aware of mixins  data and caused hasUnsavedChanges() to return true

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/wizard/ContentWizardPanel.ts
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/wizard/ContentWizardPanel.ts
@@ -642,7 +642,7 @@ export class ContentWizardPanel extends api.app.wizard.WizardPanel<Content> {
 
         this.updateWizardHeader(content);
         this.updateWizardStepForms(content, unchangedOnly);
-        this.updateMetadataAndMetadataStepForms(content.clone(), unchangedOnly);
+        this.updateMetadataAndMetadataStepForms(content, unchangedOnly);
         this.resetLastFocusedElement();
     }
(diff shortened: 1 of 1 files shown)


modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -1071,6 +1073,22 @@ export class ContentWizardPanel extends api.app.wizard.WizardPanel<Content> {
...
 
+    // synch persisted content extra data with data from mixins
+    // when rendering form - we may add extra fields from mixins; as this is intended action from XP, not user - it should be present in persisted content
+    private synchPersistedItemWithMixinData(mixinName: MixinName, mixinData: PropertyTree) {
+        var persistedContent = this.getPersistedItem(),
+            extraData = persistedContent.getExtraData(mixinName);
+        if (!extraData) { // ensure ExtraData object corresponds to each step form
+            extraData = new ExtraData(mixinName, mixinData.copy());
+            persistedContent.getAllExtraData().push(extraData);
+        } else {
```
alansemenov: We more or less repeat the code that pushes extra data into content in three different methods now: ``doLayoutPersistedItem``, ``synchPersistedItemWithMixinData`` and ``updateMetadataAndMetadataStepForms``. Let's extract it to a separate method with arguments of ``content``, ``mixinName`` and optionally (if empty - create new instance) ``propertyTree``.

PR #4180, opened 2016-12-16
XP-4733 Don't override native browser outline highlighting

alansemenov:
* Removed overrides of native browser outlining


alansemenov:
Good point - will fix

alansemenov:
@pmi Fixed - please review

alansemenov:
We don't need outline inside the launcher panel since we already highlight apps when you browse with up and down arrows. It's only the launcher/X icon that we need to outline.

I'm pretty sure I tested the launcher button in apps :) - see attached. Did you run the full build? Strangely enough, _launcher.css is only generated with gradle build, not enough to build with gulp.

image
image

pmi:
Right, full build fixed missing icon ! 
Focus works nicely !

PR #4189, opened 2016-12-19
XP-4740 Publish dialog - Endless spinner on the "Publish" button

edloidas:
Fixed the stash lock/unlock update.
Added `.no-publish` class to hide spinner, when dialog locked and publish is disabled.

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/publish/ContentPublishDialog.ts
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/publish/ContentPublishDialog.ts
@@ -146,12 +146,12 @@ export class ContentPublishDialog extends ProgressBarDialog {
             return this.reloadPublishDependencies(childrenNotLoadedYet);
         } else {
             // apply the stash to avoid extra heavy request
-            this.lockControls();
+            this.updateSubTitleShowScheduleAndButtonCount();
             this.loadMask.show();
             setTimeout(() => {
                 this.setDependantItems(stashedItems.slice());
                 this.centerMyself();
-                this.unlockControls();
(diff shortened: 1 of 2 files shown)


modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/publish/ContentPublishDialog.ts
```diff
@@ -146,12 +146,12 @@ export class ContentPublishDialog extends ProgressBarDialog {
...
             // apply the stash to avoid extra heavy request
-            this.lockControls();
+            this.updateSubTitleShowScheduleAndButtonCount();
             this.loadMask.show();
             setTimeout(() => {
                 this.setDependantItems(stashedItems.slice());
                 this.centerMyself();
-                this.unlockControls();
+                this.updateSubTitleShowScheduleAndButtonCount();
                 this.loadMask.hide();
```
alansemenov: @edloidas  ``this.updateSubTitleShowScheduleAndButtonCount`` calls ``this.countTotal`` which is pretty heavy performance-wise (think of publishing several thousands of items). Can you explain why we need to call ``this.updateSubTitleShowScheduleAndButtonCount`` twice with 100ms delay? The number of items cannot change in this period, can it?

edloidas: You are right. I replaced the first `this.updateSubTitleShowScheduleAndButtonCount()` call with simple `lockControls()` call. We need the second `this.updateSubTitleShowScheduleAndButtonCount()` call, though, since the dependent items are restored from stash and we should check, whether we need to enable the Publish. Also, we need to recalculate the number of items to be published.

alansemenov:
Looks fine now

PR #4190, opened 2016-12-19
XP-4700 Fragment-wizard won't get updated after the related image has…

pmi:
… been updated (zoomed, focused)

- added refresh on content updated event for fragment -, content containing fragment -, and image wizards if updated content is present on page

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/wizard/ContentWizardPanel.ts
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/wizard/ContentWizardPanel.ts
@@ -86,6 +86,15 @@ import Toolbar = api.ui.toolbar.Toolbar;
 import CycleButton = api.ui.button.CycleButton;
 
 import Permission = api.security.acl.Permission;
+import Region = api.content.page.region.Region;
+import Component = api.content.page.region.Component;
+import ImageComponent = api.content.page.region.ImageComponent;
+import ImageComponentType = api.content.page.region.ImageComponentType;
+import ObjectHelper = api.ObjectHelper;
+import LayoutComponentType = api.content.page.region.LayoutComponentType;
+import LayoutComponent = api.content.page.region.LayoutComponent;
+import FragmentComponent = api.content.page.region.FragmentComponent;
+import FragmentComponentType = api.content.page.region.FragmentComponentType;
 
 export class ContentWizardPanel extends api.app.wizard.WizardPanel<Content> {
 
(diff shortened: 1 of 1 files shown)


modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -747,17 +754,28 @@ export class ContentWizardPanel extends api.app.wizard.WizardPanel<Content> {
...
+                if (this.doHtmlAreasContainId(contentId.toString()) ||
+                    this.doesFragmentContainId(this.getPersistedItem().getPage(), contentId)) {
+                    containsHandler();
+                } else {
+                    this.doImageComponentsContainId(contentId).then((contains) => {
+                        if (contains) {
+                            containsHandler();
+                        }
+                    });
+                }
```
alansemenov: The logic here is getting difficult to follow. Let's add one more method, call it for example ``doPageComponentsContainId``which calls all these three methods (HtmlAreas, Fragment, Image) under the hood and returns a promise with ``contains`` (it should be defined as ``boolean`` in arguments, btw). This way we can always expand this method with logic for other components in the future. Then the call here will boil down to:
```Typescript
this.doPageComponentsContainId(contentId).then((containsId: boolean) => {
  if (containsId) {
    containsHandler();
  }
});
```

modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -845,6 +863,49 @@ export class ContentWizardPanel extends api.app.wizard.WizardPanel<Content> {
...
+                    return false;
+                } else if (ObjectHelper.iFrameSafeInstanceOf(component.getType(), ImageComponentType)) {
+                    return (<ImageComponent>component).getImage().equals(id);
+                } else if (ObjectHelper.iFrameSafeInstanceOf(component.getType(), LayoutComponentType)) {
+                    return this.doRegionsContainId((<LayoutComponent>component).getRegions().getRegions(), id, fragments);
+                }
+            })
+        });
+    }
+
```
alansemenov: Can you change this one so that it doesn't have ``else`` and has the last line with default ``return``?
```Typescript
{
  if (condition1) {
    do stuff;
    return false;
  }
  if (condition2) {
    return stuff;
  }
  return false;
}
```

PR #4203, opened 2016-12-21
XP-4730 Wrong feedback in the publishing dialog when invalid contents…

sshklianko:
… prevent publishing

- Added new field 'allContentsAreValid" ' to ResolvePublishContentResultJson class that indicates if all resolved contents that do not pend deletion are valid. Adjusted appropriate response and json handlers on UI
- Added new method checkAllContentsValid() to ContentService that uses CheckContentsValidCommand
- Adjusted ContentResource.resolvePublishContent() to use new service method and set value to new field in response json
- Adjusted PublishContentCommand.pushAndDelete() method to set contentIds that did not pass validity check into failed field of resultBuilder
- Adjusted publishDialog to store back-end validity check result and use it. NB! that on UI we additionally check that resolved items do have display name which is a little inconsistent with back-end check


alansemenov:
Can you rename the field from "allContentsAreValid" to "containsInvalid"? It sounds better and is consistent with the "containsRemovable" field from before.

image

PR #4230, opened 2016-12-29
XP-4748 AppBar - Enable configurable tabMenu

ashklianko:
-Refactored Content,Users and Applications  app panels code structure to reflect the fact that content app panel does not hold tabs anymore
-Moved tab navigation out of content and applications app panels to generic NavigatedAppPanel.ts which is now parent only for Users app panel
-Added TabbedAppBar.ts class to represent app bar that can hold tabs; regular app bar now only holds home and launcher buttons


pmi:
All seems to be good, except for flickering launcher button during page load, making app button to jump

ashklianko:
https://github.com/enonic/xp/pull/4235

PR #4242, opened 2017-01-03
XP-4796 Insert Link Dialog: doesn't get focus when opened in 6.9 B3

edloidas:
Added `setFirstFocusField` call to the `LinkModal` dialog.


modules/admin/admin-ui/src/main/resources/web/admin/common/js/util/htmlarea/dialog/LinkModalDialog.ts
```diff
@@ -52,6 +52,8 @@ module api.util.htmlarea.dialog {
                 this.textFormItem.hide();
                 this.textFormItem.removeValidator();
             }
+
+            this.setFirstFocusField(this.textFormItem.getInput());
```
alansemenov: I don't think we need to focus the input when the field is hidden, do we? This line should probably be moved inside the "if" block 5 lines above.

PR #4284, opened 2017-01-12
XP-4855 Content Wizard - principal is not refreshed from time to time…

ashklianko:
…, when a content version was changed

-Issue is that async calls are made in doSetValue() of ComboBox, but method invoking it does not know about it and continue processing with values that are not yet updated. This is actual for classes with 'displayMissingSelectedOptions' set to true because they make backend call to check if contents are missing


ashklianko:
https://github.com/enonic/xp/pull/4295

PR #4325, opened 2017-01-20
XP-4890 Publishing Wizard - Enable excluding any item from the depend…

vbradnitski:
…ants list if it's not a parent to other items in the list


srs:
Please add more tests so we get better code-coverage on the new code. It's only 70.58%. Try to get over 75%.

alansemenov:
@runarmyklebust Can you look at this one please? It works as it should but I'm not sure about the backend part.

runarmyklebust:
Im sceptic to add this "required" stuff to the node domain. This will impact performance for something needed specific for content stuff I think

PR #4330, opened 2017-01-30
XP-4889 Option Set - Don't clear input values when an option is unsel…

sshklianko:
…ected

- Added BeforeContentSavedEvent which is fired in ContentWWizardPanel.saveChanges()
- Added handling of that event to FormOptionSetOptionView
- Added check to  combobox.selectExistingAndHandleMissing() method so that we do not notify about missing options when missing ids array is empty
- Removed redundant events


alansemenov:
Will be merged from another PR

PR #4339, opened 2017-01-31
XP-4926 Publishing wizard - Unnamed dependencies are allowed to be pu…

vbradnitski:
…blished

### DIFF
--- a/modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/CheckContentsValidCommand.java
+++ b/modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/CheckContentsValidCommand.java
@@ -36,8 +36,8 @@ public class CheckContentsValidCommand
 
         final LogicalExpr contentIsUnnamed = LogicalExpr.
             or( new NotExpr( CompareExpr.
-                like( FieldExpr.from( ContentPropertyNames.DISPLAY_NAME ), ValueExpr.string( "*" ) ) ), CompareExpr.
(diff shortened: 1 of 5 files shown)


modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/CheckContentsValidCommand.java
```diff
@@ -29,11 +34,16 @@ public boolean execute()
             return true;
         }
 
+        final QueryExpr selectInvalidExpr = QueryExpr.from( LogicalExpr.
+            or( LogicalExpr.
+                    or( new NotExpr( CompareExpr.
+                        like( FieldExpr.from( ContentPropertyNames.DISPLAY_NAME ), ValueExpr.string( "*" ) ) ), CompareExpr.
```
runarmyklebust: Why do you need the DISPLAYNAME = "*" here?

vbradnitski: It's not DISPLAYNAME = "*", it's NotExpr from DISPLAYNAME = "*"

runarmyklebust:
Ok, I think that the displayName != "" should be a part of the regular content.valid-check, waiting for TSI to mean something

runarmyklebust:
Also mister, please add tests in ContentServiceImplTest_isValidContent for the different new scenarios

runarmyklebust:
Please add the required checks in the "valid" check instead, so these new query-parts could be removed.

PR #4340, opened 2017-02-01
XP-4881 HTML Area - Implement Find/Replace modal dialog

ashklianko:


alansemenov:
Please verify changes with lint - it gives tons of errors now. Make sure you do that before every commit.

ashklianko:
Updated; Esc and Enter fixes not yet in that brach so navigation only on pressing buttons for now.

PR #4372, opened 2017-02-07
XP-4955 Page Editor - Add "Edit" to the context menu of image and fra…

ashklianko:
…gment components

-Added additional hierarchy layer between ComponentView and Inheritors that are backed up with some Content (Currently it is Image and Fragment Component Views), layer adds 'edit' button into context menu

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/common/js/liveedit/ContentBackedComponentView.ts
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/common/js/liveedit/ContentBackedComponentView.ts
@@ -1,0 +1,59 @@
+module api.liveedit {
+
+    import Component = api.content.page.region.Component;
+    import ContentSummary = api.content.ContentSummary;
+    import ContentSummaryBuilder = api.content.ContentSummaryBuilder;
+    import ContentSummaryAndCompareStatus = api.content.ContentSummaryAndCompareStatus;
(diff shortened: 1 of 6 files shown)


modules/admin/admin-ui/src/main/resources/web/admin/common/js/liveedit/ContentBackedComponentView.ts
```diff
@@ -0,0 +1,59 @@
...
+    import ContentTypeName = api.schema.content.ContentTypeName;
+
+    export class ContentBackedComponentViewBuilder<COMPONENT extends Component> extends ComponentViewBuilder<COMPONENT> {
+
+        contentTypeName: ContentTypeName;
+
+        constructor() {
+            super();
+        }
+
```
alansemenov: It's not necessary to have an empty constructor with only call to ``super()``. Maybe we should pass ``contentTypeName`` as an argument to the constructor since it's a mandatory property for the class.

alansemenov: Also, let's rename to ``ContentBasedComponentViewBuilder`` and ``ContentBasedComponentView``

ashklianko:
https://github.com/enonic/xp/pull/4388

Issue #4405, opened 2017-02-13
Move SnapshotService to api, remove snapshot-methods from NodeService


runarmyklebust:
We need more functionality in the SnapshotService, and the NodeService is not the correct placement for these functions.

GlennRicaud:
nodeService.restore and snapshotService.restore were not equivalent.
Skipping the calls to nodeStorageService  and eventPublisher  will cause problems

GlennRicaud:
Created issue #4605

PR #4435, opened 2017-02-20
Move SnapshotService to api, remove snapshot-methods from NodeService…

ashklianko:
… (#4405)


ashklianko:
@runarmyklebust , @GlennRicaud  data toolbox code to be updated acccording to these changes

Issue #4419, opened 2017-02-15
Publishing Wizard - Exclude from publishing is not properly working


alansemenov:
* Select Features site
* Create Image Selector content. Leave name empty.
* Upload a new image (from disk) in one of the selectors.
* Save content.
* Select Publish Tree for "Features"

/features/unnamed is first item in the list and cannot be excluded from the list because its child item (the uploaded image) is also in the list - fine.

![image](https://cloud.githubusercontent.com/assets/11612490/22972645/dc5ee880-f37a-11e6-9886-6beb9c148292.png)

* Click "X" for the image under /features/unnamed

The list of dependants is refreshed but there's still no "X" icon for /features/unnamed - this is a bug.

![image](https://cloud.githubusercontent.com/assets/11612490/22972658/e83d74dc-f37a-11e6-9a9c-258d06044443.png)

* Close the Publishing dialog, delete the image under /features/unnamed and open "Publish Tree" for "Features" again there will be "X" icon for /features/unnamed, which is correct.

![image](https://cloud.githubusercontent.com/assets/11612490/22972670/f36ff0a0-f37a-11e6-80c2-990177f3bf3d.png)

vbradnitski:
https://github.com/enonic/xp/pull/4439
dependent list should be reloaded instead of simple filtering.

alansemenov:
Also it seems like direct parent of selected items is allowed to be excluded from the list (see last item on the screenshot)

![image](https://cloud.githubusercontent.com/assets/11612490/23159328/a818a410-f822-11e6-962a-93f038d6925d.png)

PR #4439, opened 2017-02-20
Publishing Wizard - Exclude from publishing is not properly working (…

vbradnitski:
…#4419)


alansemenov:
@vbradnitski Can you add tests so that coverage is sufficient?

Issue #4410, opened 2017-02-14
Page Editor/Inspect Panel - Give a warning before changing to page template


alansemenov:
When switching from Custom to Automatic or a specific page template in the Inspect panel we automatically reload the page to render it with selected controller. As a result, all the custom changes done to the page will be lost.
Give a warning (use ConfirmationDialog, the result should look like SaveBeforeCloseDialog) which will say "Switching to the page template will result in losing all custom changes made to the page. Are you sure?" and two buttons "Yes" and "No". Switch to the template only if user selects "Yes".

alansemenov:
Nice refactoring!

A few things to fix:

1. The confirmation must appear **only** when switching from Custom to Auto or specific Template. Now I get a confirmation when switching from Auto to Template or from Template to Auto as well. The idea is that in Custom mode there might be custom changes that can be lost when switching to page template mode. When user is switching between Auto and a page template there are no custom changes to lose.

2. Text inside the dialog must be inside <h6> tag instead of <div>. Look at SaveBeforeCloseDialog in the Users app - it should look the same.

3. It would be nice if we could show a confirmation when switching between page controllers in the Custom mode as well, since we also reload the page and may lose changes. The confirmation will look the same except for the text: "Changing a page controller will result in losing changes made to the page. Are you sure?"

4. I'm not sure if it's possible with our current tracking of changes but it would be nice to implement these two use cases:
* I open the page with either Auto or Page Template selected. I change to "Custom". I don't make any changes to the page. I switch back to either Auto or Page Template. No confirmation should be shown since I haven't done any changes to the page yet.
* (This one is about the second confirmation I mentioned above, that is yet to be implemented). I open the page with either Auto or Page Template selected. I change to "Custom". I select another controller from the dropdown (haven't made any changes yet). In this case we should not ask for confirmation either since user haven't made any changes with previous controller yet.

ashklianko:
Ok, on it!

PR #4467, opened 2017-02-23
Page Editor/Inspect Panel - Give a warning before changing to page te…

ashklianko:
…mplate #4410


alansemenov:
Found one small UI issue. If I try to switch from "Custom" to a page template and click "No" in the confirmation, selected option is reverted to "Custom" but when I expand the dropdown it has the page template highlighted (see below). It should originally highlight the currently selected option.

image

ashklianko:
Ok, will fix

ashklianko:
Done.

ashklianko:
That would be non-intuitive; If I read this method I think what is it really doing: selecting option or reverting active selection, that seems totally unrelated to each other

Issue #4423, opened 2017-02-16
User with "Users Administrator" rights cannot change a password


alansemenov:
1. Create a new user in the Users App.
2. Add him to the following groups: Users Administrator, Users App and Administration Console Login
3. Log out and log in as this new user
4. Go to the Users App, wizard page of the user
5. Try to change the password. Nothing will happen when you click the "Change Password" button

There are two problems here:
1) User Administrators must be able to change the password
2) We don't give any error message in the wizard even though there's exception in the XP log: "Access denied to [/identity/system/users/alan] for [MODIFY] by user [user:system:alan] ''Alan'". We should show error notification at the bottom of the page, as always. Also, see if you can change this ugly error message to something like "Insufficient rights for user ``<displayName>``".

alansemenov:
There's a support ticket for this issue: https://enonic.zendesk.com/agent/tickets/1823

PR #4469, opened 2017-02-23
User with "Users Administrator" rights cannot change a password (#4423)

vbradnitski:


GlennRicaud:
@vbradnitski  I noticed on multiple recent commits from you that you have a different reformatting. It makes reading a bit harder. Could you fix it.

vbradnitski:
hmm, I use the latest codestyle scheme from dev-tools
image
 Should I add any additional settings?

GlennRicaud:
You are not the only one apparently. Seems to depend on the IDEA version.
Will mention it tomorrow at the sprint meeting.

PR #4486, opened 2017-02-27
Move dialog - Show tree structure inside the content selector

sshklianko:
- Splitted DropdownGrid into DropdownListGrid (which holds regular grid that was previously used) and DropdownTreeGrid (which contains tree grid)
- Created parametrized OptionTreeGrid that  extends tree grid in order to manage Options. Created OptionDataLoader and OptionDataHelper interfaces to assist OptionsTreeGrid to managed parametrized data. Created ContentSummaryOptionDataLoader and ContentSummaryOptionDataHelper that implement these interfaces.
- Created ContentSummaryFetcher similar to ContentSummaryAndCompareStatusFetcher
- Renamed RichComboboxCombobox to LoaderComboBox.
- Adjusted Grid and GridBuilder classes to accept defaultWidth and defaultHeight params as strings so that such values as 'auto' and '100%' could be passed. It enabled removing some '!important;' entries from tree-grid.less
- Fixed hasClass check in ElementHelper
- Code refactoring

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/browse/ContentDropdownTreeGrid.ts
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/browse/ContentDropdownTreeGrid.ts
@@ -1,0 +1,621 @@
+import '../../api.ts';
+
+import Viewer = api.ui.Viewer;
+import TreeGrid = api.ui.treegrid.TreeGrid;
+import TreeNode = api.ui.treegrid.TreeNode;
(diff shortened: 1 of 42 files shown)


alansemenov:
Great job! Some things to fix:
1. The first time I open the dropdown, scrolling is not working. If I collapse and expand again, it works fine.
2. If I search by keyword, then select a node, then click "x" to delete selection, the keyword remains in the field. We should clear it after clicking "x".
3. Would be convenient to auto-expand the tree down to selected item, and highlight selected item in the dropdown. This will help a lot if I want to move the item one level up, for example, or to a sibling folder.

Issue #4437, opened 2017-02-20
Unselecting an item will not refresh BrowseItemsSelectionPanel when more than 10 items selected in the grid


sgauruseu:
1. open the CS and select more than 10 items in the grid
2. click on 'icon remove' and remove 5-10 items from the SelectionPanel
BUG: SelectionPanel was not refreshed after the removing of items
 expected: when any items were removed from the panel, new selected items should be displayed on the panel
![sel](https://cloud.githubusercontent.com/assets/3728712/23126018/cfbfa7ae-f785-11e6-85da-d23b46feb0e9.png)

PR #4493, opened 2017-02-28
Unselecting an item will not refresh BrowseItemsSelectionPanel when more than 10 items selected in the grid [#4437]

edloidas:
Updated `removeItem()` method to call `updateDisplayedSelection()` upon removing item from displayed selection.
Updated `updateDisplayedSelection()` to append only the absent children.


edloidas:
Based on commits, done to #4484

Issue #4491, opened 2017-02-28
Insert Image dialog -Errors appears in the browser console when an imported image was selected in the dialog


sgauruseu:
1. Open htmlArea dialog and click on the Insert Image menu item
2. try to Insert an image, that was imported via the API CreateMediaParams
BUG: 
-.when image was selected, error appears in the console. 
-. click on the 'keep original size' checkbox and select the 'tall 9:16' cropping effect. Endless spinner appears
and 500 ServerError appears in the server.log
![500](https://cloud.githubusercontent.com/assets/3728712/23399557/b3d187e2-fdb1-11e6-9ec5-bc4e6f408d37.png)

![pop](https://cloud.githubusercontent.com/assets/3728712/23399310/d33220de-fdb0-11e6-87a8-78c3c5c89e9b.png)

PR #4499, opened 2017-02-28
Insert Image dialog -Errors appears in the browser console when an im…

ashklianko:
…ported image was selected in the dialog #4491


ashklianko:
I was not able to reproduce, only on Sergei's machine;

PR #4500, opened 2017-02-28
Content Wizard - NullPointer exception when saving unnamed content

sshklianko:
- Added site and mixin services to rename command as they are required for validation command when content is site or has mixins


sshklianko:
Issue occurred with Features contents because they had mixins inside, which were supposed to be validated, but faile due to missing mixin service.

Issue #4425, opened 2017-02-17
Show content status as "New" when the "First Published" field is empty


alansemenov:
1. When content item is not in master branch (compareStatus is "NEW") and the "First Published" field is empty, show item status as "New" (instead of "Offline") everywhere in the admin UI.

2. If item is not in master branch but "First Published" field is set, that means that the item was published and then taken offline, in this case we should show "Offline".

Use color ``@admin-font-gray1 (#acacac)`` for Offline status.

![image](https://cloud.githubusercontent.com/assets/11612490/23059773/2194fc42-f4fb-11e6-948d-e3958dbc18ca.png)

runarmyklebust:
I think this should be done in the ContentResource, since we will need to handle this status in the publish wizard anyway because of the filtering of "offline" vs "new items" in the wiz

PR #4501, opened 2017-02-28
Show content status as "New" when the "First Published" field is empty #4425

vbradnitski:


alansemenov:
Client-side code looks good. @GlennRicaud can you verify java code?

GlennRicaud:
There is something strange here. Why is ContentSummary.js in conflict? It should not.
"issue-4425" branch should have been created from "issue-4424".

GlennRicaud:
The pull request is strange as I said, it does not update correctly the changed files. There should actually be 2 lines of java code. So I will close this PR and create a new one

PR #4514, opened 2017-03-01
SelectionController checkbox doesn't change state after content was l…

sshklianko:
…azy loaded

- Adjusted isAllSelected() method to check case when there is mock load node within treegrid, which don't get selected (by default behavior)

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/common/js/ui/treegrid/TreeGrid.ts
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/common/js/ui/treegrid/TreeGrid.ts
@@ -1444,7 +1444,14 @@ module api.ui.treegrid {
         }
 
         isAllSelected(): boolean {
-            return this.grid.isAllSelected();
+            return this.grid.isAllSelected() || this.thereIsOneUnselectedEmptyLoadNode();
+        }
+
+        private thereIsOneUnselectedEmptyLoadNode(): boolean {
+            return this.grid.getDataLength() - this.grid.getSelectedRows().length == 1
+                   && this.grid.getDataView().getItems().some((data: TreeNode<DATA>) => {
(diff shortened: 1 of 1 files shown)


modules/admin/admin-ui/src/main/resources/web/admin/common/js/ui/treegrid/TreeGrid.ts
```diff
@@ -1444,7 +1444,14 @@ module api.ui.treegrid {
...
         isAllSelected(): boolean {
-            return this.grid.isAllSelected();
+            return this.grid.isAllSelected() || this.thereIsOneUnselectedEmptyLoadNode();
+        }
+
+        private thereIsOneUnselectedEmptyLoadNode(): boolean {
+            return this.grid.getDataLength() - this.grid.getSelectedRows().length == 1
+                   && this.grid.getDataView().getItems().some((data: TreeNode<DATA>) => {
+                    return data && data.getDataId() === '';
+                });
```
alansemenov: I'm very confused by this fix. Why is the bug reproduced in the first place? Why does it happen only when there's one empty node? (Based on the bug description, all items are selected in the grid.) 

I understand that it's caused by a fake node created by the lazy loader, but it's absolutely not obvious from looking at the code.

I think we should come from the opposite direction here and change ``isAllSelected`` to return ``true`` when all **non-empty** nodes are selected.

```Typescript

        isAllSelected(): boolean {
            if (this.grid.isAllSelected()) {
                return true;
            }

            let selectedNodes = this.grid.getSelectedRows();

            if (!selectedNodes) {
                return false;
            }

            let nonEmptyNodes = this.gridData.getItems().filter((data: TreeNode<DATA>) => {
                return (!!data && data.getDataId() !== '');
            });

            return nonEmptyNodes.length === selectedNodes.length;
        }
```

Issue #4427, opened 2017-02-17
In Publishing Wizard dependants should not by default include "Offline" items


alansemenov:
The list of dependant items should not include "Offline" items (those that were unpublished), unless they are mandatory references for the main selected items.
Add a checkbox called "Include offline items" in the bottom left corner of the dialog ("off" by default). When turned on, it should fetch "Offline" references (also children, if "Include children" icon is "on") and add them to the list.

runarmyklebust:
My first thought; filter these in the ContentResource.

GlennRicaud:
After discussion:
* The filtering of offline items is done only on children resolution (before dependency resolution)
* contentLib.push/publish include offline items

PR #4522, opened 2017-03-01
In Publishing Wizard dependants should not by default include "Offline" items #4427

vbradnitski:
…e" items (#4427)


alansemenov:
@runarmyklebust review please?

PR #4590, opened 2017-03-10
Create IssueService with method storeIssue

sshklianko:


sshklianko:
@runarmyklebust 
As discussed, Ive left 1. as is - we should be able to set Issue Id when traversing node to issue
2. I created IssueInitialzier and moved initIssueNode() there. I did not add there initializeRepository() and createDraftBranch() methods as to me it seems that ContentServiceImpl will call it's initialize method earlier and will do all the necessary repo inits.
3. I made IssueNodeTranslator to be a class with static methods.

sshklianko:
See updated PR: https://github.com/enonic/xp/pull/4596

PR #4600, opened 2017-03-13
In Publishing Wizard dependants should not by default include "Offline" items (#4427)

GlennRicaud:


GlennRicaud:
Bedre? @runarmyklebust

Issue #4561, opened 2017-03-07
Create IssueService.updateIssue


runarmyklebust:
Depends on #4559 

Create method updateIssue in IssueService.

Accepts a updateIssueParams, returns updated Issue.

PR #4606, opened 2017-03-14
#4561 - Create IssueService.updateIssue

sshklianko:


sshklianko:
See updated PR: https://github.com/enonic/xp/pull/4608

Issue #4547, opened 2017-03-06
Implement create issue REST API 


alansemenov:
Depends on #4559 
Implement REST API to create an issue, create in a new Resouce IssueResource

Fields:
- title (mandatory)
- list of contentId (mandatory)
- list of approver user ids (mandatory)
- description (optional)

Id, date/time, status should be set automatically. New issues should all get "Open" status by default.

API should respond with JSON of the new issue (if it was created successfully) or error message (if not).

PR #4617, opened 2017-03-14
#4547 - Implement create issue REST API

sshklianko:
- Updated CreateCommand to set Creator and CreatedDate in command, not params
- Added issue service with three methods and appropriate classes
- Missing tests


sshklianko:
@NB, its missing tests and implements create, update and get api

sshklianko:
@runarmyklebust , assign to @vbradnitski if change is required

Issue #4580, opened 2017-03-08
Automatically update the details panel whenever widgets are updated


alansemenov:
Widget list inside the dropdown in the details panel must be updated automatically whenever there's a change in the list of available widgets (a new one installed, or existing one updated or uninstalled). This should result in repaint of widget view if the updated widget is currently selected.

Issue #4581, opened 2017-03-09
'Version history' and 'dependencies' are displayed at the same time 


sgauruseu:
1. select a site and open the Details Panel
2. open the Dependencies view
3. switch to the 'Version history' view
4. switch to Dependencies again
BUG  both views are displayed at the same time
![dep](https://cloud.githubusercontent.com/assets/3728712/23747643/77cdfbdc-04d1-11e7-831f-87521e1f9032.png)

PR #4630, opened 2017-03-16
Automatically update the details panel whenever widgets are updated (#4580)

edloidas:
Added widgets update handler when apps are started, stopped or updated.

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/view/detail/DetailsView.ts
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/view/detail/DetailsView.ts
@@ -117,12 +117,13 @@ export class DetailsView extends api.dom.DivEl {
 
                 const activeIndex = this.widgetViews.map(view => view.getWidgetName()).indexOf(activeWidgetName);
                 const active = this.widgetViews[activeIndex];
+                const isSameActive = active && (active.getWidgetName() === activeWidgetName);
 
                 if (!active && event.getEventType() === ApplicationEventType.STOPPED) {
                     this.activateDefaultWidget();
                     this.widgetsSelectionRow.updateState(this.activeWidget);
                     this.updateActiveWidget();
-                } else if (active) {
+                } else if (isSameActive) {
                     this.setActiveWidget(active);
                     this.updateActiveWidget();
                 }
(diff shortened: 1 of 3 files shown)


edloidas:
I also discovered a problem with details panel. The widgets display is really messed up. To see what I'm talking about, do the following:
* Select a content
* Open Details panel. Info Widget will be displayed.
* Select Dependencies Widget
* Press button to display the Info Widget again
* Select the Version history Widget.

You will see both Dependencies and Version history Widgets. But only one widget should be displayed at a time.

alansemenov:
@edloidas Yes, we already registered a bug on this issue: #4581. I assigned it to you, since you worked with widgets already.

edloidas:
Added 1s latency with debounce function and updated the list of events, that will lead to widgets update. Since widgets don't require the immediate update, 1s is small enough amount of time for the user to wait, until the app is started till it's ready to be used. It's also enough for the server to do some background jobs, and enough time to skip the repeated progress/status app events.

alansemenov:
All the points mentioned above are now fixed and working fine, but I found two more issues:

1. If current widget is uninstalled or deleted, we reset selection to default widget (Version history) but the details panel doesn't seem to re-render correctly (see below):
image

2. When a new widget is deployed or installed, we instantly make it active. This is unnecessary and even irritating, if I currently work with another widget. We should only activate the widget when it's updated.

alansemenov:
About #2 - the behaviour is not exactly what I said, but strange nevertheless - the new widget doesn't seem to be selected, but this code is executed every time I add or remove a widget:
```
if (active) {
  _this.setActiveWidget(active);
  _this.updateActiveWidget();
}
``

If I'm on "Version history" and add or delete a widget, it should not repaint the details panel, since active widget is not affected in any way.

edloidas:
Since the basic solution became too unreadable and had some unsolvable problems, I fully refactored it, almost from scratch:
* All App Events will be handled by the aggregator `handleWidgetsUpdate()`. That function will filter the events and call the debounce function for each Widget. Thus we are now sure to call update for each application, without fear to skip one.
* `handleWidgetUpdate()` call is debounced. It will remove/install/upgrade the widget, that is related to the event.
* The handler won't repaint the active widget if wasn't updated and will switch to default (Info) widget if the active widget was removed.
* __Also__ refactored `getAndInitCustomWidgetViews`: renamed it to `fetchAndInitCustomWidgetViews` and simplified it.

Please, test again for each previous cases, because I may forget to test something.

alansemenov:
Nice job, @edloidas !

Issue #4462, opened 2017-02-23
Principal with uppercase name


GlennRicaud:
Different problem seen or reported by Rune:

* Programatically:
authLib.createUser({
        userStore: 'system',
        name: 'G013080',
        displayName: 'Rune Forberg',
        email: 'rune.forberg@gjensidige.no'
    });

- [x] Creates a user but the email is not displayed
- [x] In Users admin tool, while listed the selection of the user throws errors: Principal [user:system:G013080] was not found 
- [x] Related but check that users can be deleted and are correctly deleted

Manually:
- [x] Creating a user with a name containing uppercase will cause the same problem as above

Some quick check showed that:
* The user are returned with a key which is the concatenation of "user:" + userstore + ":" + _name
* But when searched, it is done on the _id field
* This is a problem when the name contains an uppercase letter. (The id is lowercased)
* Solution to discuss: I guess we should return users (and principals in general) with as key the _id

Linked issue: https://youtrack.enonic.net/issue/XP-4687
Maybe linked discuss topic: https://discuss.enonic.com/t/changes-to-virtual-host-config-in-6-9-1/919/14
And now linked to Rune problem

GlennRicaud:
Noticed a problem in admin tool 'Users': After successful creation of principal the name field is still editable...

Will create another issue

GlennRicaud:
Solution chosen to be backward compatible is to deduct the NodePath from the PrincipalKey and not the NodeId, and then CRUD the principals on this NodePath

PR #4645, opened 2017-03-20
Principal with uppercase name (#4462)

GlennRicaud:
I still have some tests to do but can you start to look at the code?


GlennRicaud:
Added a test for update by path, and updated some tests with uppercases. If it is okay I will merge it and create another PR for master

Issue #4636, opened 2017-03-17
NPE on import targeting a manually sorted node


GlennRicaud:
NullPointerException on import targeting a manually sorted node

https://discuss.enonic.com/t/data-toolbox-import-export-not-working-for-templates/982

PR #4656, opened 2017-03-20
NPE on import targeting a manually sorted node #4636

runarmyklebust:
Checking if initial folder contains a node definition or not, acts accordingly.
Refactored tests to integration tests since its so hard testing this properly without a real backend.
Fixed issue with NPE on insert in manual ordered parent.


GlennRicaud:
2 corrections for the price of one (thumbsup)

Issue #4400, opened 2017-02-13
Bug when restoring one snapshot, then another


runarmyklebust:
Reproduce:
* Created a repository
* Added branch t1 to the repository
* Took a snapshot
* Added a second branch, t2 to the repository
 * Took a second snapshot
 * Deleted the repository

Then, I tried restoring the snapshots:

     Restoring the first snapshot worked well.  t1 appeared as it should.
         But, if I forwarded to the second snapshot, the t2 branch did not appear.

I tried deleting the branch again:

     Restoring the second snapshot worked well.  Both t1 and t2 appeared as they should
         Then I tried going back one more step by restoring the first snapshot.  t1 was there as it should, but so was t2.  t2 contained 0 nodes, but it was still there.

PR #4667, opened 2017-03-21
Bug when restoring one snapshot, then another (#4400)

runarmyklebust:


GlennRicaud:
I still reproduce the bug described in the issue :S

runarmyklebust:
Stupid last minute refactoring. Fixed now

GlennRicaud:
More proper now with event handling that direct call to services. I like (thumbsup)

Issue #4553, opened 2017-03-06
Call create issue API from the "Create Issue" dialog


alansemenov:
Depends on #4547, #4551 

When the "Create issue" button is clicked in the dialog, call issue API to create an issue. 

If there was an error creating an issue, show the red error notification.

If the issue was created successfully, close the dialog and show blue notification "New issue <issueId> created successfully.".

PR #4678, opened 2017-03-22
Call create issue API from the "Create Issue" dialog (#4553)

vbradnitski:


alansemenov:
A few small things to fix:
1. Don't close the dialog after clicking the "Back" button. If I filled out the fields and went "back" to look at the items, I want to keep the values in the "Create issue" dialog when I go there again.
2. Remove "<" from the Back button
3. There should be no "..." on the "Create Issue" button inside the "New issue" dialog
4. Align right side of the "Create issue" button with right border of the fields above.
5. Automatically set focus on the "Title" field

Issue #4692, opened 2017-03-23
In Publish Wizard hide "Show children" toggler when an item doesn't have any unpublished children


alansemenov:
Add boolean ``hasChildren`` field to object inside ``requestedContents`` of json response from ``resolvePublishContent``. If requested item doesn't have children that can be published (new, offline, modified etc.), hide the "show children" toggler.
```
"requestedContents":[  
      {  
         "id":"features",
         "hasChildren":true
      },
      {  
         "id":"my_folder",
         "hasChildren":false
      }
   ]
```

runarmyklebust:
Dont we use hasChildren in another context as content that actually has children? I bit confusing naming there - maybe it should be 'hasChildrenToPublish'

alansemenov:
Yeah, or ``hasUnpublishedChildren``

PR #4706, opened 2017-03-24
In Publish Wizard hide "Show children" toggler when an item doesn't h…

vbradnitski:
…ave any unpublished children #4692


alansemenov:
Works fine in UI

runarmyklebust:
Ok, I have commited a way to resolve this directly from the node-service with minimal overhead. Please use this, and I still suggest removing this all togheter from the resolvePublishContent resource, move it to a separate method and implement a content-service method that uses the the new method in node-service

PR #4714, opened 2017-03-27
In Publish Wizard hide "Show children" toggler when an item doesn't have any unpublished children #4692

vbradnitski:


alansemenov:
Please move the "main" content items to the left if none of them have the child toggler:

image

The rest is fine, great job!

Issue #4649, opened 2017-03-20
Send email notification to assignee(s) of a new Publishing Issue


alansemenov:
Depends on #4553 

After a new publishing issue was successfully created, an email notification should be sent to selected approver(s).

**From**: email address of a user who created the issue
**To**: email address of an approver
**Subject**: You were assigned to a new issue "``<Issue Title>``" (#``<Issue Id>``)

**Body**:
``<Issue description>``

**[Open Issue button]** (direct link to the "Issue details" dialog) (#4651)

**Issue Items** (``<Total number of items in the issue>``)

Only list items that are stored together with the issue, not references. 

The button should not do anything at this point. Later it will open the Issue Details dialog in the Content Studio. 

![image](https://cloud.githubusercontent.com/assets/11612490/24099200/4d2f4c96-0d6e-11e7-8c6e-bfd0b3819cec.png)

PR #4729, opened 2017-03-28
Send email notification to assignee(s) of a new Publishing Issue #4649

ashklianko:
-requires com.enonic.xp.mail.cfg to be adjusted
-no email styling now


alansemenov:
It does send an email but it looks nowhere close to the sample. I saw you wrote "no email styling now" but that was a mandatory part of this task :). It's ok to use minimal inline styling. 

It's probably difficult to use content type icons (unless there's a tricky way to insert them inline) but we should at least show display name, path and status. Please make it look as close to the sample as possible (except for sender's avatar and item icons).

ashklianko:
OK!

Issue #4701, opened 2017-03-24
In Publish Wizard hide "Include offline" checkbox when there are no offline dependencies


alansemenov:
Add boolean ``containsOffline`` field to the root of the same json response (like ``containsInvalid``). If there are no dependent items in "offline" status, the "Include offline" checkbox should be hidden.

It's important that ``containsOffline`` takes into account ``excludeChildrenIds`` parameter of the request. For example, if I select one Offline item which has an Offline child, but the "Show children" toggler is OFF, then the checkbox must be hidden. But if I toggle the "Show children" toggler ON, then the checkbox must appear.

PR #4732, opened 2017-03-28
In Publish Wizard hide "Include offline" checkbox when there are no o…

vbradnitski:
…ffline dependencies #4701


alansemenov:
@runarmyklebust Discussion with Thomas was about something else - dependencies in Offline status are visible in the Publishing Wizard even when the "Include Offline" checkbox is off. I made a separate issue for that: #4739 

This one is basically to prevent situation when user clicks the "Include Offline" checkbox and nothing happens. In this case we should simply hide the checkbox to avoid confusion.

Issue #4734, opened 2017-03-29
Auth login should attempt to authenticate with user stores in alphabetical order


aro:
The `login()` function of auth-lib specifies for the userStore parameter: 
> userStore: "Name of user-store where the user is stored. If not specified it will try all available user-stores in order."

But it does not enforce any order. They are attempted in the random order they come from the Node query.
Sort the user stores alphabetically by name, before attempting the login in sequence.

PR #4760, opened 2017-04-03
Auth login should attempt to authenticate with user stores in alphabe…

aro:
…tical order (#4734)


aro:
Updated setting child order to "_name asc" on user store parent node.

Issue #4397, opened 2017-02-13
Missing support for index as path in node-API


runarmyklebust:
Add alias "path" and to full settings to index field as path

PR #4770, opened 2017-04-04
Missing support for index as path in node-API (#4397)

runarmyklebust:
Added alias and mappings for path indexing


runarmyklebust:
Hmm, yes, I suppose so, nice catch

Issue #4774, opened 2017-04-04
Remove uneccessary offline-status code from publish-methods


runarmyklebust:
Remove all code present to support the filtering of offline-items from Content-domain and up.

* Remove the extra method in ContentService and de-deprecate the ContentService.resolvePublishDependencies
* All Results with offline stuff
* ...

Leave the initialDiffFilter-param in the ResolveSyncWorkParams, this could be useful anyhow.

PR #4775, opened 2017-04-04
Remove uneccessary offline-status code from publish-methods #4774

vbradnitski:


alansemenov:
Tested - works fine UI-wise

Issue #4542, opened 2017-03-06
Error message appears on the wizard-page, when JPG was replaced with a svg


sgauruseu:
1. Upload a JPG content(New Content Dialog)
2. open the content and press the 'Upload' button on the wizard
3. try to replace the JPG with a SVG-file(see on the first screenshot)

14:40:19.703 ERROR c.e.x.j.i.e.JsonExceptionMapper - The image-based media type image/svg+xml is not supported for writing
java.io.IOException: The image-based media type image/svg+xml is not supported for writing
        at com.enonic.xp.core.impl.image.ImageServiceImpl.getFormatByMimeType(ImageServiceImpl.java:121) ~[na:na]
        at com.enonic.xp.admin.impl.rest.resource.content.ContentImageResource.resolveResponseFromContentImageAttachment(ContentImageResource.java:148) ~[na:na]
        at com.enonic.xp.admin.impl.rest.resource.content.ContentImageResource.getContentImage(ContentImageResource.java
BUG: error notification message appears, see on the second screenshot
![result](https://cloud.githubusercontent.com/assets/3728712/23608631/57fcbac6-027b-11e7-8daa-7957fa591b36.png)

![upload](https://cloud.githubusercontent.com/assets/3728712/23608602/35696248-027b-11e7-977a-9b8c0cba5b01.png)

PR #4787, opened 2017-04-06
Error message appears on the wizard-page, when JPG was replaced with …

ashklianko:
…a svg #4542

-Forbidding to update media content with a content of diffrent media type


ashklianko:
Thanks @runarmyklebust ,will update right away; first thing I've seen was other classes injecting contentService so it confused me. Do we need additional task to refactor other classes where injection of contentService is not required?

PR #4789, opened 2017-04-06
Path-search in selectors doesnt work (4786)

runarmyklebust:
Fixed search expression so you dont get double // in beginning if starting search expression with /


alansemenov:
Beautiful as a Hawaii sunset

Issue #4792, opened 2017-04-07
Applications -  applications not filtered, when one or more applications were checked and 'Show Selection' has been pressed


sgauruseu:
1. open Applications
2. click checkbox and select several applications 
3. click on show selections 'circle'
BUG: applications are not filtered
![app](https://cloud.githubusercontent.com/assets/3728712/24792575/4e0fb2c8-1b87-11e7-8aed-99de3ff5c6a8.png)

PR #4793, opened 2017-04-07
Applications - applications not filtered, when one or more applicatio…

ashklianko:
…ns were checked and 'Show Selection' has been pressed #4792


alansemenov:
Nice refactoring!

Issue #4791, opened 2017-04-07
Users App grid is empty in "Show selection" mode


sgauruseu:
1. Open Users
2. click checkbox and select Roles or  Groups or  System User store
3. click on the 'Show Selections'
BUG: grid is getting empty
![sel](https://cloud.githubusercontent.com/assets/3728712/24792388/68286016-1b86-11e7-8e82-f510c3e14095.png)

PR #4799, opened 2017-04-10
Users App grid is empty in "Show selection" mode #4791

ashklianko:


alansemenov:
There's a bug when unchecking items in the selection mode - it removes more than it should

Issue #4794, opened 2017-04-07
Refactor content-service out of Content commands


runarmyklebust:
Some content commands gets the content-service injected - remove this and use AbstractContentCommand or existing commands directly.

PR #4807, opened 2017-04-11
Refactor content-service out of Content commands #4794

ashklianko:


runarmyklebust:
Lets wait until next release, since we are so late in the release cycle, and this is just sugaring

Issue #4470, opened 2017-02-23
Browser freezes after an image was dragged to the drop zone in the ImageWizard


sgauruseu:
1. open the SuperHero, open the ComponentView and insert an Image
2. open the inspect panel and click the Edit button on it and try to edit the image
3. Press the Upload button and drag an image from the dialog-window and drop it in the ImageWizard
4 Do not close the Upload Dialog and drag and drop another image
5. Do not close modal dialog and  do the step-4  5-7 times 
BUG: browser not responding, need to stop the application in Task Manager
![upload](https://cloud.githubusercontent.com/assets/3728712/23263797/5eab424a-f9f0-11e6-9b6a-f11edcfe0ac7.png)
![not_responding](https://cloud.githubusercontent.com/assets/3728712/23263814/655f2048-f9f0-11e6-8635-e6527b261d0b.png)

PR #4808, opened 2017-04-11
Browser hangs(Not Responding) when try to drag an image to the drop z…

vbradnitski:
…one in the ImageWizard #4470


vbradnitski:
disallow uploading race.

alansemenov:
@vbr Can you change "busy" to "uploading"? Also, "isBusy" to "isUploading", "setBusy(true)" to "startUpload" and "setBusy(false)" to "finishUpload".

Issue #4652, opened 2017-03-20
Implement "Issue List" modal dialog


alansemenov:
Implement new modal dialog that will list publishing issues split in tabs based on issue status. 

* Each tab should show number of issues listed on the tab, except when the number is 0.

* The two first tabs should be:
"Assigned to me" - list of **open** issues where current user is an approver
"My issues" - list of **open** issues created by current user

* Rest of the tabs should be generated dynamically based on status of issues in the system: currently only Open, Closed. Note that issues from the first two tabs will also be shown on tabs of their corresponding statuses.

* If the tab is empty, show centered message "No issues found"

* By default open the first tab that has more than 0 item (ie check "Assigned to me" first, then "My issues" etc.)

* Issues should be listed by modified date/time in descending order.

* The list should contain issue title, issue id, creator name and modified date/time.

* Each issue should be highlighted on hover (lightgrey background, like in the grid)

Add new "Issues" action to the toolbar that will open this new dialog.

![image](https://cloud.githubusercontent.com/assets/11612490/24948336/477fe932-1f6a-11e7-9d00-3a68c0721484.png)

PR #4828, opened 2017-04-17
Implement "Issue List" modal dialog #4652

ashklianko:
-Dialog itself added
-Backend supplies mock data


alansemenov:
Nice! Just a few small things:
1. Add tooltip to the icon that opens the dialog: "Publishing Issues"
2. Add tooltip to the new Issue icon: "Create an issue"
3. Remove the notepad icon and "Open" button from the list - an issue will be opened on click anywhere. Change cursor on hover to "hand".
4. Change dialog title to "Publishing Issues"
5. Message "No issues found" should be 18px big. It also looks like it's incorrectly centered, too far to the right.
6. The dialog doesn't seem to be correctly centered initially - it does when I slightly resize the window.
7. Seems to be a problem displaying issue Id (#[object Object]) but I guess this will be fixed when the data is in place.
8. Add tooltip to the entire issue div that will display issue's description, but only if it's not empty
9. User name (in "Opened by...") should be bold
10. The focus outline (blue border) is not following active tab:
image

ashklianko:
Fixed everything @alansemenov mentioned except .10

ashklianko:
Regarding backend: still lots to do, test to add etc.

alansemenov:
@runarmyklebust this will be done in #4831

Issue #4646, opened 2017-03-20
Alert about unsaved changes appears when closing content that was moved out of a site


sgauruseu:
This issue does not reproducing on FF
1. select the Features and add an Article(or checkbox) content
2. select the article and move it to the root directory
3. open the just moved article and try to close the wizard
BUG: there are no unsaved changes, but Alert dialog is present when the 'close' button was pressed
![article](https://cloud.githubusercontent.com/assets/3728712/24096973/f1b6074a-0d74-11e7-9fb4-d043902a4224.png)

PR #4829, opened 2017-04-17
Alert about unsaved changes appears when closing content that was mov…

edloidas:
…ed out of a site (#4646)

Added additional optional save after content is rendered and loaded.
Content can be moved from its Site, but persisted content will still have metadata from the Site, which will result in the wrong hasUnsavedChanges() return value.

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/wizard/ContentWizardPanel.ts
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/wizard/ContentWizardPanel.ts
@@ -608,6 +608,17 @@ export class ContentWizardPanel extends api.app.wizard.WizardPanel<Content> {
 
             this.setSteps(steps);
 
+            // Content can be moved from its Site, but persisted content can still have metadata
+            // Such content needs update
+            const updateHandler = () => {
+                const needUpdate = !this.getPersistedItem().extraDataEquals(this.getExtraDataFromMetadata());
+                if (needUpdate) {
+                    this.saveChanges();
+                }
+                this.unRendered(updateHandler);
+            };
+            this.onRendered(updateHandler);
+
             return mixins;
         });
     }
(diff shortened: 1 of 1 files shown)


modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -608,6 +608,17 @@ export class ContentWizardPanel extends api.app.wizard.WizardPanel<Content> {
 
             this.setSteps(steps);
 
+            // Content can be moved from its Site, but persisted content can still have metadata
+            // Such content needs update
+            const updateHandler = () => {
+                const needUpdate = !this.getPersistedItem().extraDataEquals(this.getExtraDataFromMetadata());
+                if (needUpdate) {
+                    this.saveChanges();
+                }
```
alansemenov: I'm a bit unsure about this fix... First of all, we have to run this check every time we open any content (while the described case will happen very rarely). Second of all, I don't like that user gets a notification about saved content when he simply opens the wizard.
I understand that we need to clear metadata but can we do it in the backend when we move the content?

edloidas:
@alansemenov Not sure. Better ask backend guys on this @aro @runarmyklebust @GlennRicaud .
Can we update the `Content`, after it was moved from its site (returned by `getNearestSite()`) into another place? You see, when `Content` is moved, the related `ExtraData` set should also be updated, since it contains `MixinName` with `ApplicationKey` of its parent site, that is no longer related to that `Content`.

modules/admin/admin-ui/src/main/resources/web/admin/apps/content-studio/js/app/wizard/ContentWizardPanel.ts
runarmyklebust: I agree @alansemenov. I think we should do this in the content-service.

1) Move the stuff from content-service into a new MoveContentCommand
2) When doing the move, check if its out of site
3) If out of site, do update for content and all children

Issue #4681, opened 2017-03-22
Implement "Issue Details" modal dialog (Read-only mode)


alansemenov:
Depends on #4552 #4652

"Issue Details" will be combination of "Create Issue" dialog and "Publish dialog", so we should do smart refactoring to not duplicate the logic. This dialog can be opened from two places: the issue list and email notification.

**In default read-only mode:**
- all of the fields are readonly
- "Back" button (left-aligned), only if the dialog was opened from the issue list
- "Publish" button (right-aligned), only if there's at least one content item in the issue (#4685)
- "Pencil" icon in the top right corner to switch to the "edit" mode (#4693)
- "Close issue on publish" checkbox next to the "Publish" button
- "Schedule" icon like in the Publishing dialog - make sure that "Back" will take user from "Schedule" dialog back to the Issue dialog 

The list of content items should reuse the logic from Publishing wizard, with a few changes:
 - show main items from the issue
- toggle "Include children" toggler on or off based on what's saved in the issue
- automatically fetch dependencies of the main items, but exclude those that are excluded from the issue
- no X icons
- "include children" togglers are not clickable

![image](https://cloud.githubusercontent.com/assets/11612490/24253304/3beeafc0-0fe0-11e7-92c5-282a79de1f70.png)

![image](https://cloud.githubusercontent.com/assets/11612490/24253325/4cdbf31a-0fe0-11e7-9b44-c5f1d626acd1.png)

PR #4879, opened 2017-04-27
Implement "Issue Details" modal dialog (Read-only mode) #4681

vbradnitski:


alansemenov:
Some things to fix:
1. Big bug: when I select a single item and click "Publish" (not "Publish tree"), all of its child items are fetched and shown in the dialog (even though the "include children" toggler is off). If I create an issue, it will save it with "Include children" toggler on.
2. There should be text right under the title saying "ZZZ opened this issue on YYY", just like in the issue list
3. Set "text-decoration: none" for the pencil icon
4. In readonly mode hide textarea and label of description, and show <p> element with description so that its height is based on how long the description is. Use color: @admin-font-gray2;
and font-size: 14px.
5. Remove all unnecessary margins and paddings inside input-views. There should be 10px above and below the "ZZZ opened this issue..." line, 15px above and below description and margin-bottom 15px for the form.
6. Left-align "Back" button inside the Schedule dialog
7. Use font-size: 14px for "Assignees" header

Issue #4831, opened 2017-04-18
Implement backend services for retrieving Publishing issues list


ashklianko:
When implementing #4652 appeared that there's no required backend services required for fetching issues by type. Do next:

-Create IssueQuery class analogous to ContentQuery, PrincipalQuery etc.
-Create IssueNodeQueryTranslator  for IssueQuery->NodeQuery
-Add IssueService findIssues and create it's implementation
-Refactor Issue Command classes to inherit from common abstract parent holding nodeService
-Add method in ContentResource to fetch issues by type (assigned to me, created by me, by status)
-Add tests for everything

PR #4883, opened 2017-04-27
Implement backend services for retrieving Publishing issues list #4831

ashklianko:


ashklianko:
Thanks @runarmyklebust , will update

ashklianko:
@runarmyklebust please check updated code

Issue #4825, opened 2017-04-14
ImageSelector wizard - page is not refreshed when selected image was changed in the new wizard-tab 


sgauruseu:
1. Start to add an ImageSelector content
2. select an image and click on it('edit' button is getting displayed)
3. click on the Edit button and change the image-content(upload another image) and save it
4. switch to the  ImageSelector content again
BUG: the previous image still displayed on the  ImageSelector -wizard
when the wizard page has been refreshed, new image is displaed

alansemenov:
This bug is in fact a bit wider than what's described. Besides updating a selected image we should also:

1. Update display name/path of selected content item(s) in a Content Selector

![image](https://cloud.githubusercontent.com/assets/11612490/25169896/eeff3144-24e8-11e7-88d8-cd15a31106aa.png)

![image](https://cloud.githubusercontent.com/assets/11612490/25169975/34850298-24e9-11e7-8391-8b848e5c534b.png)

2. Update thumbnail, display name and path inside the selector's dropdown. It does work for Content Selector but not for Image Selector, for some reason.

pmi:
Updated the PR

PR #4886, opened 2017-04-28
ImageSelector wizard - page is not refreshed when selected image was …

pmi:
…changed in the new wizard-tab #4825


alansemenov:
Nice refactoring, @pmi!

Issue #4878, opened 2017-04-27
HTML Area won't scroll inside the Inspect panel


alansemenov:
1. Add an HTML Area to the "cities-list" part config inside the Features app.
2. Deploy the app
3. Create or edit a page inside the Features app in Content Studio
4. Drag the "Cities List" part onto the page and select it to open the Inspect panel
5. Find the HTML Area inside the config, open Source mode (icon in the bottom right corner) and paste some long HTML there
6. Close the Source mode
7. Now hover cursor over the HTML Area and try to scroll up or down. It won't scroll.

It scrolls fine when cursor is outside of the HTML Area. It also scrolls fine when the same content is pasted into HTML Area inside a text component.

PR #4903, opened 2017-05-03
HTML Area won't scroll inside the Inspect panel #4878

pmi:


alansemenov:
Scrolling works fine now, but there's an issue with toolbar after resizing of the Inspect panel - it seems to always keep the initial width.

Before resize (ok):

image


After resize (not ok):
image

alansemenov:
Actually, there are more serious issues... Seems like HTML Area is ALWAYS in inline mode now, also inside the wizard. The idea was to switch it to inline mode only inside the Inspect panel to solve the scrolling issue. Now source/fullscreen icons in the HTML Area's footer are gone, and there are issues on resize too. That's what I get in the wizard after resize, for example:

image

If we can't avoid the changes to affect the wizard then we'll have to roll them back.

Issue #4891, opened 2017-05-02
Disable "Create Issue" button when all items are published


alansemenov:
When all of the items are published, the Create Issue button is still enabled. We should disable it.

![image](https://cloud.githubusercontent.com/assets/11612490/25614302/b6e631ce-2f31-11e7-865c-410c6ecd5564.png)

edloidas:
The fix was included (committed) as part of the [PR for #4833 task](https://github.com/enonic/xp/pull/4905).

Issue #4833, opened 2017-04-18
Always allow creating a Publishing Issue from the Publishing dialog


alansemenov:
The way it's done now, if user doesn't have enough rights to publish at least one item in the Publishing dialog, we replace the Publish button with "Create Issue" button:

![image](https://cloud.githubusercontent.com/assets/11612490/25128462/24da01b8-243a-11e7-8b1e-5ab9b31114ad.png)

otherwise we show the Publish button.

We should allow both options (Publish and Create Issue) when user has enough rights to publish all items. Make sure that in this case we show default texts instead of "Create a new Publishing Issue with selected item(s)" and "Other items that will be added to the Publishing Issue".

What we need is a MenuButton that will have "Publish (X)" as main action and "Create Issue... (X)" in the dropdown:

![image](https://cloud.githubusercontent.com/assets/11612490/25129108/e8ed9e78-243c-11e7-991f-0fc13bd575b9.png)

If user doesn't have sufficient right then we will show only the "Create Issue..." button, like now.

PR #4905, opened 2017-05-04
Always allow creating a Publishing Issue from the Publishing dialog #4833

edloidas:
* Refactored `ModelDialog` and it's descendants to use config, instead of multiple properties.
* Updated `ModalDialogButtonRow` to be able to extend ModalDialog with ease.
* Added `ContentPublishDialogButtonRow` with publish `MenuButton`.
* Disable "Create Issue" button when all items are published #4891: Fixed the way the buttons are enabled and visible in `ContentPublishDialog`.
* Fixed `ResponsiveRange`: it was not set on Dialog before any `resizeEvent `call.
* Fixed wrong input margin in Create Issue dialog, since ResponsiveRange was not set.


alansemenov:
1. Both buttons are visible at the same time for me - that should not happen
image

Tried under different users (both with and without publish rights) in Chrome and FF.

2. Number of items should be shown in brackets after "Create Issue..." inside the dropdown, just like on the button

3. There's a linter error in ContentPublishDialog.updateButtonCount - you don't need "count" as an argument of labelWithNumber since it's already accessible inside the method body.

4. Change width of the dropdown button so that it's square (32x32), now it's (40x32)

5. Hide the dropdown button when it's disabled

image

Issue #4918, opened 2017-05-09
Issues assigned to a user don't show up on his "Assigned to me" tab in the issue list


alansemenov:
The title says it all - issues assigned to a user will never show up on the "Assigned to me" tab, basically it's always empty.

PR #4923, opened 2017-05-10
Issues assigned to a user don't show up on his "Assigned to me" tab i…

ashklianko:
…n the issue list #4918

-By some reason approvers field was set to be not indexed


ashklianko:
Note that only issues created after merging this code will be shown in 'assigned to me' tab.

Issue #4654, opened 2017-03-20
Add content item selector to the Create Issue/Issue Details dialog


alansemenov:
Depends on #4681 

* When user creates a new issue, he should be able to add content items to it right away.
* When user opens an existing issue, he should be able to add or remove items in the issue

Add a content selector to the dialog, above the button toolbar. What we need is something like LoaderComboBox in the Move dialog, but with multi-select option (checkboxes next to content items).

Selected items will be shown under the textbox with "X" icons to remove from selection.

* For a new issue, we should pre-select items that are already selected in the grid (if any)
* For an existing issue, we should hide the selector and show only the list of items saved with the issue. 
* When the dialog is switched to "edit" mode, then both selector and the selection should be shown.

![image](https://cloud.githubusercontent.com/assets/11612490/24254899/51679c96-0fe4-11e7-910c-633c4efe4bac.png)

PR #4932, opened 2017-05-12
Add content item selector to the Create Issue/Issue Details dialog #4654

vbradnitski:


alansemenov:
Everything works great, but there are several UI things we need to fix here
1) We need an option to remove selected items directly from the list by clicking X icon:
image

2) Too much space above the item selector (the "Invite users..." form item above should have margin-bottom: 0):
image

3) Need a header above the item selector in the CreateIssue dialog: "Items that will be added to the issue" (also when I manually create an issue). For an existing issue it's ok that it simply says "Items".

alansemenov:
Still some issues left:

1. "contans-toggleable" style doesn't seem to get assigned correctly in the "Create Issue" dialog. As a result, items without the toggler icon don't get correct margin-left:
image

2. Header "Items that will be added to the issue" is still missing when I create an issue from the "Publishing Wizard" (but it's shown above the item selector when creating an issue manually):

image

3. If I select several items in the grid and one of them doesn't have any unpublished children (which means it doesn't have toggler icon in the Publishing Wizard) then I get a bunch of console errors in the browser when trying to click "Create Issue":

image

Issue #4926, opened 2017-05-11
Implement auto-incremental numbering of issues


alansemenov:
While it's ok to use uuid as unique id of an issue, it's hardly readable so we should add a new field which will store issue number that will be shown to endusers in emails and issue detail dialog.

The way it should work is when a new issue is created, we should retrieve max value of this field from all existing issues and add 1 to it. If no issues exist, or the field is empty for all existing issues, use 1.

Use value from the new field everywhere where we currently show issue Id:
- issue list 
- title of the existing issue dialog 
- emails 

Also, see if we can get rid of the issue path field - I don't see any use for it.

PR #4939, opened 2017-05-15
Implement auto-incremental numbering of issues #4926

ashklianko:
-added new field 'index' for numbering issues
-removed 'path' from issue as not used


ashklianko:
Need to clean repo prior to testing.

alansemenov:
Nice!

Issue #4685, opened 2017-03-22
Implement closing the issue after publishing from the Issue dialog


alansemenov:
Depends on #4681 

Clicking the "Publish" button inside the "Issue Details" dialog should publish selected items exactly like it does in the Publish dialog.

If "Close issue on publish" checkbox is "ON", update issue status to "Closed" (as well as modified by/modified date fields) after publishing.

Show notification: "Issue ``"<Issue Title>"`` is closed".

![image](https://cloud.githubusercontent.com/assets/11612490/24255060/c59b58be-0fe4-11e7-80ad-88a3f0b17a90.png)

PR #4941, opened 2017-05-15
Implement closing the issue after publishing from the Issue dialog #4685

edloidas:
* Changed success message for published issue.
* Added update of the issue lists after the issue is (closed).
* Fixed `.creator` styles in issue dialog.
* Fixed `IssueList.reload()` not removing actual items.

### DIFF
--- a/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/issue/IssueResource.java
+++ b/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/issue/IssueResource.java
@@ -147,13 +147,13 @@ public final class IssueResource
         if ( type.equalsIgnoreCase( "CREATED_BY_ME" ) )
         {
             final AuthenticationInfo authInfo = ContextAccessor.current().getAuthInfo();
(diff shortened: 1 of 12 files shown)


alansemenov:
@edloidas 

1. I ALWAYS get a notification saying "Issue was closed", even when I didn't check off the checkbox and the issue was not in fact closed. We should give a notification only *if* the issue was closed. It should also be a info-notification, the blue one.

2. I didn't understand this part: "Take a note, that issue won't be moved to Closed". This is the main part of this task. If it's not done, then we need to add that. We do have REST APIs for issue update (#4547), why can't we use that for status update?

edloidas:
@alansemenov Implemented frontend part of the REST API, fixed `IssueList` reload, and fixed all issues mentioned above.

alansemenov:
Works as a charm, great job! Two small things:
1. After the issue is closed, it still stays on the "My issues" tab (while being listed on the "Closed" tab at the same time). It might easily be the bug of the issue list, but please fix that as well.
2. After we publish an issue with "Close..." checkbox ON, we should redirect to the "Closed" tab of the issue list.

Issue #4938, opened 2017-05-12
Dynamically hide/show notification icon on issue changes


alansemenov:
Depends on #4928 #4685

We now show the blue notification icon over the "Open Issue List" button in the toolbar when there are issues assigned to current user. 

1. Dynamically hide this icon after the last issue assigned to current user is closed (when he published it with "Close issue after publish" checkbox - see #4685)
2. Dynamically show this icon if there were originally no issues assigned to current user and he created a new issue/updated existing issue and assigned it to himself.
3. Change the button's tooltip accordingly when showing/hiding the notification icon.

In 1 we should listen to issue status change event (create one if we don't have it yet). Issue can be closed after publish but also manually.
In 2 we should listen to issue create/update events.

ashklianko:
@ase, If we close issue, is it still supposed to be shown in 'Assigned to me' tab? If no then need a separate task to fix this.

alansemenov:
@ashklianko No, that was fixed by @edloidas in another task and should already be merged in #4550

PR #4958, opened 2017-05-18
Dynamically hide/show notification icon on issue changes #4938

ashklianko:


alansemenov:
Great job with the event hub class. I'm merging this, but please make a note to yourself about getting rid of ``setTimeout`` in ``ShowIssuesDialogButton.ts``. We will have to handle issue events in many different places and we can't rely on timeouts all the time. Issue events should be fired after all the backend updates are finished.

Issue #4693, opened 2017-03-23
Support edit mode in the "Issue Details" modal dialog


alansemenov:
Depends on #4681, #4654

When user clicks the Pencil icon in the top right corner, the issue dialog goes into Edit mode.

- All fields are editable
- Content Selector control is shown above the item list
- The list of main selected items must be in sync with the dropdown (change of selection inside the dropdown must refresh the list below and vice versa)
- Two buttons at the bottom: Save and Cancel
- Cancel will simply close the edit mode
- Save will update the issue and close the edit mode
- Validation of mandatory fields (title and assignees)

![image](https://cloud.githubusercontent.com/assets/11612490/24253756/7959bc0a-0fe1-11e7-9c1f-db8571571e70.png)

alansemenov:
Note: Save and Cancel should not be like in the screenshot, but switched places, like in all our dialogs

PR #4964, opened 2017-05-19
Support edit mode in the "Issue Details" modal dialog #4693

vbradnitski:


alansemenov:
Some things that need to be fixed:
1. Hide the entire header div when the issue is in edit mode
2. "X" icon in the assignee list should be vertically aligned with the dropdown's handle
image
3. After I updated the issue, its ID changed to 0:
image
4. The buttons should be "Save" and "Cancel" (like in screenshot in the task's description, but Save to the left and Cancel to the right)
5. There is still too much space above the "Items" label, but maybe that was fixed in another branch?
image
6. When I removed all items from the issue and saved, the "Items" label remained visible:
image
7. And the opposite issue - after I added items to an issue that didn't have any, the "Items" label didn't appear:
image

ashklianko:
I fixed thing with issue number broken in https://github.com/enonic/xp/pull/4967

That was about 'index' field incorrectly updated during update operation.

alansemenov:
@ashklianko :+1:

Issue #4961, opened 2017-05-18
Refresh the issue list dialog on issue server events


alansemenov:
Farther improvement of #4938

We should update active (open) dialog with issue list upon receiving issue event from the server - whether a new issue was created or an existing one updated.
Mask the dialog and show the spinner while updating the dialog contents. Make sure user stays on the same tab.
Show info notification (the blue one): "The list of issues was updated" (only if the dialog is currently open).
Would be nice to add debounce of this event handling so that we don't update/notify more than once every 3 seconds, for example.

ashklianko:
@alansemenov Please also check that #49825 is gone

PR #4985, opened 2017-05-23
Update the issue list dialog on issue server events #4961

ashklianko:


alansemenov:
Great job, @ashklianko !  See #4994 for an improvement on user notification.

Issue #4914, opened 2017-05-09
Site.xml with only one input type with foldout doesn't show properly


Bellfalasch:
Tested in 6.9 and 6.10, both fail.
Tested in Chrome and Firefox, both fail.

Add the Favicon app from market. This is an app that only has one input type in its site.xml file. This is a ImageSelctor, but the error would also occur if it was a ContentSelector or a ComboBox.

When searching, and/or, clicking the drop down arrow, the dropdown list is displayed above and behind the actual field. It is also very small, showing only half an item. I'd expect the drop down list to display over the edges of the site.xml modal so that it can be used.

See attached screenshots of the error.

<img width="786" alt="screen shot 2017-05-08 at 19 43 54" src="https://cloud.githubusercontent.com/assets/1465598/25838228/60200674-3492-11e7-8ce2-185ea76a69cd.png">

![screen shot 2017-05-09 at 08 29 52](https://cloud.githubusercontent.com/assets/1465598/25838235/68af06d2-3492-11e7-9431-ec4e136ed728.png)

PR #4995, opened 2017-05-24
Site.xml with only one input type with foldout doesn't show properly #4914

edloidas:
Fixed problem with `overflow: auto`, when it prevented dropdown to be rendered outside the parent element in small dialogs without scrolls.


alansemenov:
Hmm... Why do we need such a complicated fix? For me it works fine when I simply remove ``overflow-y: auto;`` from the ``.dialog-content`` selector.

edloidas:
It only works fine, if the content in the dialog has a height less than the maximum height of the dialog itself. Try open the Superhero site and open the SiteConfigurator by editing Superhero theme. The sole purpose of adding `overflow-y: auto;` was exactly that case.

alansemenov:
Good point - seems like a stable fix.

Issue #4987, opened 2017-05-23
Use inline SVG in content list inside issue email


alansemenov:
When resolving content items for issue email, check if the item's content icon is an SVG file (like our built-in content types). If it is, add inline SVG to email content. If it's not (if it's a remote image), just leave the placeholder empty.

I tried it manually and it worked fine:
![image](https://cloud.githubusercontent.com/assets/11612490/26353270/df82ee62-3fbf-11e7-9136-220482bb4b70.png)

This is the code I used:
```
<div style="display: flex; align-items: center;">
    <div style="flex-basis: 34px; margin-right: 15px; color: initial">
        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 32"><path d="m0 10h32l-2 20h-28l-2-20m29-4l1 2h-28l2-4h11l1 2h13" fill="#53585f"/></svg>
    </div>
    <div style="padding: 4px 8px; width: 100%;">

        <h6 style="height: 18px;
					line-height: 18px;
					margin: 0;
					padding: 0;
					font-size: 14px;
					white-space: pre;
					text-overflow: ellipsis;
					font-weight: normal;
					overflow: hidden;">Features</h6>
        <p style="color: #777;
					font-size: 12px;
					margin: 0;
					padding: 0;
					line-height: normal;
					font-weight: normal;
					overflow: hidden;
					white-space: pre;
					text-overflow: ellipsis;">/features</p>
    </div>
    <div style="flex-basis: 80px; padding-top: 10px; color: initial">Offline</div>
</div>
<div style="display: flex; align-items: center;">

    <div style="flex-basis: 34px; margin-right: 15px; color: initial"></div>

    <div style="padding: 4px 8px; width: 100%;">
        <h6 style="height: 18px;
					line-height: 18px;
					margin: 0;
					padding: 0;
					font-size: 14px;
					white-space: pre;
					text-overflow: ellipsis;
					font-weight: normal;
					overflow: hidden;">Mindful Monsters</h6>
        <p style="color: #777;
					font-size: 12px;
					margin: 0;
					padding: 0;
					line-height: normal;
					font-weight: normal;
					overflow: hidden;
					white-space: pre;
					text-overflow: ellipsis;">/scope</p>
    </div>
    <div style="flex-basis: 80px; padding-top: 10px; color: initial">Modified</div>
</div>
```

PR #5000, opened 2017-05-24
Use inline SVG in content list inside issue email #4987

ashklianko:
-Added content type svg icon (if available) for publish content item email representation


alansemenov:
@ashklianko It works fine, but we need tests for the backend code. Current coverage is only 38.46%.

Issue #5016, opened 2017-05-26
Replace scheduler icon with an action in the Publish menu button


alansemenov:
We now show calendar icon to the left of the Publish icon in the Publishing wizard and in the Issue dialog. Instead, add "Schedule..." action to the Publish menu button's dropdown and remove the icon. Logic-wise it will work exactly the same as now.

Note that in the Issue dialog we are not using menu button now, so we will have to add that. There we will have only Publish and Schedule actions of course, not "Create Issue".

![image](https://cloud.githubusercontent.com/assets/11612490/26497200/dd39a43a-422a-11e7-980a-dc711562670c.png)

PR #5029, opened 2017-05-30
Replace scheduler icon with an action in the Publish menu button #5016

vbradnitski:


alansemenov:
All fine, except for two things (one small, one big):

* We should show number in brackets after "Schedule..." option, like we do for "Create Issue" for example
* There are conditions under which the calendar icon was hidden. For example, if all selected items are already Online, the calendar icon is hidden. We should use the same conditions to hide the "Schedule..." option in the dropdown

Issue #5037, opened 2017-05-31
Replace "Assigned to Me" and "My Issues" tabs with filter on the Open/Closed tabs


alansemenov:
* Remove Assigned to Me and My Issues tabs in the Issue List
* Implement filter above the list of issues on both the "Open" and "Closed" tab
* The filter will for now have two checkboxes: "Assigned to Me" and "My Issues" (both unchecked by default)
* After either of the checkboxes clicked, the list should automatically be filtered according to the states of both checkboxes'
* If the filter produced no matches, show "No issues found" message
* NB! If the are originally no issues on the tab, the filter should be hidden (just "No issues found" message)
* Logic of showing issue subtitles ("Assigned by... " etc) should remain the same as it was on "Open" and "Closed" tabs

PR #5047, opened 2017-06-02
Replace "Assigned to Me" and "My Issues" tabs with filter on the Open…

ashklianko:
…/Closed tabs #5037

-also updated IssueDetails dialog to now show issue id in status string


alansemenov:
1. Checkboxes should be inside the tab (but not inside the scrollable area!) and on one line, as shown below (note alignment and padding)

2. Show number of items in parenthesis, but not when 0. It's important that the number doesn't change when the filter applied. If there are originally 3 items under "My issues", it should always show "(3)", even when I apply "Assigned to me" filter.

3. It looks like we disable checkboxes only after filtering. When there's originally nothing under "My issues", the checkbox should be disabled right away and should not be clickable.

image

Issue #5046, opened 2017-06-02
Setting manual order value doesnt mark children as modified for publish


runarmyklebust:
When switching to manual order, the children must be published because the order value is set on every child.

This is not detected when resolving publish work.

PR #5052, opened 2017-06-06
Setting manual order value doesnt mark children as modified for publi…

runarmyklebust:
…sh (#5046)

Fixed missing timestamp on set childOrder. Added test for this bug


GlennRicaud:
Marvelous

Issue #4915, opened 2017-05-09
Ask for confirmation when trying to close modal dialog with changed site configuration


Bellfalasch:
Some apps have a lot of settings for their site.xml. But if you edit these and by accident click outside the modal (on the black overlay background), the modal is closed without warning and all your changes are lost.

Prompt me with a warning before closing the modal! Also make sure the same happens when clicking any other close or cancel button. The user needs an extra chance to stop changes from being lost. The same way closing the edit of any other content works.

![screen shot 2017-04-28 at 08 46 38](https://cloud.githubusercontent.com/assets/1465598/25838329/da1f0092-3492-11e7-9552-38983a984d6b.png)
Click on the dark area and all changes are lost and the modal is closed. Not optimal UX.

alansemenov:
In onclick listener check if the form inside the dialog is "dirty". If it is, show the following confirmation message (similar to SaveBeforeCloseDialog): 

*You have made changes to the form, do you want to apply them?*

The dialog itself should be masked when the confirmation is visible.

Also, important - make this check a setting ("off" by default) inside the ModalDialog class so that we could turn it on for any modal dialog we want. Confirmation message should also be configurable, but the default text should be as stated above.

PR #5063, opened 2017-06-07
Don't close site.xml modal without warning, if something is not saved #4915

edloidas:
Added ConfirmationDialog instance and config to ModalDialog.
Confirmation check is `off` by default.
Implemented `isDirty()` check in Forms.
Added `isDirty()` check to `ModalDialog` before opening confirmation dialog.


alansemenov:
A few things to fix:
1. There should be no confirmation when user explicitly clicks the Cancel button. Now he gets the same confirmation as when clicking outside the dialog.
2. It seems that we don't remove ConfirmationDialog element from DOM after it's closed, so the elements are just sitting there, unused. We should remove them.
3. It would be great if clicking the "X" icon inside the confirmation dialog would simply close the confirmation but not the modal dialog that opened it. As of now there's no way to return from confirmation to the dialog.
4. Apply a border around the masked modal dialog when there's a confirmation over it:
image

alansemenov:
Great job!

Issue #5064, opened 2017-06-07
repositoryService.deleteBranch does not delete recursivly


runarmyklebust:
nodeStorageService only deletes the node it's told to delete, so only the root-node is deleted when doing the repositoryService.deleteBranch

PR #5066, opened 2017-06-07
repositoryService.deleteBranch does not delete recursivly #5064

runarmyklebust:
Using DeleteByIdCommand to delete rootNode with children


GlennRicaud:
Merged now due to intense pressure of rmy. Will check later

GlennRicaud:
Tested and reviewed. Good work!

Issue #5061, opened 2017-06-07
Implement viewer class for rendering principal in a circle with initials


alansemenov:
Implement a new class in ``api.ui.security.PrincipalViewer`` that will render a principal in the form of a circle with initials inside. Later on this will be improved to render avatar instead of the circle. The circle should have a tooltip with full principal name. 

Call the new class ``PrincipalViewerCompact``.

Use a CSS class so that we could change the circle color/size etc. via CSS.

Change ``UserAccessListItemView`` to use this new class when it renders a line of circles for principals (inside the ``doLayout`` method). 

![image](https://user-images.githubusercontent.com/11612490/26867322-2df92ccc-4b65-11e7-8359-c6530f82ed86.png)

It will also be used in the compact version of ``PrincipalSelectedOptionView`` (#5058)

Issue #5059, opened 2017-06-06
Issue details dialog - switch PrincipalComboBox to compact version


alansemenov:
Depends on #5058

![image](https://user-images.githubusercontent.com/11612490/26837351-8763bfc6-4add-11e7-9094-6c8aea625d01.png)

Issue #5058, opened 2017-06-06
Implement compact version of PrincipalSelectedOptionView


alansemenov:
Depends on #5061 

Implement a new class in PrincipalComboBox for displaying selected users in the same way we do inside the UserAccessWidgetItemView in the Details Panel. It should use the viewer class implemented in #5061 for rendering.

![image](https://user-images.githubusercontent.com/11612490/26837067-bfaa2696-4adc-11e7-84a0-8666b4f9be49.png)

![image](https://user-images.githubusercontent.com/11612490/26837039-b20b74fe-4adc-11e7-8808-4b857e46000b.png)

Tooltip of each circle should show full display name of the principal, just like in the widget.

This view will be for now used only inside the issue details dialog, so default class for displaying selected principals should still be PrincipalSelectedOptionView.

PR #5070, opened 2017-06-07
Implement viewer class for rendering principal in a circle with initi…

ashklianko:
…als #5061


ashklianko:
Also includes #5059 and #5058

alansemenov:
Very good

Issue #5068, opened 2017-06-07
Support headers in attachments in mail lib


aro:
Support setting specific headers for the email attachments.

This is necessary for embedding images in the email. We need to be able to set the "Content-ID" in the attachment multipart.

https://tools.ietf.org/html/rfc2392

PR #5074, opened 2017-06-08
Support headers in attachments in mail lib (#5068)

aro:


GlennRicaud:
Did not try, but code looks good

Issue #5065, opened 2017-06-07
Http Library - Temporary file should be deleted after use


GlennRicaud:
Temporary files created in ResponseMapper::getResponseBodyStream are deleted on exit.
But servers are supposed to work without restarting so the files keep stacking up.

Force the deletion of the files after they have been used

I have already worked on something similar for data-toolbox if it helps: https://github.com/GlennRicaud/data-toolbox-app/tree/master/src/main/java/systems/rcd/enonic/datatoolbox

GlennRicaud:
I leave it in progress. Until report in master also

GlennRicaud:
The decision was to create a new external library.
This has been done: https://github.com/enonic/lib-http-client

Since this has been done in 6.10, I think it should merged in master also. Will close this task when it is done

PR #5076, opened 2017-06-08
Http Library - Temporary file should be deleted after use (#5065)

aro:


jsi:
Why is the code coverage dropping here?  Important not to miss this opportunity to add unit tests if at all possible!

GlennRicaud:
@aro Could you report the correction for 6.10 also and do a push request.

GlennRicaud:
And it actually does not work since the stream is not necessarily opened and closed :S

aro:
This task was still in progress. Because I'm not sure if it's ok to put the temporary files in $XP_HOME/temp/ . In principle I thought the work directory was supposed to be for that kind of use. But apparently since we put pre-processed cached images there, the customers want to keep it when they make backups. So I thought about creating a directory only for temporary files. This should probably be discussed before merging it in master.

@gri The temporary file will be created only when it is a big file (> 1.000.000 bytes). Normally when some code is fetching such a file, it will do something with it and so it will be opened and closed. If it is not open and closed, we still have the previous handling where a normal JVM shutdown will remove the files.

To improve this we need to make some kind of background task to periodically check and delete old files from the temp directory (this is a reason for putting them now in XP_HOME/temp instead of the OS temp directory). But this background task is hard to do inside the library, because then every app using the http-client lib will start its own background task for deleting files.

PR #5084, opened 2017-06-09
Http Library - Temporary file should be deleted after use (#5065)

aro:


GlennRicaud:
Amazing solution! Congrats to the brilliant mind behind the idea and cheers to the perfection of the implementation

Issue #5096, opened 2017-06-14
Open issue details dialog after creating a new issue


alansemenov:
After a new issue is created we open the issue list on the "Open" tab. Change it to open the issue details dialog of the new issue.

PR #5101, opened 2017-06-15
Open issue details dialog after creating a new issue #5096

ashklianko:


alansemenov:
I'm getting this after I create an issue from the Issue List dialog. We should apply "masked" CSS class to the underlying dialog to hide it, like we do when creating an issue from the Publish dialog.

image

ashklianko:
On it.

Issue #5153, opened 2017-06-21
Implement imagePlaceholder in portal lib


aro:
It is already implemented as a view function within  portal namespace:
http://xp.readthedocs.io/en/stable/reference/view-functions/imagePlaceholder.html

But was forgotten from portal-lib. And has been requested in Discuss:

https://discuss.enonic.com/t/imageplaceholder-only-a-view-function/1029

aro:
Added test to Features app, in js-libraries-portal part

PR #5154, opened 2017-06-21
Implement imagePlaceholder in portal lib (#5153)

aro:


GlennRicaud:
Did not test but code looks good

PR #5177, opened 2017-06-22
Dump and load versions (#4998)

runarmyklebust:


GlennRicaud:
Will merge. While checking, discovered some bugs:
  * Should ignore files like .DS_Store
  * After importing of a 6.11 dump, the version history seems wrong. The online step is not displayed ( After test with rmy, we saw duplicated contents)
  * Has been there forever, but for a future feature: Applications

Issue #5136, opened 2017-06-20
Page Component View - Don't expand layout after DnD


alansemenov:
We expand a collapsed layout after it was dragged'n'dropped in the Page ComponentView. Fix it so that it doesn't happen - it's much easier to drag collapsed layouts.

PR #5193, opened 2017-06-23
Page Component View - Don't expand layout after DnD #5136

edloidas:
Updated DnD, so that items won't expand after the drop.


jsi:
Wow!  This merge will bring test coverage abover 78%!!!  Another milestone on our way to 80%.  Nice job!

Issue #5120, opened 2017-06-19
Don't close 'Edit Permissions' dialog without warning, if something is not saved


sgauruseu:
1. click on the 'Edit Permissions' link (details  panel)
2. change permissions
3. click on the Close button
Expected:  "confirm" dialog should be present if something was not saved

alansemenov:
Same as with SiteConfigurator dialog - show confirmation when clicking either "X" icon or outside of the dialog.

sgauruseu:
verified on the my local machine

PR #5202, opened 2017-06-26
Don't close 'Edit Permissions' dialog without warning, if something is not saved #5120

edloidas:
Added confirmation dialog to the `EditPermissionsDialog`.

Moved `isDirty` function to `AppHelper`.
Fixed focus blinking in dialogs with a confirmation dialog, when it was shown.


edloidas:
Unpredictable dialog and mask order broke the complex CSS selector for hiding main dialog `X` icon. Fixed.

alansemenov:
Perfect!

Issue #4933, opened 2017-05-12
Switch content selector to default tree-structure


alansemenov:
Everybody loves tree-structure inside content selector in the Move dialog, so we should make this default behaviour.

NB! Since content selector can have different kinds of filtering applied to it (``allowPath``, ``allowContentType``), we should make sure that we always show nodes that have children, but don't let them be selected (don't render checkbox for multi-select and don't handle the click for single-select) if they are not allowed by filters. Nodes that don't have children and are not allowed by filters should not be shown in the tree.

To enable having current "flat" structure, implement support of a config option ``flat``:

```xml
<input name="name" type="ContentSelector">
  <label>Cited In</label>
  <occurrences minimum="0" maximum="0"/>
  <config>
      <flat>true</flat>
  </config>
</input>
```

Bellfalasch:
+1 for also adding this to Image Selector, equally important to have it there too. It's almost hopeless to text-search in 1000+ item archives of images and content.

alansemenov:
@Bellfalasch It's a bit more complicated for image selector since there we should make sure that parent nodes are not selectable and add deep filtering of nodes that don't have any image content. But we'll get there.

Bellfalasch:
Doh, worth a shot =) At least having it added to ContentSelector is good progress. As you saw on meetup, both of these are highly requested by users. @alansemenov

PR #5206, opened 2017-06-26
Switch content selector to default tree-structure #4933

vbradnitski:


alansemenov:
Fantastic job, Slava!

Small issues:
1. When I expand the content selector in the Create Issue dialog, the dropdown is supposed to expand above the selector, but as seen on the screenshot below it's expanded under and is not visible:

image

That's how it used to be:

image

2. Small styling issue - display content name on highlighted row must be in white, not gray:
image

3. Selected content items must follow the same order as they were in the selector. For example, in the screenshot above `Large tree node 2` was before `Large tree node 2-100`, and they should be listed in the same order in the list of selected options:
image

4. For some reason the order of items inside Content Selector in "All input types" content type is not the same as in the grid. But at the same time in the Content Selector in the "Create Issue" it's correct

Content Grid:
image

Content Selector in the Create Issue dialog (same order):
image

Content Selector in the "All input types" content type in Features app (wrong order):
image

vbradnitski:
fixed.

alansemenov:
Two more issues:
1. Selection (checked state of checkboxes) gets lost when collapsing/expanding nodes in the tree (unless items were already selected when the dropdown was opened)
2. In the "Create Issue" dialog order of selected items is the opposite to what it should be (correct in "All input types"). We should still put a new batch of selected items on top, but the order inside a batch must be the same as in the tree.

alansemenov:
This is still wrong:

> In the "Create Issue" dialog order of selected items is the opposite to what it should be (correct in "All input types"). We should still put a new batch of selected items on top, but the order inside a batch must be the same as in the tree.

aro:
@vbradnitski  The logic of the code in `ContentResource.treeSelectorQuery` is a bit hard to follow. What exactly is the filter doing with the paths?

I think the new method in `ContentPath.getParentPath` should be called something else. It does not return the parent if you pass something else than 1. I'm also unsure if this method should be in ContentPath, seems too specific to be in API, might not be used for anything else. At least rename to getAncestorPath, getSubPath(levels) ?.I would move it to `ContentResource` where it's used.

There are no calls to the new `ContentPaths` method, taking a Stream. And there are anyway enough variants already. This class should be left as it was.

Is there any change in what `NodeStorageServiceImpl.doReturnNodes` does? It's less readable, more verbose and probably less efficient since it now creates an intermediate list.
Same in `ContentNodeHelper`, it's less readable after the change and does the same. I think sometimes a for is better than a stream with a lambda expression.

alansemenov:
Merged to "next" branch.

PR #5213, opened 2017-06-27
Make sure all data in system-repo:master is deleted before loading da…

runarmyklebust:
…ta (#5209)

Added deletion of all system repo nodes before loading system-repo again. Optimized away one invocation on the batched executor.


GlennRicaud:
Nice 👍

Issue #5023, opened 2017-05-29
Mobile resolution - ImageEditor dialog doesn't fit the viewport


sgauruseu:
1. set 480x800 and open an image
2. click on the button-crop and try to crop the image
BUG- 'Apply' and 'Close' are not available
![imageeditor](https://cloud.githubusercontent.com/assets/3728712/26554654/0be98fb8-449a-11e7-8e11-a5b5a8ffd85f.png)

alansemenov:
The editor toolbar and zoom control also have wrong width.

PR #5237, opened 2017-06-29
Mobile resolution - ImageEditor dialog doesn't fit the viewport #5023

edloidas:
Fixed the width of the image editor, since it overflows the screen in mobile mode.
Fixed the resize of the image editor in edit mode.


alansemenov:
It's almost fine. The only issue is that edit mode is missing right margin compared to non-edit mode (and compared to left margin in edit mode).
See below:

Non-edit mode (equal margins on the sides of the image)
image

Edit-mode (toolbar and zoom control don't have correct right-margin):
image

Issue #5019, opened 2017-05-29
Mobile resolution - Don't unhighlight grid item and always open Preview Panel on click 


sgauruseu:
1. open browse panel and click on any row, preview panel and 'Show Details Panel' is appeared
2. click on the Back button,  navigate to the grid again
ISSUE:
need to click twice on the row to open the Preview Panel again
![preview_panel_mob](https://cloud.githubusercontent.com/assets/3728712/26541890/54af52da-4460-11e7-866d-e7164661e21f.png)

alansemenov:
In mobile res we should not unhighlight an item on click, instead we should always show the preview panel on click. Another alternative is to get rid of highlighting at all in the mobile mode (its purpose is to function as a cursor which doesn't make sense on mobile), but I believe it's more work.

PR #5295, opened 2017-07-06
Mobile resolution - Don't unhighlight grid item and always open Previ…

edloidas:
…ew Panel on click #5019

Disabled the highlighting in mobile mode.
Changed the rendering item of the preview panel in the mobile mode to clicked item.

_The solution, when we keep the highlighting on click and just show the preview panel, has issues with highlight remove and with the root node operations (when nothing is selected)._


alansemenov:
Nice solution!

Issue #5185, opened 2017-06-23
Enable saving a custom site page as a template


alansemenov:
When a site is rendered by a custom controller, we should let user save the page as a template for future use.

![image](https://user-images.githubusercontent.com/11612490/27474469-a3353d60-5802-11e7-9e41-cc54cabeb26b.png)

Conditions for showing the button:
* Custom page with selected controller
* Only if the wizard is opened for a site 
* User has rights to create new content

What should happen when user clicks the Create button:
* Create a new template containing current structure of the page under ``/<site>/_templates``
* Apply the same controller to the template 
* Display name of the new template should be the same as the site
* Path should be ``"template-<sitename>"``, for example "template-features".
* If path already exists, add "-2" etc., for example "template-features-2"
* Open a new wizard tab with the new template, so that user could change name, path etc. if needed

PR #5307, opened 2017-07-07
Enable saving a custom site page as a template #5185

pmi:


modules/admin/admin-ui/src/main/resources/web/admin/common/js/liveedit/PageView.ts
```diff
@@ -197,7 +203,23 @@ module api.liveedit {
                 this.resetAction.setEnabled(false);
             }
 
-            this.addContextMenuActions([inspectAction, this.resetAction]);
+            this.saveAsTemplateAction = new api.ui.Action('Save as Template').onExecuted(action => {
```
alansemenov: Action label here is not localised

modules/admin/admin-ui/src/main/resources/web/admin/common/js/liveedit/PageView.ts
```diff
@@ -197,7 +203,23 @@ module api.liveedit {
...
+                    .setDisplayName(content.getDisplayName())
+                    .setSite(content.getPath())
+                    .setSupports(content.getType())
+                    .setName(content.getName())
+                    .sendAndParse().then(createdTemplate => {
+
+                    new EditContentEvent([ContentSummaryAndCompareStatus.fromContentSummary(createdTemplate)]).fire();
+                });
+            });
+
```
alansemenov: I suggest extracting SaveAsTemplateAction into a separate class which both PageView and PageInspectionPanel classes will use. As of now this code is duplicated in these two classes.

modules/app/app-contentstudio/src/main/js/app/wizard/page/contextwindow/inspect/page/PageInspectionPanel.ts
```diff
@@ -66,9 +78,30 @@ export class PageInspectionPanel extends BaseInspectionPanel {
 
         this.pageControllerSelector = new PageControllerSelector(this.liveEditModel);
         this.pageControllerForm = new PageControllerForm(this.pageControllerSelector);
+        this.pageControllerForm.onShown(event => this.updateSaveAsTemplateVisibility());
         this.pageControllerForm.hide();
         this.appendChild(this.pageControllerForm);
 
+        this.currentUserHasCreateRights = null;
+        this.saveAsTemplateButton = new Button(i18n('action.saveAsTemplate'));
+        this.saveAsTemplateButton.addClass('blue large save-as-template');
```
alansemenov: See comment above - this should be an ActionButton linked to SaveAsTemplate action.

modules/app/app-contentstudio/src/main/js/app/wizard/page/contextwindow/inspect/page/PageInspectionPanel.ts
```diff
@@ -111,6 +144,27 @@ export class PageInspectionPanel extends BaseInspectionPanel {
...
+                });
+            } else {
+                this.saveAsTemplateButton.setVisible(this.currentUserHasCreateRights.valueOf());
+            }
+        } else {
+            this.saveAsTemplateButton.setVisible(false);
+        }
+    }
+
     private doSelectTemplate(selectedOption: PageTemplateOption) {
```
alansemenov: Code in this method is almost identical to the one changing visibility of the action in PageComponentView's context menu (method with the same name). If we switch this to the SaveAsTemplate action then this code should also be centralized inside the SaveAsTemplateAction class and the button will hide/show based on the action.

alansemenov:
Also, one more thing - after we created a new template, we should refresh the list of templates in PageTemplateSelector in the Inspection panel to make sure the new template comes up there.

alansemenov:
Great stuff!

Issue #5301, opened 2017-07-07
IssueDetailsDialog - Publish button should be disabled, when user has no rights to publish the content


sgauruseu:
1. add a user with roles: Content Manager Administrator + Content Manager App
2. add a content and create issue to publish the content
3. open the content and update permissions:
    deny the Publish operation for the user
4. login with the user and open IssueDetails dialog
5. press the Publish button
BUG: the user has pressed the Publish button, the content had not published but Issue is closed
![publish1](https://user-images.githubusercontent.com/3728712/27954808-66878f76-631a-11e7-88af-81f7de4ed5df.png)

alansemenov:
Publish button in the Issue dialog must follow exactly the same behaviour as in the Publishing wizard. See also #5268.

PR #5311, opened 2017-07-10
IssueDetailsDialog - Publish button should be disabled, when user has…

vbradnitski:
… no rights to publish the content #5301

### DIFF
--- a/modules/app/app-contentstudio/src/main/js/app/issue/view/IssueDetailsDialog.ts
+++ b/modules/app/app-contentstudio/src/main/js/app/issue/view/IssueDetailsDialog.ts
@@ -56,8 +56,8 @@ export class IssueDetailsDialog extends SchedulableDialog {
 
     private constructor() {
         super(<ProgressBarConfig> {
-            dialogName: i18n('dialog.issue'),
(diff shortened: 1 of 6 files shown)


modules/app/app-contentstudio/src/main/js/app/issue/view/IssueDetailsDialog.ts
```diff
@@ -396,7 +396,8 @@ export class IssueDetailsDialog extends SchedulableDialog {
                 const countToPublish = this.countTotal();
                 this.updateButtonCount(i18n('action.publish'), countToPublish);
 
-                this.toggleAction(countToPublish > 0 && !depResult.isContainsInvalid());
+                this.toggleAction(countToPublish > 0);
+                this.getButtonRow().getActionMenu().setEnabled(!depResult.isContainsInvalid() && depResult.isAnyPublishable());
```
alansemenov: Shouldn't it be ``isAllPublishable`` in this condition? The button should be disabled unless **all** of the items are publishable. 

It also doesn't seem to work. Here you can see that the "Publish" action are disabled for selected item:

image

But here the button is enabled for the same item (and I think it's because of ``isAnyPublishable``):
image

PR #5344, opened 2017-07-14
Localize XP Tour #5323

ashklianko:
-Added localization for all phrases in xp tour
-Updated links to Content Studio and Applications


ashklianko:
Updated

Issue #5304, opened 2017-07-07
Create Issue dialog - Filter out assigness without publish permissions


sgauruseu:
Only users of these roles should be shown in the Assignees selector:
     -Administrator
     -Content Manager Administrator
     -Content Manager Expert
     -Content Manager App
 
Note that if existing issue is assigned to a user that doesn't match this criteria, it should still be visible in the Issue details dialog (in readonly mode).

alansemenov:
See comment in https://github.com/enonic/xp/pull/5363

PR #5363, opened 2017-07-20
Create Issue dialog - Filter out assigness without publish permission…

ashklianko:
…s #5304

-Hiding users depending on their possibilty to publish and the fact if they are among approvers or not brings to complicated logic, thus showing users that don't have roles allowing publish as readonly, so if issue has approvers without publish rights they can be deselected but not selected back; Feel free to remove this if find better solution.


alansemenov:
@aro can you look at the java code please?

aro:
I don't think including all the group members in the REST response is a good idea. In a real installation a group can have thousands of users, it will kill the client. Do we need to get group members for this issue?

It is also mixing members and memberships. When the `resolveMemberships` parameter is set, it returns the users with their memberships and the groups with their members. So you can find out if a user has a certain role but not if a group has the role.

It also looks like it's not including transitive dependencies. If a user is a member of group A, and group A is member of role "Content Manager Administrator", will that work?

I would remove the fetching of group and role members if it's not needed for this case.

Issue #5334, opened 2017-07-12
Implement min and max config for Double input type


alansemenov:
Similar to #5333

PR #5364, opened 2017-07-21
Implement min and max config for Double input type #5334

vbradnitski:


alansemenov:
* Validation of values for min/max should happen in ``valueBreaksRequiredContract method``. It should not be needed to clone an instance of InputValidationRecording. See https://github.com/enonic/xp/commit/32f4300f62e9efdf0b36304b3d8761fa3a40f091 for proposal.
* For some reason a field value that doesn't match min/max conditions doesn't make the content itself invalid (you can still publish it, for example)
* Error messages should be localised

Issue #5328, opened 2017-07-12
Implement adjustment of orientation in the Image Editor


alansemenov:
Add a new icon to rotate an image 90 degrees clockwise (``rotate_right2`` on Icomoon). Tooltip: 'Rotate clockwise'.

![image](https://user-images.githubusercontent.com/11612490/28114791-9c67258e-6702-11e7-8d1c-24b87641cb35.png)

Orientation should be stored in metadata (data.media). Test that it's taken into account when an image is rendered, if not - implement this as well.
![image](https://user-images.githubusercontent.com/11612490/28115036-9a761d60-6703-11e7-9597-d783a73f82d6.png)

Read about available options for orientation meta here: http://sylvana.net/jpegcrop/exif_orientation.html

Note that it's not a toggle button as the other two, clicking on it will rotate the image right away.

"Reset filters" should reset orientation to original value from media.cameraInfo

PR #5380, opened 2017-07-27
Implement adjustment of orientation in the Image Editor #5328

pmi:


alansemenov:
It doesn't seem right at all...

- Image is not rotated 90 clockwise as the task description says. The first time I clicked the icon, the image got mirrored. The second time it rotated 180 degrees. Seems totally chaotic. Tried with several images under Features and it was the same problem with all of them.

- Orientation is not taken into account by the rendering engine. It should be the same as with cropping/focal settings, whenever these are set the changes must be applied when the image is rendered.
image

- The least of all the problems - there's no tooltip on the icon

Issue #5285, opened 2017-07-05
Add selection of Role in the User Group wizard


alansemenov:
It's possible to assign user to role(s) in the User wizard and we should enable the same for user group. Add UserMembershipsWizardStepForm section, like we have in the User wizard, but it should only have Roles, not Groups:

![image](https://user-images.githubusercontent.com/11612490/27867058-de7964c4-6198-11e7-979b-ecb273c28f53.png)

PR #5385, opened 2017-07-29
Add selection of Role in the User Group wizard #5285

edloidas:
Added support of roles selection in Group wizard.

### DIFF
--- a/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/security/SecurityResource.java
+++ b/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/security/SecurityResource.java
@@ -518,19 +518,6 @@ public final class SecurityResource
         return PrincipalKeys.from( members.stream().filter( PrincipalKey::isUser ).collect( toList() ) );
     }
 
-/*    private PrincipalKeys getGroupMembers( final PrincipalKey principal )
-    {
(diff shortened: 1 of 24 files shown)


alansemenov:
It works fine but there are some things that need to be fixed.

1. Pls remove commented code in ``SecurityResource.java```

2. Roles in the group's statistics panel are listed too tight (compare to roles in the user's panel):
image

3. The three ``append[...]Metadata`` methods in ``UserItemStatisticsPanel`` should be refactored (I know it's not a part of this task, but still).

Each of these three methods first creates an empty div with a header and lists' subheaders in DOM and then removes them from DOM and recreates again (this time with data) once the data is fetched. This feels like too many unnecessary operations with DOM. 

While it's fine to instantly create empty groups with headers so that user could see some content right away, I don't think we need to show subheaders of empty lists, this feels confusing.

For example, what's "Email" here? Is it a user name? No, it's a header of an empty list. Should not be there.

image

What we need is:
a) Create all groups with their headers (``h2``)
b) Make a request to fetch data
c) If data for the list is not empty, create both the ``list-header`` (if needed) and the list and append them to the group from a)

Note that this will require new methods and some refactoring in the ``ItemDataGroup`` class as well (for example, ``addDataArray`` and ``addDataElements`` are 90% identical)

alansemenov:
Very nice refactoring, @edloidas

Issue #5388, opened 2017-08-01
Status message when deleting multiple users should say how many are deleted


jsi:
This is a follow-up to this improvement: https://github.com/enonic/xp/issues/5284

When deleting many users, a blue status line appears at the bottom for one of the deleted users.  Instead, it should say "X users deleted".

PR #5392, opened 2017-08-01
Status message when deleting multiple users shouls say how many are d…

edloidas:
…eleted. #5388

Changed the notification message, when deleting multiple users from the grid.


alansemenov:
@edloidas Looks fine but we need this in app-users too

Issue #4571, opened 2017-03-07
Implement TaskDescriptorService


srs:
Implement TaskDescriptorService in `com.enonic.xp.task`. This should implement the following methods:

```
TaskDescriptors getTasks();
TaskDescriptors getTasks(ApplicationKey app);
```

**NOTE**: See other *DescriptorService's for example.

It should load tasks from location in app `/tasks/<name>/<name>.xml` or .js extension. The xml contains both description and form.

PR #5395, opened 2017-08-02
Implement TaskDescriptorService (#4571)

aro:


GlennRicaud:
Only looked at the code but it looks good.
I will test the entire epic issue with app-features once it is in review.

Issue #5336, opened 2017-07-12
Implement max-length config for TextArea input type


alansemenov:
Same as #5335

Issue #5335, opened 2017-07-12
Implement max-length config for TextLine input type


alansemenov:
```
    <input name="requiredTextLine" type="TextLine">
      <label>Required Text Line</label>
      <indexed>true</indexed>
      <custom-text>Custom text</custom-text>
      <occurrences minimum="1" maximum="0"/>
      <help-text>Some help text</help-text>
      <config>
        <max-length>10</max-length>
      </config>
    </input>
```
Max number of characters will be shown under the textbox, right-aligned, and will change (decrease/increase) as user is typing inside the field (ala-Twitter). 

![image](https://user-images.githubusercontent.com/11612490/28120915-c233062c-6719-11e7-8b7b-0094a640a3db.png)

![image](https://user-images.githubusercontent.com/11612490/28121335-0436c1e8-671b-11e7-80ff-234f0035dccc.png)

It should still be possible to continue typing in the field even once the limit is reached (but the counter should not go below 0). If user is trying to save a longer text than allowed, the validation should kick in:

![image](https://user-images.githubusercontent.com/11612490/28120788-5c8bbc9c-6719-11e7-9097-d4897299bab7.png)

aro:
Shouldn't we also have some kind of backend validation? 

I think that if a content is created via API and the value set to something longer than the max, validation should fail and it should not be possible to publish it using JS API neither.

vbradnitski:
I have it stashed already, just wanted to discuss it with @alansemenov

vbradnitski:
implemented in #5399

PR #5398, opened 2017-08-02
Implement max-length config for TextArea input type #5336

vbradnitski:


vbradnitski:
Contains changes from #5335 
@alansemenov TextLine contains 'data-regex-status', which displays 'Valid/Invalid' text depending on regexp check. But with new validation it looks a bit not very obvious in my opinion. Should it be changed?
image

modules/admin/admin-ui/src/main/resources/web/admin/common/js/form/inputtype/text/TextLine.ts
```diff
@@ -6,21 +6,28 @@ module api.form.inputtype.text {
...
 
         private readConfig(inputConfig: { [element: string]: { [name: string]: string }[]; }): void {
-            let regexpConfig = inputConfig && inputConfig['regexp'] && inputConfig['regexp'][0];
-            let regexp = regexpConfig && regexpConfig['value'];
-            this.regexpStr = regexp || null;
+            const regexpConfig = inputConfig['regexp'] ? inputConfig['regexp'][0] : {};
+            const regexp = regexpConfig ? regexpConfig['value'] : '';
+            this.regexp = new RegExp(regexp);
+
+            const maxLengthConfig = inputConfig['max-length'] ? inputConfig['max-length'][0] : {};
```
alansemenov: ``max-length`` property from config xml gets converted to ``maxLength`` on the server, so this should be changed to ``inputConfig['maxLength']``

modules/admin/admin-ui/src/main/resources/web/admin/common/js/form/inputtype/text/TextInputType.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        private maxLength: number;
+
+        constructor(config: api.form.inputtype.InputTypeViewContext) {
+            super(config);
+            this.readConfig(config.inputConfig);
+        }
+
+        protected readConfig(inputConfig: { [element: string]: { [name: string]: string }[]; }): void {
+            const maxLengthConfig = inputConfig['max-length'] ? inputConfig['max-length'][0] : {};
```
alansemenov: ``max-length`` property from config xml gets converted to ``maxLength`` on the server, so this should be changed to ``inputConfig['maxLength']``

modules/admin/admin-ui/src/main/resources/web/admin/common/js/form/inputtype/text/TextInputType.ts
```diff
@@ -0,0 +1,78 @@
...
+                    inputEl.getParentElement().appendChild(lengthCounter);
+                }
+            });
+
+            return inputEl;
+        }
+
+        private updateLengthCounterValue(lengthCounter: DivEl, newValue: string) {
+            lengthCounter.setHtml(`${this.maxLength - newValue.length}`);
+        }
```
alansemenov: The counter should not go below 0 according to the task description, but now it does.

modules/admin/admin-ui/src/main/resources/web/admin/common/js/form/inputtype/text/TextInputType.ts
```diff
@@ -0,0 +1,78 @@
...
+        protected readConfig(inputConfig: { [element: string]: { [name: string]: string }[]; }): void {
+            const maxLengthConfig = inputConfig['maxLength'] ? inputConfig['maxLength'][0] : {};
+            const maxLength = NumberHelper.toNumber(maxLengthConfig['value']);
+            this.maxLength = maxLength > 0 ? maxLength : null;
+        }
+
+        protected initOccurenceListeners(inputEl: FormInputEl) {
+
+            inputEl.onValueChanged((event: api.ValueChangedEvent) => {
+                if (NumberHelper.isNumber(this.maxLength)) {
```
alansemenov: Value in ``this.maxLength`` will be set only once and never change, so there's no point in checking for ``isNumber`` every time inside these two handlers. Instead, we should create both event handlers only **if** ``this.maxLength`` is a number:

```Typescript
if (NumberHelper.isNumber(this.maxLength)) {
                inputEl.onValueChanged((event: api.ValueChangedEvent) => {
                    const lengthCounter = Element.fromHtmlElement(
                        (<HTMLElement>inputEl.getParentElement().getHTMLElement().querySelector('.length-counter')));
                    if (lengthCounter) {
                        this.updateLengthCounterValue(lengthCounter, inputEl.getValue());
                    }
                });

                inputEl.onRendered(() => {
                    const lengthCounter = new DivEl('length-counter');
                    this.updateLengthCounterValue(lengthCounter, inputEl.getValue());

                    inputEl.getParentElement().appendChild(lengthCounter);
                });
            }
```

PR #5404, opened 2017-08-03
Implement submitTask (with name) in TaskService

aro:


GlennRicaud:
Only looked at the code but it looks good.
Will create a task to replace TaskKey by DescriptorKey
Will test the entire epic issue with app-features once is in review

PR #5410, opened 2017-08-04
Implement executing of named task in task-lib

aro:


GlennRicaud:
@aro: Why a TaskKey and not reuse DescriptorKey?

Issue #5119, opened 2017-06-19
Implement Create user entity dialog


alansemenov:
When nothing is selected in the grid, "New" button in the toolbar should open a new modal dialog with the list of available entity types. When user clicks an entity, we open the wizard for selected entity type.
Use the same on-hover styling as in the "New Content" dialog.
Note "..." in "New..." - it means the action will open a new dialog.

![screen shot 2017-06-19 at 11 07 34](https://user-images.githubusercontent.com/11612490/27277754-9b3f5cce-54df-11e7-8539-a31fe9cf8fbc.png)

![image](https://user-images.githubusercontent.com/11612490/28110127-ada93424-66f1-11e7-82b2-fea5a3ca8816.png)

Note that User node has second level with the list of available user stores - this is to indicate which user store the user will be added to. User Group will follow the same behaviour. In this case a click on User or User Group will simply expand/collapse the node, user must click on the user store to go to the wizard.

Selection of user store is required only if there's more than one user store in the system, otherwise it's enough to select User or User Group and it will be created in the System User Store.

Role and User Store nodes will always have one level.

sigdestad:
+1

PR #5416, opened 2017-08-07
Implement Create user entity dialog #5119

edloidas:
Created a new grid, grid item, and viewer to handle new principal dialog.


alansemenov:
Comment to the last item - I see that the icon is already in the DOM, but hidden by CSS. So please just unhide it.

image

alansemenov:
@edloidas Nevermind the first comment - it should be 45px, but I will fix that in master. Don't do anything about it.

Issue #5329, opened 2017-07-12
Implement image flip in the Image Editor


alansemenov:
Similar to #5328 

Add a new icon to mirror an image (``mirror`` icon on Icomoon). Tooltip: 'Mirror image'.

![image](https://user-images.githubusercontent.com/11612490/28115650-1577ea46-6706-11e7-9f2f-b4fc284d1d81.png)

If an image has original orientation 'Top, left side' then mirroring it will change orientation to 'Top, right side'.

Store orientation similar to #5328.

"Reset filters" should reset orientation to original value from media.cameraInfo

PR #5417, opened 2017-08-07
Implement adjustment of orientation in the Image Editor #5328

pmi:
Implement image flip in the Image Editor #5329
Implement rendering of orientation changed in the Image Editor throughout the admin #5382


jsi:
Is it possible to do something about the MediaInfoServiceImpl class, so the test coverage is not reduced?  Ideally, we should have 100% coverage on such a class, if that is possible.

alansemenov:
@jsi The last commit from Pavel was exactly the tests for MediaInfoServiceImpl

jsi:
Ok.  I guess I don't understand the flow on this page.  It seems the MediaInfoServiceImple class have had 62.96% coverage in every commit, but I see now, that there is no new CodeCov report after the last commit.

pmi:
Ok, here came the codecov report for last commit, but I'm not quite sure I understand how it works, because it didn't help much :(

pmi:
Or did it use previous results in the box where it says that some checks were not successful ?

jsi:
Well, one problem with CodeCov, is that the Coverage in the master column is the CURRENT coverage, so say you branched from master when the coverage was 78,00%, increased coverage on your branch to 78,02%, but at the same time, the coverage on master was increased to 78,05%.  The report in the pull request will then state that your commits have reduced the coverage by 0,03%, even though you have actually increased it by 0,02%, and the result of merging will be 78,07% on master.  (Did that make sense?)

So, anyway, that is why I try to look at individual files instead of the overall report that compares to master, because it is quite likely that master have changed since you branched, but it is quite unlikely that the files you are working on also have been changed in master, since you branched.

pmi:
Thanks for explanation @jsi ! That makes sense !

jsi:
Good!  😇

alansemenov:
It's still not working 100% correctly for me.

This is what I do:
1. Test with any image under features/media
2. Open it in the wizard
3. Add controller (expand the right panel first)
4. Add an image component to the page
5. Select the same image in the selector that was opened

This is how it initially looks:
image

Then I start playing with the new controls.
6. Click Rotate icon once

**Problem 1**: The image rotates correctly but rotation is very slow, seems like it's done on the server. I remember you struggled with client-side transformation, but what was really the problem there? It should be doable.

**Problem 2**: If the wizard panel is narrow, the "reset" link and upload button are wrapped to the second line and that looks bad. We should set max length for the link and overflow it if it's long, or at least make it so that it doesn't overlap the image.

image

7. Then I save the changes.

**Problem 3**: The image is correct in the wizard panel and the small thumbnail to the left of the display name, and also in Preview, but it's wrong in the Page Editor, even though the editor was refreshed after I saved.

image

I thought it was simply not refreshed but when I reloaded the page, it was still the same. This is very strange because in preview everything was correct and I thought that we are using preview mode in the editor.

8. Click the Mirror icon. The image is correctly (but slowly) changed. 
9. Save changes.
10. Open Preview

**Problem 4**: Preview is still displaying the image as it was after the first rotate, it doesn't take mirroring into account. Correct in the wizard but wrong in the preview. And the Page Editor is still displaying the original image, before all the changes.

pmi:
Okay, addressed all the points except the rotation speed, because it's the speed of the canvas rotating the image and then loading it into image editor. It's all done on the front end, but the image is 2,4 * 3,6 thousand pixels which takes time. 
I tried to use CSS transformations here, but it didn't work due to the fact that despite visually transformed all coordinates and dimensions are still calculated on the original image and image editor makes heavy use of mathematical calculations.
The rest of the points were caused by caching the images by browser that I fixed by adding timestamps.
There was also one issue with content wizard not updating the orientation field after save, but that was caused by not updating the form context's content when new content arrived after save.

pmi:
Long story short, should work now ;)

alansemenov:
Very impressive job, @pmi!

aro:
@pmi Instead of adding the cache parameter in ```ImageUrlParams``` (and other places), we should include the orientation in the image hash of the URL.
This is done in ```ImageUrlBuilder.resolveHash()```. Now we take the `binaryKey+focal_point+cropping`, so when one of those change the hash will be different, which will make the URL different and the browsers won't fetch the cached version. We need to add the Orientation value to the key.

What are the changes in ```ImageServiceImpl```? Isn't it just another way to express the rotation angle?

In MediaInfoServiceImpl.getImageOrientation() it would be better to just check on the xtradata for camera info. No need to iterate over all of them:
```final ExtraData next = media.getAllExtraData().getMetadata( MediaInfo.CAMERA_INFO_METADATA_NAME );```

aro:
Looks good. Good catch about ImageServiceImpl, I guess we didn't test all the possible orientations.

One last thing: ```Media.getOrientation()``` should return a ```ImageOrientation```. That also handles another issue, when the metadata is extracted from a new image, sometimes the orientation field gets a value like this: `"Top, left side (Horizontal / normal)"`, and the Integer.valueOf would throw and exception. ImageOrientation.from() handles that and returns the default orientation.

pmi:
Right, updated.

Issue #5415, opened 2017-08-04
Site configuration with mixin


GlennRicaud:
If a site descriptor includes a mixin in its configuration, listing applications when this application is not started will fail.

Look at
ProcessingCache.doProcess
(From what I quickly saw, we list applications (all installed), but the resolution of mixin is done only against started application).

GlennRicaud:
The question here is what should be accessible when.

ResourceServiceImpl loads resource from started application only
SiteDescriptorRegistry stores site descriptor for installed applications. 
ApplicationDescriptorServiceImpl stores application descriptor for installed applications.

GlennRicaud:
AuthDescriptorServiceImpl
PageDescriptorServiceImpl
PartDescriptorServiceImpl
...
all use ResourceService.processResource. that as mentionned above filter stopped applications

Rather than handling BundleEvents and having its own registry, SiteDescriptorService should use ResourceService.processResource and its caching. That would make the code more coherent and solve the problem

Check if there is implications in stopped applications being returned without their site descriptor. For the moment there already seem to be problem when a site is being edited and an application used is not started. The content is apparently modified.

![screen shot 2017-08-07 at 10 33 41](https://user-images.githubusercontent.com/12030113/29018706-70107ea4-7b5c-11e7-98a3-3462eb73e0a8.png)
![screen shot 2017-08-07 at 10 34 15](https://user-images.githubusercontent.com/12030113/29018711-726b0fa2-7b5c-11e7-8ec4-c2998548f3d1.png)

PR #5428, opened 2017-08-07
Site configuration with mixin (#5415)

GlennRicaud:


GlennRicaud:
Oopsi, I forgot about the tests. Thank you @aro .
I agree about the descriptor remark but all the other descriptor are doing it this way so for coherence I will leave it implemented like this.

Issue #5451, opened 2017-08-09
Optimize delete content command


runarmyklebust:
Now, we are doing an extra fetch of children that may not be necessary, and children are wrongly fetched recursive when we have to do traverse them to check statuses anyway

PR #5453, opened 2017-08-09
Optimize delete content command (#5451)

runarmyklebust:
Removed recursive fetch and moved fetching children until its needed


GlennRicaud:
I like it!

PR #5477, opened 2017-08-11
Read-only mode #5423

runarmyklebust:


GlennRicaud:
Looks good.

- For git branch names, the convention is to start with "issue-"
- There will be some adaptation in Content studio and Users to handle these new exceptions

Issue #5072, opened 2017-06-08
Implement the possibility to use bountycastle security


srs:
Implement the possibility to use bountycastle security.

PR #5485, opened 2017-08-16
Implement the possibility to use bountycastle security #5072

aro:
In the libs where it will be used we need to exclude the bountycastle artifact.

build.gradle
```
dependencies {
    compile "org.bouncycastle:bcprov-jdk15on:1.54"
    include( "nl.martijndwars:web-push:3.0.0" ) {
        exclude group: 'com.google.guava', module: 'guava'
        exclude group: 'org.bouncycastle', module: 'bcprov-jdk15on'
    }
}
```

In Java class where is used:
```
import java.security.Security;
import org.bouncycastle.jce.provider.BouncyCastleProvider;

static
{
    if ( Security.getProvider( BouncyCastleProvider.PROVIDER_NAME ) != null )
    {
        Security.removeProvider( BouncyCastleProvider.PROVIDER_NAME );
    }
    Security.addProvider( new BouncyCastleProvider() );
}
```

### DIFF
--- a/modules/distro/build.gradle
+++ b/modules/distro/build.gradle
@@ -25,6 +25,9 @@ addBundle( 'org.apache.felix:org.apache.felix.eventadmin:1.4.4', 5 )
 addBundle( 'org.apache.felix:org.apache.felix.configadmin:1.8.8', 5 )
 addBundle( 'org.apache.felix:org.apache.felix.scr:2.0.2', 5 )
 
+// Java Cryptography Extension
+addBundle( 'org.bouncycastle:bcprov-jdk15on:1.54', 6 )
+
 // Library dependencies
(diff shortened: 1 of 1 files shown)


GlennRicaud:
It works! Nice! We can finally used notification \o/

Reminder of the original problem and how to reproduce:
The problem was encountered while implementing notifications for Office League which was using the ECDH algorithm implemented in a external library BouncyCastle
The real problem was an exception "java.lang.SecurityException: Cannot verify jar:com.enonic.app.officeleague" thrown in JceSecurity.verifyProviderJar
  KeyAgreement.chooseProvider
    JceSecurity.canUseProvider
      JceSecurity.verifyProviderJar  
This same code was working when called from a static void main method instead of our OSGi environment.

Issue #5129, opened 2017-06-19
Configure CKEditor to work in inline mode


alansemenov:
The biggest challenge here is probably getting the sticky toolbar to function properly.
http://docs.ckeditor.com/#!/api/CKEDITOR-method-inline

Issue #5142, opened 2017-06-20
Integrate "Find/Replace" modal dialog with CKEditor


alansemenov:

alansemenov:
Issue moved to [enonic/lib-admin-ui #397](https://github.com/enonic/lib-admin-ui/issues/397) via [**ZenHub**](https://www.zenhub.com/)

Issue #5132, opened 2017-06-19
Add support for fullscreen mode


alansemenov:
http://sdk.ckeditor.com/samples/fullpage.html

In the "wizard" mode the icon should be placed the footer's right corner, as now. In the inline mode we don't need it (not in the toolbar either).

Issue #5130, opened 2017-06-19
Customize CKEditor's toolbar 


alansemenov:
Customize the default configuration of the toolbar so that we have the same set of tools as we do now. Implement support for custom configuration via HTML Area schema. 

http://docs.ckeditor.com/#!/guide/dev_toolbar

alansemenov:
Also, don't forget that toolbar should only become visible when the HTML Area gets focused.

ashklianko:
Sure

alansemenov:
@ashklianko 
There were a few small things to fix, see if you can do this in one of the next PRs:
* The footer of the area is now always visible (also in TinyMCE), we should hide it when the area is not active
* Remove the "Code" (it doesn't work because ``<code>`` tag must be placed inside a container element, btw) and "Create Div container" buttons
* Move "Block quote" inside the group of buttons to the left
* Remove 20px margin inside the HTML Area's body element

PR #5495, opened 2017-08-17
Configure CKEditor to work in inline mode #5129

ashklianko:
Missing features for now:
-source dialog; will be added in other issue
for both inline and classic modes:
-blur/focus handlers commented now; cke wraps natural events with their own events, need to investiage more on this because some listeners expect natural FocusEvent that not present in cke event
-keyboard events handling

### DIFF
--- a/modules/admin/admin-ui/src/main/resources/web/admin/common/js/form/inputtype/text/HtmlArea.ts
+++ b/modules/admin/admin-ui/src/main/resources/web/admin/common/js/form/inputtype/text/HtmlArea.ts
@@ -225,6 +225,10 @@ module api.form.inputtype.text {
 
                 this.removeTooltipFromEditorArea(textAreaWrapper);
 
+                this.moveFullScreenButtonToBottomBar(textAreaWrapper);
+
+                this.moveFullSourceButtonToBottomBar(textAreaWrapper);
+
                 const removeButtonEL = wemjq(textAreaWrapper.getParentElement().getParentElement().getHTMLElement()).find(
                     '.remove-button')[0];
                 removeButtonEL.addEventListener('mouseover', () => {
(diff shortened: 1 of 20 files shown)


alansemenov:
* Multiple console errors whenever HTML area is clicked and gets focus
* Bottom margin of the toolbar should be the same as on the top
* "X" icon closing the edit mode should have 'pointer' cursor, not 'move'
* Remove margin inside the area (what I mentioned in #5130)

ashklianko:
On it

ashklianko:
Done

ashklianko:
Had to add #5142 and #5132 in this PR , having issues with local build

modules/admin/admin-ui/src/main/resources/web/admin/common/js/form/inputtype/text/HtmlArea.ts
```diff
@@ -237,6 +241,17 @@ module api.form.inputtype.text {
             });
         }
```
alansemenov: Replace the two methods below with one called ``appendButtonToBottomBar``where you pass the identifying class of the button to move.

modules/admin/admin-ui/src/main/resources/web/admin/common/js/util/htmlarea/dialog/CharMapDialog.ts
```diff
@@ -1,10 +1,11 @@
 module api.util.htmlarea.dialog {
 
     import i18n = api.util.i18n;
+    import editor = CKEDITOR.editor;
```
alansemenov: Change this to ``import HTMLAreaEditor = CKEditor.editor;`` (in all classes)-
It's a bit confusing when it says "``(editor: editor)``" in function signature.

ashklianko: ok!

modules/admin/admin-ui/src/main/resources/web/admin/common/js/form/inputtype/text/HtmlArea.ts
ashklianko: Done

Issue #5480, opened 2017-08-14
Move Content Dialog - incorrect behavior of filtering, when a search-text was typed in the options-filter


sgauruseu:
1. select the Features site,press the Move
2. type Features in the options-filter
Expected: Features option should be disabled
BUG: Features-site is enabled in the filtered-options
Click on the Feature-option and press the Move button: "Not allowed to move content to itself (/content/features)"  message appears
![move](https://user-images.githubusercontent.com/3728712/29263994-f5c37bfa-80e3-11e7-8bfc-7769261a051d.png)

Option is selected:
![step2](https://user-images.githubusercontent.com/3728712/29264008-05690c14-80e4-11e7-9202-c2b6bd3c723f.png)

BUT: if do not type a search-text   and  expand the options, Features is disabled in the options

PR #5496, opened 2017-08-18
Move Content Dialog - incorrect behavior of filtering, when a search-…

vbradnitski:
…text was typed in the options-filter #5480


vbradnitski:
done

Issue #5467, opened 2017-08-11
Remove site content type from new content dialogue if user does not have the "content manager admin" role


sigdestad:
Currently, this returns a nasty error on the new page anyway.

Issue #5466, opened 2017-08-11
Hide "base:unstructured" completely from the New Content dialog


sigdestad:
Currently, no-one need to create unstructured items from the UI. This should then be hidden from the createDialogue, but remain working as normal for everything else.

PR #5562, opened 2017-08-30
Hide "base:unstructured" completely from the New Content dialog #5466

ashklianko:
Remove site content type from new content dialogue if user does not have the "content manager admin" role #5467


ashklianko:
#5467 also included

Issue #5581, opened 2017-09-04
Incorrect info in Page Template widget when template is not found 


alansemenov:
(See also #5488)

When a site is using a page template which was deleted, the info inside the Page Template widget inside Details panel is incorrect - it shows info from the last selected content item and there are console errors when we try to fetch template info. We should handle this in the same way as we do in the Wizard. Since the site is still linked to the deleted template until saved, we should still show "Page Template" in the widget but on the second line instead of template name we should show "Not found".

![image](https://user-images.githubusercontent.com/11612490/30029118-215aab30-9188-11e7-935a-be1087efdb12.png)

PR #5583, opened 2017-09-05
Incorrect info in Page Template widget when template is not found #5581

vbradnitski:


alansemenov:
@vbr All good, but please localize 'Not found'

vbradnitski:
need to translate 'widget.pagetemplate.notfound'

vbradnitski:
and 'live.view.page.error.pagetemplatenotfound'

PR #5596, opened 2017-09-07
Add parameters to named tasks execution (Issue 5579)

aro:


jsi:
Do not merge this if the code coverage drop is real!!!  ;)

Issue #5655, opened 2017-09-29
Controller mappings returns 200 for WebSocket response


aro:
When a mapping controller handles a WebSocket request, the server responds with status 200, instead of 101, which breaks the WebSocket protocol.

```Javascript
function handleGet(req) {
    if (!req.webSocket) {
        return {
            status: 404
        };
    }

    return {
        webSocket: {
            data: {}, subProtocols: ["foo"]
        }
    };
}
```
Reported in Discuss: https://discuss.enonic.com/t/controller-mappings-returns-200-when-request-is-websocket-request/1235

PR #5656, opened 2017-09-29
Controller mappings returns 200 for WebSocket response (#5655)

aro:


GlennRicaud:
Looks good :+1: Fast fixing + Unit test. A second :+1: for you sir

Issue #5664, opened 2017-10-06
Rendering of images with legacy media property fails


aro:
Rendering of images created in older versions of XP fail while trying to obtain the image orientation.

The problem is that originally the images had a property with name `media` of type **String**. But later it was extended making it of type **PropertySet** with more complex data. So the code accessing this property must always handle both cases.

This bug was introduced in changes done for 6.12.

jsi:
Set "Not in Changelog", since it has been introduced since 6.11.

PR #5665, opened 2017-10-06
Rendering of images with legacy media property fails (#5664)

aro:


GlennRicaud:
Did not test, but the correction looks good

Issue #5651, opened 2017-09-27
Lib Auth - Create role


edloidas:
@edloidas commented on [Wed Sep 27 2017](https://github.com/enonic/xp-apps/issues/82)

Currently, only the users and groups use the authLib for common operations and checks. Make the roles and user stores do the same. Here are the steps:
* Create the required beans in `com.enonic.xp.lib.auth.*Handler`
* Make them use `SecurityServiceImpl`
* Create the bindings for those beans in `auth.js` in `lib-auth`
* Replace node-layer calls in `app-users` with authLib calls, where possible.

__Extra:__
* Make sure the client receives proper errors, like when the user tries to create a new group on the existing path, he should the message, that the group on that path already exists.

PR #5702, opened 2017-10-17
Replace node layer with `lib-auth` calls for `createRole` #5651

edloidas:
Refactored `createRole`.


GlennRicaud:
This function does not follow the same interface as createGroup and createUser.
Please replace the parameter 'principalKey' by 'userStore' and 'name'

Issue #5645, opened 2017-09-22
Wrong language resolution in admin


aro:
When I visit the admin with the following language settings in Chrome:

![screen shot 2017-09-22 at 08 50 07](https://user-images.githubusercontent.com/430739/30732749-50bb3b70-9f74-11e7-8d25-f13158bc5116.png)

It renders the page in English, while I set Norwegian preferred over English.

The request header is the following:
```Accept-Language:ca,no;q=0.8,en-US;q=0.6,en;q=0.4,es;q=0.2```

It just take the first language and if not found it goes to the default. It should do a proper handling of language preferences.

aro:
@alansemenov , @jsi : I looked into this. The problem is not the parsing of the HTTP header. The whole i18n system in XP is based on the idea that you get resources based on a selected locale (e.g. 'es'). I think it should be that you get it based on a list of prioritized locales (['es', 'en', 'no']).

We would need to change the way we read and handle the i18n bundles. Now they are loaded based on one specified language, and filling the blanks with the default language properties file. There is not any kind of registry saying we support this list of languages for this bundle. So I think this might be too many changes just before release.

For example, I think the [getLocale()](http://repo.enonic.com/public/com/enonic/xp/docs/6.11.0/docs-6.11.0-libdoc.zip!/module-admin.html#.getLocale) function should be ```getLocales()``` and return an array of locale codes, in order of priority according to the user browser settings. And then use that list to get the translated resources.

jsi:
Sorry for being ignorant here, but who made "the whole i18n system"?  Is it ours?  I thought Localization was a built in Java feature.

Your solution sounds good to me.  Whatever is needed to make this work.  We should give the user the language they "mostly" prefer out of the ones we can offer.  😊

aro:
Sorry, I might have expressed it wrong. It's ours. Localization is a Java feature, but our i18n Java and JavaScript API on top of that, expects one language as input for the translations.

jsi:
Well, everybody else probably knew that.  😜

PR #5705, opened 2017-10-17
Wrong language resolution in admin (#5645)

aro:


GlennRicaud:
Looks good :+1:. Was hoping to get you on backward compatibility but no :(
Will test when merging https://github.com/enonic/xp-apps/issues/146

Issue #5146, opened 2017-06-20
Site mapping - Post processing


GlennRicaud:
Check if page conributions can be executed for mappings

http://xp.readthedocs.io/en/latest/developer/ssjs/http-response.html?highlight=postProcess
postProcess: true ?
pageContributions?

GlennRicaud:
@jsi: Why the help needed?

GlennRicaud:
Does not work. To be fixed

PR #5712, opened 2017-10-20
Page contribution on mapping (#5146)

GlennRicaud:


GlennRicaud:
Would like your opinion on the fix. There may be some cases I do not think about

aro:
Closed pull request. Let's try implementing a new Renderer as discussed.

Issue #5716, opened 2017-10-23
Restore snapshot fails on first attempt on empty repo/index


runarmyklebust:
When restoring a repository-snapshot on an empty/new index, the path-cache is not cleared properly before the index is restored. 

I WILL work of doing the following:
1) Restore snapshot (get an exception)
2) Restart the xp-instance
3) Do the restore snapshot again.

PR #5719, opened 2017-10-23
Restore snapshot fails on first attempt on empty repo/index (#5716)

runarmyklebust:


GlennRicaud:
It now works :+1:
But this PR was for master and not 6.12. We should fix it in 6.12 branch too

Issue #5627, opened 2017-09-15
Create index-configuration for page properties


runarmyklebust:
Customers needs to get data from page indexed when storing content.

````
  "controller": "com.enonic.app.corporate.theme:default",
  "config": {
       "someConfig" : "Somevalue"
   },
  "regions": { ... }
````

* "controller" should be added to index-config with config "minimal"
* "config"-elements should be indexed with "BY-TYPE" as default
* The config-elements must be parsed according to page-schema, and html-areas must add the html-stripper processor.

Refactor and extends ContentIndexConfigFactory as needed, try to keep the code clean and testable - you will probably add some new classes or interfaces here, so it may be a good idea to move this stuff to a new package.

Make sure the test-coverage is 80%-ish.

runarmyklebust:
Some issues found;
1) Site-config, html-area property not html-stripped.

PR #5723, opened 2017-10-25
Create index-configuration for page properties #5627

vbradnitski:


vbradnitski:
works only with PageControllers

Issue #5772, opened 2017-11-09
InputType handled case-sensitive in class, but not in usage


runarmyklebust:
The content InputTypeName is case-sensitive in equals, but when used in ContentStudio, the name is normalized. This should be handled the same way both places, so the class' equal should do case insensitive equals.

PR #5773, opened 2017-11-09
 InputType handled case-sensitive in class, but not in usage (#5772)

runarmyklebust:


GlennRicaud:
equalsIgnoreCase would be better but I will validate it

PR #5779, opened 2017-11-13
Rewrite the reindex of IndexService to us a proper batched executor (…

runarmyklebust:
…#5764)


GlennRicaud:
Real batch execution. Nice :+1:

Issue #5620, opened 2017-09-12
ScriptExportsCache - Atomic load


GlennRicaud:
A javascript file should only be loaded if necessary and only once. For this we have the exportsCache in ScriptExecutorImpl, but the get and put is not atomic. 
Ex: This causes problem when two requests hit at a same moment the same service. The JS file is then loaded twice.

I first tried:
  1 - Transform the get&put into a computeIfAbsent on the ConcurrentMap. 
  2 - Remove the put which is not necessary (and blocking for computeIfAbsent).
But with the require in JS, this method is actually recursive and ConcurrentMap cannot be used like this since we "must not attempt to update any other mappings of this map."

So we will have to implement our own solution to lock by resourcekey
- Remember of recursivity
- Think about deadlock cases

As inspiration:
https://github.com/enonic/xp/pull/2181/files
https://dzone.com/refcardz/core-java-concurrency

GlennRicaud:
This has always been the behaviour in XP and I will not implement this 2 weeks after feature release. So I will put this issue back to Ready and tag it for 6.13.0.

I will create another issue with a specific fix for app-users

GlennRicaud:
To be tested thoroughly during test sprint

PR #5788, opened 2017-11-15
ScriptExportsCache - Atomic load (#5620)

aro:


GlennRicaud:
I think the tryLock should be outside of the try.

We can discuss the constants. Especially the number of stripes and what happens if we reach it. I like the tryLock since the application developer might have implemented cyclic dependency.

An alternative to this solution is to use com.enonic.xp.core.impl.image.ImmutableFilesHelper methods lock and unlock (Only replace Path by the ResourceKey). Not sure what are the pros and cons of using Conditions instead of Locks

PR #5797, opened 2017-11-17
MoveContentRequest should create a task on the server and respond wit…

edloidas:
…h taskId #5756

Updated `move` command to use `RunnableTask` class and the same approach as `delete` and `publish`.


edloidas:
@runarmyklebust Can you look at it again?

edloidas:
@runarmyklebust Fixed the `count` value.

And no, I don't have large repo. But I use debugging instead and can confirm that it actually counts the real progress. You see:
```java
// MoveContentCommand.doExecute()
// ...
nodeMoved( 1 );
for ( final NodeBranchEntry nodeBranchEntry : nodeBranchEntries )
{
    doMoveNode( nodeToMoveBuilder.build().path(), getNodeName( nodeBranchEntry ), nodeBranchEntry.getNodeId() );
}
```
When executed, the method will move only one node at once. And it will report that the node is moved `nodeMoved( 1 )`, after which it will execute the `move` command for each child node and also do report. So, every movement, be it actual content or it child, or the child of the child, will be reported.

Issue #5827, opened 2017-11-27
Changing snapshot-location does not work


runarmyklebust:
If a snapshot location is already created, its not possible to change to another location without deleting the index.

Related to issue: https://enonic.zendesk.com/agent/tickets/2222

GlennRicaud:
Bug fixed.

But I still have on the startup of the system the creation of a folder snapshots, even though the value snapshots.dir of the config file points to something else

runarmyklebust:
Well, its not easy to change since this is done when ES is starting and has an earlier reference to a snapshot-repo. ES will automatically try to make sure this path exists. If doing operations to the snapshots-repo, e.g list, it will update the reference according to the configured directory.

PR #5828, opened 2017-11-27
Changing snapshot-location does not work (#5827)

runarmyklebust:
Checking configured path when deciding if snapshot repo must be initialized.


GlennRicaud:
Bug fixed.

But I still have on the startup of the system the creation of a folder snapshots, even though the value snapshots.dir of the config file points to something else

Issue #5812, opened 2017-11-21
Add deployment info to the app info REST API response


alansemenov:
``deployment.url`` - when the app has "main.js" file in the root this will be the full app url, like "http://localhost:8080/app/com.xp.myapp"

```JSON
{
  "deployment": {
    "url": "http://localhost:8080/app/com.xp.myapp"
  }
}
```

alansemenov:
Coped Alex' comments from https://github.com/enonic/xp/pull/5833:

Some changes:
1. We will only show the app URL path when connecting on localhost. (Because due to vhost it won't work in other cases).
2. We need to check also if the main.js is exporting any request method: "get", "post", "head",... or "all" method. It could be that there is main.js but it's only used for initialisation, without any http function. (The "all" is new, see #5648)

For 1. Inject HttpServletRequest instead of UriInfo, ```@Context HttpServletRequest req```
And check if target host is localhost: 
```if (req.getServerName().equals( "localhost" ) )```

For checking 2., you should do something like this:
```java
PortalScriptService portalScriptService = ...;

final ScriptExports exports = portalScriptService.execute( resource.getKey() );

if ( exports.hasMethod( "get" ) || exports.hasMethod( "post" ) || 
     exports.hasMethod( "head" ) || exports.hasMethod( "all" ) ) {
   ...

```

Issue #5648, opened 2017-09-27
Support all function to handle all method requests in controllers


aro:
Currently for handling an HTTP request in a JS controller, we export a function with the name of the HTTP method:

Example:
```javascript
exports.get = function (req) {

}
```

It would be useful to be able to export a function to handle all and any method, as in PurpleJS:
```javascript
exports.all = function (req) {

}
```

We need this to use [lib-router](https://enonic-docs.s3.amazonaws.com/com.enonic.lib/lib-router/index.html) in a general way, not depending on the HTTP method.

sigdestad:
Seems reasonable. Is it a quick fix?

srs:
That should be a quick fix in XP. It's wise to check existence like this (in this order):

* hasMethod(req.getMethod().toLowerCase()) == true ---> execute the method specific.
* hasMethod('service') == true ---> execute the service method.

GlennRicaud:
I think that the equivalent in PurpleJS is `exports.service`
We should use the same naming for both

Or are exports.service and exports.all two different things?

sigdestad:
We should go for "all", and get rid of "service"

GlennRicaud:
I agree. And routers in NodeJs world also use "all". So it makes sense.
We will have to adapt PurpleJS then.

Merged.

PR #5833, opened 2017-11-30
Add deployment info to the app info REST API response #5812

vbradnitski:


aro:
Some changes:
1. We will only show the app URL path when connecting on localhost. (Because due to vhost it won't work in other cases).
2. We need to check also if the main.js is exporting any request method: "get", "post", "head",... or "all" method. It could be that there is main.js but it's only used for initialisation, without any http function. (The "all" is new, see #5648)

For 1. Inject HttpServletRequest instead of UriInfo, ```@Context HttpServletRequest req```
And check if target host is localhost: 
```if (req.getServerName().equals( "localhost" ) )```

For checking 2., you should do something like this:
```java
PortalScriptService portalScriptService = ...;

final ScriptExports exports = portalScriptService.execute( resource.getKey() );

if ( exports.hasMethod( "get" ) || exports.hasMethod( "post" ) || 
     exports.hasMethod( "head" ) || exports.hasMethod( "all" ) ) {
   ...

```

PR #5838, opened 2017-12-01
Implement REST API for fetching combined application info

vbradnitski:


aro:
Looks good. 
@ase @GlennRicaud  But does a user store have a "path"? Shouldn't it be "name" instead? Or "key"?

alansemenov:
@aro Yes, "key" is what we need.

Issue #5257, opened 2017-07-03
Event returned by eventlistener in lib-event is not formatted as a javascript object


RF0:
An event returned by an eventlistener in javascript library lib-event returns a  toString() from an object. the value is not json and makes it hard to handle in javascript. the separator is = not colon.

"nodes": "[{id=13b2285c-9d57-4936-8916-c6ad908079a3, path=/content/dk/ansatte/kantinen, branch=draft, repo=cms-repo}]"

@srs suggests we should probably implement MapSerializable on the object and then handle it in the javascript bridge.

GlennRicaud:
Was going to create this bug and just saw it at the top of the list. Perfect :+1:

GlennRicaud:
@RF0 This will break your regexp. Your app will have to be adapted for 6.13
@GlennRicaud: Same for me for the guillotine app.

@aro: We should add something in the upgrade steps to 6.13

PR #5846, opened 2017-12-06
Event returned by eventlistener in lib-event is not formatted as a ja…

aro:
…vascript object (#5257)

### DIFF
--- a/modules/script/script-api/src/main/java/com/enonic/xp/script/serializer/MapGeneratorBase.java
+++ b/modules/script/script-api/src/main/java/com/enonic/xp/script/serializer/MapGeneratorBase.java
@@ -1,8 +1,12 @@
 package com.enonic.xp.script.serializer;
 
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
 import java.util.Stack;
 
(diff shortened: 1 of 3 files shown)


GlennRicaud:
It works but not a big fan. When using the method rawValue of the interface MapGenerator I am expecting it to be set raw.
Instead in ScriptEventListenerImpl::toJson, we could use
  MapperHelper.serializeMap(entry.getKey(), entry.getValue(), true)

Or actually a better solution that would handle also collections/arrays: Add a new method MapperHelper.serialize taking an Object instead of a map.
  It would check what the object is and call the MapGenerator as it is supposed to be used.

aro:
Fixed so that maps and lists are not converted when using `rawValue` method.

I am not a big fan of having a Helper class in API for something that should just be handled in general. I don't see any reason why maps and lists should not be properly converted when passed to a MapSerializable.
If it's necessary in any case to pass it through without converting, we can use the rawValue method.

GlennRicaud:
Agreed on the value method :). Merged.

edwardgronroos:
Fingers crossed this will reach 6.13 🤞

Issue #5738, opened 2017-11-01
Add tracing info to executing tasks


aro:
Add tracing information for the tasks executed by XP, using Task api.

PR #5865, opened 2017-12-12
Add tracing info to executing tasks (#5738)

aro:


GlennRicaud:

Issue #5790, opened 2017-11-15
Remove Tracer from stacktrace when tracing is disabled


aro:
Avoid wrapping calls with Tracer.trace if tracing is disabled.

PR #5867, opened 2017-12-12
Remove Tracer from stacktrace when tracing is disabled (#5790)

aro:


GlennRicaud:

GlennRicaud:

PR #5868, opened 2017-12-12
Issue 5820

aro:
Merge after #5738


GlennRicaud:
I have not executed the code so not sure how it really works. But, in TaskWrapper::run, is setThreadName() supposed to be in the else instead?

 Can you resolve the conflicts also

GlennRicaud:
My bad about the ifelse. Not tested but code looks good. Merged

Issue #5207, opened 2017-06-27
JavaScript API for handling content attachments


aro:
This was a task from YT that seems to have been forgotten on the way to Github. ([XP-4693](https://youtrack.enonic.net/issue/XP-4693))

----
Currently it is only possible to create/update/remove attachments from content with the Java API

There is a request for this in Discuss:
https://discuss.enonic.com/t/how-to-store-attachment-when-creating-content-with-js-api/676

We should either add extra parameters to contentLib.create and contentLib.modify , or have separate functions: createAttachment , removeAttachment .

We already have getAttachments and getAttachmentsStream in the contentLib.

-----
Option 1:
```javascript
    var result = contentLib.create({
        name: 'mycontent',
        parentPath: '/a/b',
        displayName: 'My Content',
        branch: 'draft',
        contentType: 'test:myContentType',
        data: {},
        attachments: [
            {
                name: 'logo.png',
                label: 'small',
                mimeType: 'image/png',
                data: streamObject
            }
        ]
    });

    var result2 = contentLib.modify({
        key: '/a/b/mycontent',
        editor: function (c) {
            c.addAttachments = [
                {
                    name: 'logo.png',
                    label: 'small',
                    mimeType: 'image/png',
                    data: streamObject
                }
            ];
            c.removeAttachments = ['logo-old.png'];
            return c;
        }
    });
```

Option 2:
```javascript
    var result = contentLib.createAttachment({
        key: '/a/b/mycontent',
        name: 'logo.png',
        label: 'small',
        mimeType: 'image/png',
        data: streamObject
    });

    var result2 = contentLib.removeAttachment({
        key: '/a/b/mycontent',
        name: 'logo.png'
    });
```

alansemenov:
@aro we need to post an update on discuss about this one

PR #5870, opened 2017-12-13
JavaScript API for handling content attachments (#5207)

aro:


GlennRicaud:

Issue #5847, opened 2017-12-06
ContentResource - move reporters and message builder to different packages


edloidas:
* Split the bulky `ContentResource` and listener stuff into the different classes/methods, like `MoveTaskQueryFactory`, `MoveTask`, etc.
* Easily test the refactored code.

PR #5877, opened 2017-12-13
ContentResource - move reporters and message builder to different packages #5847

edloidas:
* Moved `duplicate`, `move`, `delete`, `publish`, `unpublish` tasks code to the separate classes with `CommonRunnableTask` as a parent class.
* Moved children query builders into the `ContentQueryWithChildren` class.
* Added tests for newly added classes.

_Almost **500** lines of code were moved from `ContentResource`, which is a **25%**, from the original size._


jsi:
Wow!  Code coverage of 78,24%  -  All time high!!!  :)

PR #5888, opened 2017-12-14
Edit Permissions Dialog - initial Acl-entry should be restored when '…

vbradnitski:
…Inherit permissions' checkbox unchecked #222


alansemenov:
I don't think we should keep commented code. If we need it, we can always look it up in the history of commits.

vbradnitski:
removed

Issue #5906, opened 2017-12-19
Boolean filters does not support arrays of filters


runarmyklebust:

GlennRicaud:
Main epic is done but the sub issues state is in progress and review

PR #5907, opened 2017-12-19
Boolean filters does not support arrays of filters (#5906)

runarmyklebust:


GlennRicaud:
Issue solved. Good for thinking about both node and content :+1:

Found a problem related but will create another issue about it

Issue #5845, opened 2017-12-05
Portal Lib - URL Generation - Context path parameter


GlennRicaud:
In portal lib URL functions, add a flag to specify if we want the URL to be generated at the rootest (like assetsUrl or idProviderUrl today) or under the current content path.
The default value of this flag would be the current behaviour (false for assets, true for service, ...)

Add this flag to the following functions:
- assetUrl
- imageUrl
- serviceUrl
- idProviderUrl
- loginUrl
- logoutUrl

GlennRicaud:
Note to self: Adapt documentation for view functions

sigdestad:
This was a bit unclear to me? What is "the rootest"? Do you have some examples of how this would work?

GlennRicaud:
Not sure what word to use to describe the way asset URLs are generated.
Ideally after the base uri ("/portal/draft" for example), but could contain some content path because of the vhosting

Assets on superhero have a URL like this:
  /portal/draft/_/asset/com.enonic.app.superhero:1514451089/css/style.css"
Images have a URL like this 
  /portal/draft/superhero/_/image/81b1e3cd-575f-4565-a618-3c85d56224f6:666fcc4247e015ab6878f91fd7502f3b882ace10/block-1024-355/superhero_featured_4.jpg

After this improvement, by setting the flag contentPath to "false" when calling imageUrl, we could have the following URL:
  /portal/draft/_/image/81b1e3cd-575f-4565-a618-3c85d56224f6:666fcc4247e015ab6878f91fd7502f3b882ace10/block-1024-355/superhero_featured_4.jpg

sigdestad:
But why would we need this?

GlennRicaud:
I proposed this feature for idProviderUrl. So that they could be generated for the current content path.
You validated and asked me to make it generic for all the URL functions. Especially for imageUrl (which could be interesting if the caching headers are also set).

For assetUrl, I do not see any interest in doing this.
For serviceUrl, it could be useful in some very specific cases if you need a fixed service URL.

sigdestad:
I start remembering.. So what do you propose calling the parameter?
We already have types right? Absolute and relative right?

GlennRicaud:
Yes

sigdestad:
What do you want to call the parameter? This parameter seems to make no sense for content or url I guess?

GlennRicaud:
I called it "contentPath".
And yes I did not add it to the functions attachmentUrl, componentUrl, pageUrl or url

[...]

PR #5914, opened 2017-12-22
URL Generation - Content path flag (#5845)

GlennRicaud:


GlennRicaud:
App-feature in branch issue-45 if you need to test.

Issue #5591, opened 2017-09-06
Improve unregistering of event listeners


GlennRicaud:
There is no way to unregister as an event listener.

We should:
1 - Add a method to the event lib. This could be called when an application is stopped for example
2- And/Or we could at least remove listeners from an application that has been stopped

aro:
Made ScriptEventManagerImpl concurrent (thread-safe).
Invalidating is a bit less efficient but iterating over events skips creating a new list for every event. Also it was probably broken with concurrent access. 

Invalidating is done very seldom, compared with publishing events.

Made also execution of application invalidators more robust, in case one invalidator fails the others should still be executed.

PR #5917, opened 2018-01-02
Improve unregistering of event listeners (#5591)

aro:


GlennRicaud:
Nice :+1: Tried hard to find something to complain about but in vain.

Issue #5770, opened 2017-11-08
Implement comments in Publishing Issues


alansemenov:

alansemenov:
Overall UI is looking fine (@runarmyklebust will verify the server-side code). Just a few cosmetic things:

1. When switching between Issue and Comments tab, the dialog changes its height even though both tabs are empty. We should keep the same min height of the dialog. Looks like it's the new comment section in the footer that makes the dialog jump.
2. We should use the same format of date/time stamp in issue email as we use in Content Studio, for example in the Content Grid. Now it seems to be using USA format ("1/15/18 12:10 PM").
3. Add title above the comment section ("Comments") just like we have above the issue items. When there are more than 5, the text should be "Comments (showing the 5 latest)"
4. Make sure that margin below the comment section is the same as above.
5. Change font-size of comment text in both the comment list and the new comment field to be 14px
6. When there are no comments, show a placeholder text in the middle of the tab saying "No comments yet" (same style as "You are wasting this space..."):
![image](https://user-images.githubusercontent.com/11612490/34940135-bb146c42-f9ee-11e7-9ad2-ac736c110fb5.png)
7. While you're at it, please do the same for the first tab with description for the case when description is empty. The text is "Click Edit to add issue description".
![image](https://user-images.githubusercontent.com/11612490/34940184-e8a06724-f9ee-11e7-8541-42675b58e581.png)
8. When new comment has a long text, the textarea won't resize to its original height after the comment has been posted.

![image](https://user-images.githubusercontent.com/11612490/34940614-7f4678d4-f9f0-11e7-8af4-7907d2b7a402.png)

pmi:
Fixed everything except n. 2, where @aro advised to use DateTimeFormatter.ofLocalizedDateTime() to return locale specific format instead of hard-coded one. I guess you have US locale set on your mac ?

PR #5941, opened 2018-01-12
Implement comments in Publishing Issues #5770

pmi:


alansemenov:
Found another problem. After a new issue is created, we used to send an email notification to all assignees with subject "You were assigned a new issue...". This stopped working for some reason (in the ``issue-5770`` branch).

alansemenov:
Ah, true - because we need to show name of the creator in the first email. Ok, let's keep it this way for now.

pmi:
Thanks mister !

Fixed that ;)

Issue #5940, opened 2018-01-12
Add Java method to get supported locales for a bundle


aro:
Add a method to `com.enonic.xp.i18n.LocaleService` to get the supported locale for a bundle based on the bundle name and the preferred locales (typically obtained from the HTTP request).

`Locale getSupportedLocale( List<Locale> preferredLocales, ApplicationKey applicationKey, String... bundleNames )`

Return the first of the preferredLocale that matches with the ones in the bundle.
If none of the preferred locales is supported for the bundle, return null. This is consistent with the other methods in LocaleService, where if locale passed is null, it will use the default locale of the bundle.

This logic is now in i18n lib, but it will be needed in other modules.
The result can be cached based on app, as is done with bundle loading.

PR #5947, opened 2018-01-15
Add Java method to get supported locales for a bundle (#5940)

aro:


GlennRicaud:
@aro: TODO: Add a not about "_" passed to "-"

Issue #5935, opened 2018-01-10
Support i18n attribute in content type scheme


alansemenov:
Localisation of literal fields in content type scheme will be done via new ``i18n`` attribute which will hold the key from the resource bundle. Value of the field itself will be used as a fallback value (if resource key is not found or resource bundles are not found).

Fields that will support localisation in the first version are:
- ``display-name``
- ``label``
- ``help-text``

Example:
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<content-type>
  <display-name i18n="text.content.type">My Content Type</display-name>
  <super-type>base:structured</super-type>
  <form>
    <input name="myTextField" type="TextLine">
      <label i18n="text.version">Version</label>
      <help-text i18n="text.version.help">This is the latest version</help-text>
    </input>
  </form>
</content-type>
```
REST request for the content scheme (``rest/schema/content``) should return already localised literals for content type fields. Localisation should use locale from the request header. Localisation will use local resource bundles provided in the application (same behaviour as i18n lib).

PR #5948, opened 2018-01-15
Support i18n attribute in content type scheme #5935

vbradnitski:


aro:
The content types from the list, when you click on Create content, are not translated.
From REST `/admin/rest/schema/content/all`
I think it's because it takes the app from the first content type, but there can be  types from different apps in the list.

PR #5987, opened 2018-01-22
Backend changes to comments

pmi:
- Changes in backend to move comments out of issues and store them under /issues/issue-X/comment-Y


pmi:
Thanks mister ! 

Used issue ids instead of names, and then fetched the issue node by id to get the name.

The comment name is unique in a single issue scope only, but it's not used for anything except the node name to store it in the same fashion as issues are stored.

Added some itests for find comments and find resource endpoint as well.

Good idea about ordering ! Bubbled that one to query.

About chopped builder methods, I actually liked it with the dot on the new line, but our codestyle apparently didn't, so I do it like you say because it's much nicer than one huge line anyway :)

sigdestad:
Is this stuff merged to master now then?

pmi:
@sigdestad nope, it was merged to the issue comments epic branch #5770, which has tasks to be done before merging to master

Issue #5995, opened 2018-01-23
Localisation of built-in content types


aro:
Set localisation keys in the built-in content types:
- displayName
- description
- input labels
- input help-text

Add the i18n properties files with the keys for 2 languages, and empty files with placeholders for the other languages supported in xp-apps.

Research how to resolve the bundles, since the appKey for built-in types is a "virtual" name (e.g. "base", "portal", "media")

Issue #5999, opened 2018-01-24
Create system app for localisation of built-in types


aro:
The app will be marked as system, so it does not appear on application resource.
It will be used for localisation texts for built-in content types, mixins and macros.

PR #6009, opened 2018-01-25
Localisation of built-in content types (#5995)

aro:


GlennRicaud:
Looks good. To be modified by #5999

PR #6016, opened 2018-01-29
Url methods of the Portal lib convert all "unknown" params properties to url parameters

vbradnitski:


sigdestad:
A bit unsure what has happened here? 
It is a basic idea with url functions is that users can pass any parameter (excluding _ prefix? ) and they will be naturally added as url params. Is this functionality still intact?

It is also documented here: http://xp.readthedocs.io/en/stable/reference/view-functions/assetUrl.html

alansemenov:
We have mismatch in the docs then, because [here](http://repo.enonic.com/public/com/enonic/xp/docs/6.13.1/docs-6.13.1-libdoc.zip!/module-portal.html#.assetUrl) it says that custom parameters should be in the ``params`` object, and that makes sense.

GlennRicaud:
First of all, I know that this code is way overdesigned and that to refactor it would take a lot of time.
But I am not a big fan of this whitelist check. We lose all the advantage of Javascript.
@aro Is it possible to just not append the underscore params when it generates the URL instead?

If not, then I will adapt the list with a parameter that is actually handled for some of them but that we did not want to be documented yet. Otherwise yes it works like expected.

GlennRicaud:
Good point from @aro about intended underscore params.
Merged.

Issue #6039, opened 2018-02-13
Enforce email uniqueness for users in Auth API


aro:
Currently it is possible to create 2 users with the same email, using either JavaScript or Java API.
The only validation is done in the UI in Users app.

Add a check to enforce the email is unique in a user store.
This should be fixed in `SecurityService.createUser()` and `SecurityService.updateUser()`. And throw an exception if the update is in conflict with an existing user.

This issue was reported in support case 2331

PR #6054, opened 2018-02-16
Enforce email uniqueness for users in Auth API (#6039)

aro:


GlennRicaud:
The .lock should be outside of the try

````
     Lock l = ...;
     l.lock();
     try {
         // access the resource protected by this lock
     } finally {
         l.unlock();
     }
````

aro:
Good catch. Updated.

GlennRicaud:
👍

Issue #5956, opened 2018-01-17
Change email subject for the case when an issue gets assigned to a user


alansemenov:
If I first create an issue with no assignees and then assign it to a user, then he will get an email with the subject saying "Issue <blabla> was updated...". It comes as a surprise because he doesn't even know what this is about and why he got this email at all.

So, **in this case** the subject should be "You were assigned a new issue..." - same as the one we send when a new issue is created and assigned to the user at the same time. With the following updates the subject should be as now ("...was updated..."), it's only when the user was added to the list of assignees that the subject should be different.

sigdestad:
Lets prioritize the other "issue" tasks first, and deal with optimizing the mail when they are completed..

PR #6055, opened 2018-02-16
Change email subject for the case when an issue gets assigned to a us…

pmi:
…er #5956


pmi:
@runarmyklebust Did as you suggested, but it seems that new factory is not injected to IssueResource preventing it from initializing :( Could you take a quick look if this is the right way to do it ?

runarmyklebust:
Its because its not a service (implementing an interface), so the @Reference-notation does not pick it up. 

Anyway; refactor it so that the factory is not a service, but a is injected with the stuff it needs to create the params, e.g

```
IssueNotificationParams params = IssueNotificationParamsFactory.create().
  issue(issue).
  comments(comments).
  recipients(...).
  url(url).
  securityService(service).
  contentTypeService(service.).
  securityService(serser).
  build(). 
  execute(); <-- Here is the resolving stuff for the params going on

```

pmi:
Thank you sir ! Will do !

Issue #4801, opened 2017-04-10
Node lib - Inherit permissions


GlennRicaud:
Current situation: By default, if the permissions are not defined, the node is created with the following permissions: [role:system.admin[+read, +create, +modify, +delete, +publish, +read_permissions, +write_permissions], role:system.everyone[+read]]

@runarmyklebust : It seems wrong. How can I create a node that simply inherits the permissions? That should be the default behaviour

PR #6056, opened 2018-02-16
Node creation - Default permissions (#4801)

aro:


GlennRicaud:
I thought it would be more difficult. Nice :+1:
I wanted to complain that it was _inheritsPermissions and not _inheritPermissions like contentLib, but the field name is actually _inheritsPermissions so it makes sense. So nothing to complain about :(

Issue #6059, opened 2018-02-16
Improve error message when moving a content to a folder with same name


pmi:
@aro commented on [Wed Feb 14 2018](https://github.com/enonic/xp-apps/issues/574)

In Content Studio:
- Create a content /mycontent
- Create a folder /myfolder
- Add a content with the same name inside the folder: /myfolder/mycontent
- Select /mycontent and try to move it to /myfolder/

The move operation fails, which is correct. But the feedback is not very good. The error message shows "Content could not be moved", but it is in green like a successful result. Also the message could be more specific, when moving to a parent containing hundreds of children it is not obvious why it fails.

This issue was reported in a support: ticket 2334

---

@alansemenov commented on [Wed Feb 14 2018](https://github.com/enonic/xp-apps/issues/574#issuecomment-365534537)

@vbradnitski this is very connected to https://github.com/enonic/xp-apps/issues/505. we need to pass type (success/warning/error) in the server-side notification and display client-side notification in correct color.

I'm not sure about the more specific message though, I don't think we can figure out why move operation failed. If we can - improve that too.

---

@alansemenov commented on [Thu Feb 15 2018](https://github.com/enonic/xp-apps/issues/574#issuecomment-365859461)

@pmi see the last comment - we have two issues with the same problem.

PR #6060, opened 2018-02-16
Improve error message when moving a content to a folder with same nam…

pmi:
…e #6059


alansemenov:
We should not send the literal error message to the client. What we need is to send the task execution status. What we get now is this:

image

What we need in addition is a ``progress.status`` which can be either a int constant (0 - success, 1 - warning, 2 - error) or a string constant ("SUCCESS", "WARNING", "ERROR").

In frontend we will give different error messages based on what kind of operation it was and what status it returned.

runarmyklebust:
Pavel; 
The ProgressReporter-interface is a bit lacking now, so to overcome this issue with not too much work;

1) Refactor according to above complaints into separate result-object etc
2) Implement a toJson in the result, with the following fields (the description created by a separate generator-class)

`{
  "state" : "SUCCESS|WARNING|ERROR",
  "description" : "Your detailed message here"
}`


Serialize this to string into the info-field of the progress-reporter, and this must be handled in the UI to ensure correct color etc @alansemenov

alansemenov:
@pmi We are not going to take this into 6.14 - too many risky changes - so please wait with fixing this and prioritise https://github.com/enonic/xp/pull/6055 and bugs with 6.14.0 milestone

pmi:
@runarmyklebust s'il vous plait ;)

Issue #6075, opened 2018-02-22
Incorrect type of notification about failed operation when the repo is in readonly mode


pmi:
@alansemenov commented on [Thu Jan 25 2018](https://github.com/enonic/xp-apps/issues/505)

When the repo is in readonly mode (``./toolbox.sh set-read-only -a su:password -b -r cms-repo true``) and user attempts to move, duplicate or delete content, he gets a notification about failed operation, see below. However, the color of this notification is wrong - it's an error, so it should be red and not green.

![image](https://user-images.githubusercontent.com/11612490/35388498-5a5ed804-01d4-11e8-9edb-85010be83a82.png)

PR #6076, opened 2018-02-22
 Incorrent type of notification about failed operation when the repo is in readonly mode #6075

pmi:


alansemenov:
Fixed in https://github.com/enonic/xp/pull/6060

Issue #5925, opened 2018-01-08
Toolbox CLI - Load of dump with skipped versions fails


GlennRicaud:
Using Toolbox CLI, the following throws an error:
```
toolbox dump -a su:password -t mydump --skip-versions
toolbox load -a su:password -s mydump
```

It also seems that the options --max-versions do not work as planned also. But I am not sure about this second one. I might have misunderstood.

PR #6078, opened 2018-02-22
Toolbox CLI - Load of dump with skipped versions fails (#5925)

runarmyklebust:
Allow for missing versions-file


GlennRicaud:
Merged.

But while testing, I had some results that I was not expecting. @rmy Can we discuss and test together the whole dump loading without versions on wednesday

GlennRicaud:
Apologies to @rmy.
It is actually a bug in the content details panel or the Version history widget.
It is not updated when you deselect and reselect the content. You have to click on another content to be refreshed.

Issue #6079, opened 2018-02-22
lib-admin.getHomeToolUrl method should support type parameter


alansemenov:
Standard Id Provider needs absolute URL to the Home app to redirect after logout.

PR #6080, opened 2018-02-22
Support url type in lib-admin.getHomeToolUrl (#6079)

alansemenov:


GlennRicaud:
:+1: for creating the issue.

The code is not correct though. It does rewriting twice (in the java helper first and then in the portal lib).
I replaced by a simple call to portalLib.url there.

The Java helper was created when the function did not exist. I think we can remove most of the helper now and reuse portalLib.url
I will create a refactoring issue for 6.15

Also added a Gradle dependency that srs forgot to add.

Merged.

Issue #6146, opened 2018-04-03
ApplicationEvent - Add systemApplication flag


edloidas:
@alansemenov commented on [Wed Mar 28 2018](https://github.com/enonic/xp-apps/issues/711)

1. Open the Applications module to see the list of installed applications
2. Delete ``app-main``'s jar file from the deploy folder
3. Build and redeploy ``app-main``
4. Note that it will appear in the grid, even though it shouldn't (because of the ``systemApp = true`` flag in `` build.gradle``).
5. Refresh the page. Now the app is gone.

Same with ``app-standardidprovider`` and other bundled xp apps.

![image](https://user-images.githubusercontent.com/3728712/38134031-2cc1774e-341a-11e8-974c-ca6f376b8166.png)

PR #6147, opened 2018-04-03
System apps pop up in the list of installed applications after deploy…

edloidas:
… #6146

Removed system applications from dispatchers of bundle events.


GlennRicaud:
This check should not be done in a generic class like ApplicationEventDispatcher and also not in the core layer.
For example a similar check is done com.enonic.xp.admin.impl.rest.resource.application.ApplicationResource::list

Please move the check to a more appropriate place

PR #6156, opened 2018-04-05
System apps pop up in the list of installed applications after deploy…

edloidas:
… #6146

Added system application flag.


GlennRicaud:
:+1:

Issue #6172, opened 2018-04-16
More dynamic update of progress bar in modal dialogs


alansemenov:
It seems like we only update the progress bar in modal dialogs on 50% and 100%.

It's hard to notice on small number of items but when there's a big batch to process then it becomes obvious. This is not good since it can take some time before 50% of the items are processed and before that it looks like the process has frozen since nothing is happening until 50% is reached. We should update the progress bar more dynamically, ideally every second or so.

PR #6192, opened 2018-04-20
More dynamic update of progress bar in modal dialogs #6172

edloidas:
Added missing report for the pre-publish content processing.

_Since the publishing consist of two steps, it'll make the progress bar update more smoothly (it will no longer jump from 0% to 50% and then instantly to 100%). But the pre-publish updates are slower than publishing itself, so the progress bar will be filled faster after 50% because we depend on the amount of the processed content._


GlennRicaud:
I was going to complain that with this change the progress count would be the double of the the total.
But it was actually bugged before, so it is actually more correct now.
But the progress count will only go to half of the total if the contents were already published though. To be fixed in another issue

It is also true that the set publish info takes most of the time for first time publish.

Validated

Issue #6154, opened 2018-04-05
Allow wildcards in allowContentType config of ContentSelector


alansemenov:
We will allow wildcards (``*`` and ``${app}``) inside ``allowContentType`` config of ContentSelector and its deriving input types (like ImageSelector and MediaSelector).

So it should be possible to write, for example, something like 
```
<allowContentType>com.enonic.app.*:media:*</allowContentType>
```
which will allow all content types of type **media** from all *com.enonic.app* apps.

Or you can use ``${app}`` as a wildcard to allow content types only from the same application:
```
<allowContentType>${app}:*</allowContentType> 
```

PR #6196, opened 2018-04-24
Allow wildcards in allowContentType config of ContentSelector #6154

pmi:


alansemenov:
@pmi  A few issues that I found.

1. It looks like `{$app}` only works for the first app assigned to the site. It should work for all (so it's a bit different from `{$site}` which is always one value). I didn't manage to find a regex that would give me content types only from applications assigned to the site (tried ``{$app}``, ``^{$app}``, ``^{$app}:*`` etc.)

2. It looks like app name is not considered to be a part of content type when parsing the value of `allowContentType`. For example,
```xml
  <allowContentType>htmlarea</allowContentType>
```
will give me the list of contents with `com.enonic.features:htmlarea` content type.

However, if I explicitly include app name into the regex, then it stops working:
```xml
  <allowContentType>com.enonic.features:htmlarea</allowContentType>
```


By the way (and note to self for the future docs), exclusion with regex seems to work fine - for example

```xml
  <allowContentType>((?!htmlarea).)*$</allowContentType>
```
will exclude content types with substring "_htmlarea_".

sigdestad:
Hmm.... Looking at your comments here Alan. 
My plan was that ${app} is _only_ suppose to be the name of the app that the current content type comes from, not all apps on a site etc. So it is  basically app in context.

sigdestad:
also, I'm not sure about writing "htmlarea" only shoud work? Does it work today? If so the ${app} has little value I guess?

alansemenov:
Hmm, I'm quite sure I posted the answer yesterday but apparently it got lost. 

Yes, `${app}` works fine - I used it the wrong way, as `{$app}` and that's why it didn't work. Tested it again yesterday and all seems to be working nicely (also the exclusion).

> also, I'm not sure about writing "htmlarea" only shoud work? Does it work today? If so the ${app} has little value I guess?

It works because _htmlarea_ is a valid regex saying "give me content types from all apps with substring _htmlarea_". Using `${app}` helps to limit this list down to only content types from the app in context.

So it's just for @aro to verify the code and we can merge.

alansemenov:
> My plan was that ${app} is only suppose to be the name of the app that the current content type comes from, not all apps on a site etc. So it is basically app in context.

I see, it works as intended then. I also figured that I was using ``${app}`` incorrectly, as ``{$app}``. 

> also, I'm not sure about writing "htmlarea" only shoud work? Does it work today? If so the ${app} has little value I guess?

"_htmlarea_" works because it's a valid regex that means "return all content types from all apps with _htmlarea_ substring". Using `${app}` in addition will help to narrow down to content types in the context of current app, f.ex. `${app}:*htmlarea*`.

I tested it again and it seems to work fine, so it's just to wait for @aro to verify the code.

Issue #6217, opened 2018-05-03
Add request context info to Tracing 


aro:
Add information from the current request and response to the Trace object.

jsi:
Hmmmm.  I do not like this!  It was done in a way that reduce the code coverage with 0.02%  Is it possible to bring the code coverage back up in some way?  Please consider adding tests!

GlennRicaud:
This modification should not require a new test.

What could be done to increase the coverage, and was proposed in the sprint meeting, is to dedicate time to write tests of portion of code not tested.

PR #6219, opened 2018-05-03
Add request context info to Tracing (#6217)

aro:


jsi:
Was this PR approved with 0.02% decrease in code coverage???  That is unacceptable!  Please add tests to bring the code coverage back up!

Issue #6214, opened 2018-05-02
New implementation of x-data mixins


alansemenov:
**As an application developer** The use of mixins for both x-data and templating is unclear. As such I want to be able to declare x-data directly in the folder called "x-data/" (as opposed to current "_mixins_"). 

New x-data schema should be similar to the existing one except ``<x-data></x-data>`` replacing ``<mixin></mixin>`` as root tags.

New x-data will be referenced from content-types and sites in in the following way:
```xml
<x-data name="menu-item"/>
```
(as opposed to current
```xml
<x-data mixin="menu-item"/>
```
)

NB! Existing mixins from the "_mixins_" folder should still be supported, until current implementation is dropped. 

There should be no changes to mixins used **inline**.

**Conditions of satisfaction:**
* I can define x-data schemas in the _x-data_ folder (similar to current definition in the _mixins_ folder: one folder per mixin, mixin name is its folder name, mixin icon is supported etc.)
* New mixin schema is similar to current schema except for ``<x-data></x-data>`` replacing ``<mixin></mixin>`` as root tags.
* X-data from the _x-data_ folder are referenced in via ``name`` attribute on the ``x-data`` field
* X-data from the _mixins_ folder are plugged in via ``mixin`` attribute on the ``x-data`` field (current implementation)
* If there are two mixins (one with old implementation from _mixins_ folder and second with new implementation from _x-data_ folder) with the same name plugged in on the same content type, the preference should be given to the new implementation
* There will be no changes to Inline mixins (this will be supported in the future too):  ``<inline mixin="menu-item"/>``
* New implementation of x-data is covered by tests in the backend
* X-data schemas can reference mixins inline too (just like content types)

PR #6221, opened 2018-05-04
New implementation of x-data mixins #6214

vbradnitski:


alansemenov:
> Not sure if I understand the comment, but I thought the plan was that it will still work using the "mixin" attribute, to maintain backwards compatibility.

We will support both ways until 7.0. After that, schemas from the ``mixins`` folder will only be supported for  inline mixins (``<inline mixin="my-mixin">``).

Issue #6215, opened 2018-05-02
Support content type filter in x-data schema


alansemenov:
**As an application developer**, For a specific x-data, I want to be able to specify a filter for content types that shall be able to use it.  I want to be able to either explicitly specify content types to allow or use a regex to allow content types based on a specific pattern (see https://github.com/enonic/xp/issues/6154).

Use the mixin only on the _landing-page_ content type :
```xml
<?xml version="1.0" encoding="UTF-8"?>
<x-data>
  <display-name>Menu</display-name>
  <allowContentType>landing-page</allowContentType>
  <items>
...
  </items>
</x-data>
```

Use the x-data on the _landing-page_ content type and any content type from a specific app containing _media_ substring:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<x-data>
  <display-name>Menu</display-name>
  <allowContentType>landing-page</allowContentType>
  <allowContentType>^app.enonic.com.features:media*</allowContentType>
  <items>
...
  </items>
</x-data>
```

Use the mixin on all content types but only from the current app:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<x-data>
  <display-name>Menu</display-name>
  <allowContentType>^{$app}:*</allowContentType>
  <items>
...
  </items>
</x-data>
```

Allow all content types except for the ones containing _htmlarea_ substring:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<x-data>
  <display-name>Menu</display-name>
  <allowContentType>((?!htmlarea).)*$</allowContentType>
  <items>
...
  </items>
</x-data>
```

**Conditions of satisfaction:**
* I'm able to specify one or more patterns filtering allowed content types via ``allowContentType`` field in the new ``<x-data>`` schema
* This does not affect the old schema setup referencing  ``<mixin>``(from site or content type directly). However, if an x-data is referenced, the filter must be applied
* Implement support of ``{$app}`` macro inside the pattern to be able to refer to content types from the same app
* If I add an ``x-data`` to ``site.xml`` or a content type scheme and the content type doesn't match this mixin's  ``allowContentType`` pattern, then the mixin should not be rendered in the Wizard.
* Filtering of content types should be covered by sufficient tests

sigdestad:
I'm just curious if it would be better to call it contentTypeFilter instead of allowContentType? Maybe filter is too unclear?

PR #6245, opened 2018-05-14
Support content type filter in x-data schema #6215

vbradnitski:


vbradnitski:
1. List and renaming are done.
2. I tried to do it with contentTypeSerice link in MixinService, but ContentTypeService already has mixinService, so container fails trying to initialize it. I can move ContentTypeNameWildcardResolver to impl and make an interface for it in api if needed, or should we find other way to do it?

aro:
Ah, it would create a cycle between `ContentService` and `MixinService`.
I would move the logic to `MixinResource`, since it's an admin feature mainly. And keep `MixinService` clean and simple. If we see this will be used in api later, we can always refactor it.

So move `filterMixinsByContentType` method to `MixinResource`, and put `ContentTypeNameWildcardResolver` class in the same package.

Issue #6233, opened 2018-05-08
Fetch the list of open issues by content Id


alansemenov:
Implement REST service that will be called to fetch the list of open publishing issues by contentId (or design it to support an array of contentId, for the future). The list should contain issue Id, issue title and created date.

PR #6266, opened 2018-05-22
Fetch the list of open issues by content Id #6233

pmi:


jsi:
What is happening here?  Why has the code coverage been reduced so severely?

runarmyklebust:
I think its not correct @jsi - its the stuff that is not merged yet that differs..

jsi:
Okay.  @runarmyklebust , if you say so.
But, how old is the branch?  Normally, the problem with comparing code coverage percentages, is that the current branch is compared to current master, instead of the latest commit on master that is common to both master and the branch.  Then it may appear as if code coverage has gone down, even if it in fact have stayed the same or even increased.  -  But, it is a long time since we had code coverage below 78,6%, so unless this branch is really old, it is most likely reducing code coverage, although, maybe not -0,06%.

Either way.  Please, just be careful to make sure the code coverage is not reduced when this PR is merged to master!

GlennRicaud:
Changing FindIssuesParams.from/size from int to Integer is the correct solution.
But adapt the construction of FindIssuesParams in IssueResource.countIssues to have a count of 0.

At the same time, use IssuePropertyNames.PUBLISH_REQUEST in the definition IssuePropertyNames.PUBLISH_REQUEST_ITEM_ID

pmi:
Good points, @GlennRicaud ! Updated PR

Issue #6269, opened 2018-05-23
Remove disabling of cluster when health check fails


aro:
It looks like when the cluster gets deactivated due to member mismatch, it causes the getNodes method in Ignite cluster to fail. So the cluster never recovers the activated state.

```java.lang.IllegalStateException: Grid is in invalid state to perform this operation. It either not started yet or has already being or have stopped [igniteInstanceName=enonic-xp-ignite-instance, state=STOPPED]```

PR #6272, opened 2018-05-24
Remove disabling of cluster when health check fails (#6269)

aro:


GlennRicaud:
Can you solve the conflict first (Because of the other PR)

GlennRicaud:
Yes, it has been discussed on Tuesday with Rmy.
It does not seem necessary and caused problems every time a new node was started or stopped.

Merged.

During previous tests, very often the Ignite cluster stayed incoherent. Not just during a split second.
Some Ignite nodes did not manage to reconnect or crashed because an action took too much time.
So we will have to test more how the nodes behave in these cases.

PR #6275, opened 2018-05-28
Enable exclusion of child content when duplicating a content item

alansemenov:


GlennRicaud:
The code looks perfect. Very clean.
@alansemenov : But are you sure about the branches ? Master does not seem to have the UI adapted yet (so I could not really test). Was it not supposed to be in the issue-6259 and then be pulled to issue-6260?

I also saw a preexisting bug on the progress report for duplication. But I will create a bug issue for this one.

Issue #6274, opened 2018-05-28
Cluster defined as operational while nodes are missing


GlennRicaud:
Cluster started with 1 node / 3. The cluster is defined as operational...

09:48:23.334 INFO  o.a.i.i.m.d.GridDiscoveryManager - Topology snapshot [ver=1, servers=1, clients=0, CPUs=8, heap=3.9GB]
09:48:23.335 INFO  c.e.x.c.impl.ClusterManagerImpl - Adding cluster-provider: ignite
09:48:23.335 INFO  c.e.x.c.impl.ClusterManagerImpl - Activating cluster providers
09:48:23.336 INFO  c.e.x.e.impl.ElasticsearchCluster - Cluster operational, register elasticsearch-client
09:48:23.336 INFO  c.e.xp.ignite.impl.IgniteCluster - Cluster operational, register ignite
09:48:25.068 WARN  o.a.ignite.internal.GridDiagnostic - Default local host is unreachable. This may lead to delays on grid network operations. Check your OS network setting to correct it.
09:48:28.549 ERROR com.enonic.xp.core.repo - [com.enonic.xp.repo.impl.repository.RepositoryServiceImpl(91)] The initialize method has thrown an exception
org.elasticsearch.ElasticsearchTimeoutException: Timeout waiting for task.
	at org.elasticsearch.action.support.AdapterActionFuture.actionGet(AdapterActionFuture.java:73)
	at org.elasticsearch.action.support.AdapterActionFuture.actionGet(AdapterActionFuture.java:65)
	at org.elasticsearch.action.support.AdapterActionFuture.actionGet(AdapterActionFuture.java:55)
	at com.enonic.xp.repo.impl.elasticsearch.IndexServiceInternalImpl.isMaster(IndexServiceInternalImpl.java:98)
	at com.enonic.xp.repo.impl.repository.SystemRepoInitializer.isMaster(SystemRepoInitializer.java:86)
	at com.enonic.xp.init.Initializer.initialize(Initializer.java:18)
	at com.enonic.xp.repo.impl.repository.RepositoryServiceImpl.initialize(RepositoryServiceImpl.java:68)

GlennRicaud:
The problem here is the issue-6269.
The discussion was to remove the disabling of cluster for members mismatch and not the entire enabling/disabling mechanism:

* Rollback issue-6269
* Remove the members mismatch check

GlennRicaud:
Original timer solution was wrong. 
[...]

PR #6278, opened 2018-05-28
Cluster defined as operational while nodes are missing (#6274)

GlennRicaud:
See comments on main issue for more information


GlennRicaud:
Agreed for the keeping and only logging.
Will fix

GlennRicaud:
Updated with a magnificient triliean status @runarmyklebust

Issue #6262, opened 2018-05-18
Same field-set and child item-set name-attribute in a contenttype throws exception in the content.modify API


RF0:
Enonic XP 6.14.2

In a contenttype, when a field-set name-attribute is the same as the name-attribute of a child item-set, it works in Content Studio, but produces exception when the content is modified with the content.modify API call.

Exception Message:
FormItem [] in [attributes] is not of type Input: com.enonic.xp.form.FieldSet

Example schema (renaming field-set name to f.x. inputAttributes will avoid exception)
[Real life example](https://bitbucket.org/openxp/freeform/src/83cba3cfe15766848b86f9f1b5b2b58075cdda94/src/main/resources/site/mixins/freeform-elements/freeform-elements.xml#lines-13)

```xml
<field-set name="attributes">
    <label>Input attributes</label>
    <items>
        <item-set name="attributes">
            <label>attributes</label>
            <occurrences minimum="0" maximum="0"/>
            <items>
                <inline mixin="freeform-name-value"/>
            </items>
        </item-set>
    </items>
</field-set>
```

sigdestad:
We must "disable" names from field-sets in 6.15!!

alansemenov:
"disable" as in "stop using" them? we cannot remove the attribute from the schema as of yet since it will break down things, but we can stop using it in the code I guess.

sigdestad:
The task is basically to make the name attribute optional in fieldsets i 6.15, for 7.0 we can drop it completely.

aro:
We cannot make name optional in fieldset now without breaking things. The FieldSet class in Java extends from FormItem which has a required name field. 
We can do that change in 7.0

I made a pull request which fixes the issue in the JSON to PropertyTree resolving, based on a form schema. So that it traverses the form ignoring the name of the fieldsets. This will solve this issues and other related that have been reported in Discuss.

PR #6294, opened 2018-06-01
Same field-set and child item-set name-attribute in a contenttype thr…

aro:
…ows exception in the content.modify API (#6262)


GlennRicaud:
I agree with the problem and the skipLayout solution.
But I would make the caller JsonToPropertyTreeTranslator::getInput be the one deciding to skip layouts.

I expect FormItems::getFormItem and FormItems::getInput to return the same item for a same path.
Plus you might want to get an input by including the name of the layout in the path.
It is the context that decides that you should skip the layout because you lost this information.

aro:
Agreed. Updated PR.

GlennRicaud:
Good

PR #6298, opened 2018-06-05
Make language text in schemas optional when setting i18n attribute (#…

aro:
…6148)


GlennRicaud:
Tested. Looks good and also cleaner :+1:

Issue #6300, opened 2018-06-05
Applications main.js executed before all bundles are started


aro:
While running in cluster, the applications main.js are executed before some of the services (e.g. SecurityService) are started, causing the main.js to fail.

Review start levels of the bundles and make sure the applications are installed at the end, after all other services/bundles are ready.

This is more visible in a cluster due to timing, but it is probably also affecting single node installations.

PR #6302, opened 2018-06-05
Applications main.js executed before all bundles are started (#6300)

aro:


GlennRicaud:
I was hoping that this would solve the problem of MainExecutor initialized too late but it does not.
MainExecutor is still initialized after StoredApplicationsDeployer

Issue #6306, opened 2018-06-06
Reduce verbosity of Ignite logging


aro:
Try to reduce of the noise in the logs from Ignite during start up.

According to [their docs](https://ignite.apache.org/releases/latest/javadoc/org/apache/ignite/IgniteSystemProperties.html) setting system properties should disable some of it: 

```
IGNITE_PERFORMANCE_SUGGESTIONS_DISABLED
IGNITE_QUIET
IGNITE_DEV_ONLY_LOGGING_DISABLED
IGNITE_NO_ASCII
IGNITE_UPDATE_NOTIFIER
IGNITE_WAL_LOG_TX_RECORDS
```

PR #6308, opened 2018-06-06
Reduce verbosity of Ignite logging (#6306)

aro:
Part of the logging is disabled by config in logback.xml


GlennRicaud:
Good

GlennRicaud:
We will have to remember to write the logback diff in the update steps

Issue #5928, opened 2018-01-10
Macro parameters from TextLine, TextArea are converted with Html Entities


RF0:
When parameters from Macro TextLine or TextArea contain special characters like f.x. æøå the context.params object in javascript controller delivered the values escaped as HtmlEntities &aelig;&oslash;&aring;

Either this is a bug, or it should be documented in the macro documentation that this is the case, and how to best decode these back to normal text.

Tested in XP 6.11.0 and 6.13.0.

<img width="1283" alt="screen shot 2018-01-10 at 11 36 39" src="https://user-images.githubusercontent.com/951293/34768396-00f6e11a-f5fb-11e7-9b64-d39e00eccf28.png">

RF0:
The code in the screenshot is from [this repo](https://github.com/storebrand/delphi.ny/tree/master/src/main/resources/site/macros/imageLink).

PR #6316, opened 2018-06-07
Macro parameters from TextLine, TextArea are converted with Html Enti…

aro:
…ties (#5928)


GlennRicaud:
Could not test manually since the build for lib-admin-ui does not work properly.
But the code looks good. Merged

Issue #6157, opened 2018-04-06
Node data information missing for root node 


aro:
Making a modify operation and passing the root node path "/" on a repository, returns the node as if it was successfully updated. But the changes are not actually stored.

GlennRicaud:
The changes are stored

PR #6317, opened 2018-06-08
Changes made to the root node are ignored (#6157)

aro:


GlennRicaud:
Good. Merged.
I will rename the issue

Issue #6313, opened 2018-06-07
Cluster configuration


aro:
Changes in the cluster configuration:

- com.enonic.xp.cluster.cfg
```
#node.name=<generated-UUID>
#discovery.unicast.hosts=127.0.0.1
#discovery.unicast.hosts=192.168.0.41,192.168.0.42
#cluster.enabled=false
```

Use ```cluster.enabled``` as a global switch for Ignite and ES. 
In ES use it instead of ```node.local```. The boolean value should be inverted.

- Remove ```node.local``` from com.enonic.xp.elasticsearch.cfg

PR #6318, opened 2018-06-08
Cluster configuration (#6313)

aro:


GlennRicaud:
- Please remove the line "discovery.unicast.hosts=192.168.0.41,192.168.0.42".

Issue #5699, opened 2017-10-16
ContentNodeHelper.translateNodePathToContentPath resolves wrong path


runarmyklebust:
ContentNodeHelper.translateNodePathToContentPath seems to resolve a relative path instead of a absolute content-path:

     final ContentPath contentPath =
            ContentNodeHelper.translateNodePathToContentPath( NodePath.create( "/content/fisk/myContent" ).build() );

This yields "fisk/myContent" but should be /fisk/myContent
Apparently this doesnt make anything fail, but it should be corrected.

PR #6319, opened 2018-06-08
ContentNodeHelper.translateNodePathToContentPath resolves wrong path …

aro:
…(#5699)


GlennRicaud:
Not tested but the code looks good

Issue #6324, opened 2018-06-11
[MS Edge] Attachment content has incorrect label and produces 404 error on click


ashklianko:
@sgauruseu commented on [Fri Apr 20 2018](https://github.com/enonic/lib-admin-ui/issues/449)

1. Open CS in `Edge`
2. Select `Features`, click on `New`, select `Attachments`
3. Upload an  attachment 
**Bug**: path to the local resource is displayed

![att](https://user-images.githubusercontent.com/3728712/39048035-ce5cb736-44a4-11e8-9fd8-19215f77c31c.png)

4.  type name and Save the content
5. Click on the attachment-link
**BUG**:  404 Error appears in the browser 
![404](https://user-images.githubusercontent.com/3728712/39048078-ed129222-44a4-11e8-9800-2993c8e391aa.png)

PR #6325, opened 2018-06-11
Edge specific bug - incorrect label in attachment-content and 404 Err…

ashklianko:
…or, when click on it #6324

-No way to fix this on frontend, have to filter value on backend


ashklianko:
Will check!

ashklianko:
Extracting name out of form as in create/update media works perfect!

Issue #6327, opened 2018-06-11
Error notification message appears when Deleted content has been published


pmi:
@sgauruseu commented on [Wed May 02 2018](https://github.com/enonic/app-contentstudio/issues/64)

1. Select a Deleted-content in the grid
2. Click on `Publish `button

Bug: error notification appears 

![image](https://user-images.githubusercontent.com/3728712/39517404-70acc3ca-4e08-11e8-8f90-d5e731b5c4d3.png)

---

@alansemenov commented on [Mon May 14 2018](https://github.com/enonic/app-contentstudio/issues/64#issuecomment-388823847)

@sgauruseu can you add more details here? when I simply delete and publish any content it works fine for me.

PR #6328, opened 2018-06-11
Error notification message appears when Deleted content has been publ…

pmi:
…ished #6327

### DIFF
--- a/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/content/task/PublishRunnableTask.java
+++ b/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/content/task/PublishRunnableTask.java
@@ -70,7 +70,7 @@ public class PublishRunnableTask
             }
             if ( deleted.getSize() == 1 )
             {
-                resultBuilder.deleted( contentService.getById( deleted.first() ).getPath() );
+                resultBuilder.deleted( result.getDeletedPaths().first() );
             }
             else
             {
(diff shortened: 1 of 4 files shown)


pmi:
Hmm, that's a good point!

We need the name of the content in the message in case there's just 1 deleted. Otherwise we just show the number so we could probably check for the length of ids ?

pmi:
The issue was that we were not unable to fetch the deleted content name after it was deleted. And that threw an exception that was caught by try/catch where it tried to load the content to get its name and failed again bubbling it to the UI as runtime.

alansemenov:
We show simply "The item is deleted" when there's one, without the name:

image

pmi:
Hmm, @ase it's about publishing content marked for deletion. In that case **PublishTaskMessageGenerator.java** is used to generate message similarly to other runnableTasks like delete/move etc.

alansemenov:
Right, we changed it to include content name in the notification when it's single... What's important is that we give exactly the same message for the case when an offline item is deleted and for the case when an online item is marked for deletion and then published.
The issue here is how easily we can get content name if it's deleted. If it's not easy then we should change it to "The item is deleted" for both cases, it's perfectly fine.

pmi:
Okay, refactored to load path in case 1 content is being deleted only

aro:
@pmi in PublishContentCommand.doDeleteNodes, where you get the path for the deleted node, it can happen that the id is not found. In that case nodeService.getById would throw NodeNotFoundException.
Add a try-catch to avoid that if fails trying to delete a node id that doesn't exist.

Otherwise the solution looks good.

pmi:
@aro Updated !

Issue #6334, opened 2018-06-12
ContentResource - Generate unique unnamed name on update


ashklianko:
@sgauruseu commented on [Thu Apr 19 2018](https://github.com/enonic/app-contentstudio/issues/28)

@sgauruseu commented on [Mon Mar 12 2018](https://github.com/enonic/xp-apps/issues/685)

Case 1:
1. Create a site(select Features)
2. Expand the site, select Templates -folder and open the template-wizard
3. Do not type a name, but select 'main region' in controllers
4. Open one more wizard for template, do not type a name, but select 'main region' in controllers

BUG: Error 409 appears in the console, when select a controller in the second unnamed template

<img width="927" alt="templ" src="https://user-images.githubusercontent.com/3728712/37286166-41875ba0-2612-11e8-9fd6-b61d021ff03c.png">

Case 2:
1. Start to add a new folder in the root, open wizard , do not type a name but remove the Owner. Save and close the wizard

2. Open wizard for the  second folder in root dir, do not type a name, but remove Owner and try to save the folder
<img width="909" alt="folder" src="https://user-images.githubusercontent.com/3728712/37286812-242650b4-2614-11e8-9898-b769298f57d2.png">

Case 3:
-Open new site-wizard, do not type a name but select an application, press the Save button, close the wizard
-repeat the first step, warning-message appears when Save button has been pressed

PR #6335, opened 2018-06-12
Error in console and 'Notification warning' when select a page-contro…

ashklianko:
…ller in unnamed template #6334

-Issue is more common: when trying to update unnamed content it's name, consisting of '__unnamed__' + <unique_identifier>, has it's  <unique_identifier> part lost so name after update is just '__unnamed__'. Error is thrown when trying to save 2nd '__unnamed__' content under same path.
Solved by generating <unique_identifier> part for unnamed content's name

### DIFF
--- a/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/content/ContentResource.java
+++ b/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/content/ContentResource.java
@@ -123,6 +123,7 @@ import com.enonic.xp.content.ContentDependencies;
 import com.enonic.xp.content.ContentId;
 import com.enonic.xp.content.ContentIds;
 import com.enonic.xp.content.ContentListMetaData;
+import com.enonic.xp.content.ContentName;
 import com.enonic.xp.content.ContentNotFoundException;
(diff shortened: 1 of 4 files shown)


GlennRicaud:
From what I see, it is more Admin UI that does not keep the name of the content (that is correctly returned from /admin/rest/content?id=xxx) and does not pass it on the update.
Could you try solve this in the Content Wizard instead by passing the correct name?

alansemenov:
Isn't that what this PR solves: https://github.com/enonic/lib-admin-ui/pull/604?

GlennRicaud:
Perfect!
Then this PR is not necessary. Or am I missing something?

ashklianko:
Partially solves, try giving content some regular name, save it, and then delete name and save again - name will be '__unnamed__' then.

ashklianko:
Curently when creating new content it's __unnamed__xxxxxx part is being generated on backend,
this PR solves issue with deleting content's name and saving. 
We could generate unique part also on ui, but is it good idea when it is primarily generated on backend? From this point I was updating backend rather than UI.

GlennRicaud:
:+1:

ashklianko:
I should have pointed this out in description to not confuse anyone :)

Issue #6247, opened 2018-05-14
lib-content sets default value of focalPoint for new media content


alansemenov:
When content lib creates a new media content, it sets ``focalPoint`` property to ``{ "x": 0, "y": 0 }``. This should not happen, default value should not be set.
 
https://discuss.enonic.com/t/createmedia-default-focus/1442

PR #6336, opened 2018-06-13
lib-content sets default value of focalPoint for new media content #6247

ashklianko:
-now setting focal points in lib only if they are set, otherwise default values 0.5 will be used

### DIFF
--- a/modules/lib/lib-content/src/test/java/com/enonic/xp/lib/content/CreateMediaHandlerTest.java
+++ b/modules/lib/lib-content/src/test/java/com/enonic/xp/lib/content/CreateMediaHandlerTest.java
@@ -3,6 +3,7 @@ package com.enonic.xp.lib.content;
 import java.time.Instant;
 
 import org.junit.Test;
+import org.mockito.ArgumentCaptor;
 import org.mockito.Mockito;
 
 import com.google.common.io.ByteSource;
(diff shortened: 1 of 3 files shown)


modules/lib/lib-content/src/test/java/com/enonic/xp/lib/content/CreateMediaHandlerTest.java
```diff
@@ -29,11 +32,32 @@ public void testExample()
...
+
+        runFunction( "/site/test/CreateMediaHandlerTest.js", "createMedia" );
+
+        final ArgumentCaptor<CreateMediaParams> argumentCaptor = ArgumentCaptor.forClass( CreateMediaParams.class );
+        Mockito.verify( this.contentService, Mockito.times( 1 ) ).create( argumentCaptor.capture() );
+
+        assertEquals( 0.5, argumentCaptor.getValue().getFocalX(), 0 );
+        assertEquals( 0.5, argumentCaptor.getValue().getFocalY(), 0 );
+    }
+
```
alansemenov: Shouldn't we have a test that will check that when content is created without focal point then the properties will not be set? Because that's exactly what this bug is fixing.

ashklianko: This is exactly the case, when not setting from content lib, default values in the CreateMediaHandler.java are set

Issue #6206, opened 2018-04-27
Implement progress for VacuumService


runarmyklebust:
Ref this issue: https://enonic.zendesk.com/agent/tickets/2445

Implement progress for the vacuum-service; progress should indicate which task, and the task should report its own progress also.

PR #6340, opened 2018-06-14
Implement progress for VacuumService (#6206)

aro:


GlennRicaud:
Replace NodeConstants.NODE_SEGMENT by NodeConstants.BINARY_SEGMENT in UnusedBinaryFileCleanerTask

GlennRicaud:
Also, other listeners are not exactly like this. The number passed to the listener is the number of element impacted on this iteration and not the sum. Please adapt by calling with parameter 1L, like it is done for NodeExportListener:nodeExported

Issue #6342, opened 2018-06-14
List of options in content selector is empty in tree mode


vbradnitski:
@sgauruseu commented on [Wed Jun 13 2018](https://github.com/enonic/lib-admin-ui/issues/589)

This bug reproduced in Move dialog as well. List of options is empty, when dropdown handle has been clicked, so this is a problem of LoaderComboBox

1. Open a site and insert ImageComponent
2. Click on mode toggler and switch to tree mode

Bug: `No matching items`  is displayed

![image](https://user-images.githubusercontent.com/3728712/41350827-154dd862-6f1d-11e8-86a2-b5c9843a4ba0.png)

GlennRicaud:
Bug due to the issue https://github.com/enonic/xp/issues/5699
There could other places impacted.
Will create an issue to go through the list of caller of ContentNodeHelper:translateContentParentToNodeParentPath and check the impact.

PR #6344, opened 2018-06-14
List of options in content selector is empty in tree mode #6342

vbradnitski:


GlennRicaud:
Hahaha loved the old tests in ContentQueryWithChildrenTest: "/contentmy-path/*" and "/content//my-path".
Merged

Issue #6343, opened 2018-06-14
Add method to copy current Context detached from the session


aro:
When backend code is executed in a different thread than the HTTP request thread, the Context still contains a reference to the Session object. In this case the Session might have been invalidated and accessing its attributes (e.g. to get the AuthenticationInfo object) will fail silently. 
The observed behaviour when this happens is that API calls are executed as if no user was logged in.

Fix it by creating a method in ```ContextBuilder``` to make a copy of a context detaching it from the session, and moving the session attributes to the local attributes map.

Related to issue https://github.com/enonic/app-contentstudio/issues/214

sigdestad:
What are the side-effects of this? Are the context objects always immutable?

GlennRicaud:
Context object are still immutable

This problem is due to the upgrade of Jetty 9.3.5 -> 9.4.8 (well it was not correct before but Jetty was not complaining about it).
Some visible effects are the problem of inherited permissions noticed on 6.15.

But there might be others. Will create an issue to go through the Context creations once this issue is done.

PR #6346, opened 2018-06-15
Add method to copy current Context detached from the session (#6343)

aro:


GlennRicaud:
Local scope attributes have the priority on session attributes.
Please switch the order of the 2 for loops in ContextBuilder: detachSession

aro:
Good catch. Updated.

GlennRicaud:
Good :+1:

Issue #6348, opened 2018-06-15
Security Resource - Retrieve users with role filtering


pmi:
@sgauruseu commented on [Wed May 23 2018](https://github.com/enonic/app-contentstudio/issues/145)

1. Open IssueDetailsDialog, go to Assignee tab and add a user **without roles**

![image](https://user-images.githubusercontent.com/3728712/40421722-824f3e76-5e95-11e8-99da-dad8556752aa.png)

2. Issue is updated and  `1` appears in  `Assignees`
3. Close the modal dialog
4. Open the issue in IssueDetailsDialog again
Bug : no any users are present in `Assignees`

GlennRicaud:
I remove the milestone 6.15 since it has been removed in the other related issues in other repositories

PR #6349, opened 2018-06-15
Issue details dialog - users without publish rights should be removed…

pmi:
… from the Assignees selector #6348


pmi:
@GlennRicaud That's true that it can return less users after filtering than requested, but I added a pre-filtered result count to the response and there's a reason I did it that way.

Members of roles are persisted in roles ( there's also case with inderect roles through groups ) so I'm unsure if there is a way to query for users having certain roles directly using PrincipalQuery.

GlennRicaud:
Forgot about this kind of problem... I get the problem now

I take back PrincipalQuery comment. But we cannot reuse this method "findPrincipal". The totalHits and hits would be just wrong.
An API user should not have to know the implementation details.
That if he passes the "roles" parameter the totalHits has a different meaning than without the parameter.

We should create a new method with in the response a "hasMore" boolean instead of totalSize.
We can discuss the details

Issue #6354, opened 2018-06-19
Fix missing config for Ignite to do discovery in Dockered env


runarmyklebust:
There is some little piece missing to make Ignite discover in a clustered Docker environment. Its difficult to debug atm since Ignite is configured to be very non-verbose.

Theory;
* We need to implement the https://ignite.apache.org/releases/latest/javadoc/org/apache/ignite/configuration/class-use/AddressResolver.html to map from the internal to the external (publish) - address. 

In case the above is correct, we can use the publish-address in combination with the discovery_tcp_localAddress config to create this map automatically

PR #6356, opened 2018-06-19
Fix missing config for Ignite to do discovery in Dockered env (#6354)

aro:


GlennRicaud:
Longest PR test I have eve rdone but finally merged :)

PR #6379, opened 2018-06-25
Fetch dragged URLs from the ImageSelector #6378

ashklianko:
-Creating resource method for checking if content with such path already exists


GlennRicaud:
Nice. Merged.

Issue #6303, opened 2018-06-05
Sessions in Ignite - Exception logged on shutdown of server


GlennRicaud:
When shutting down the server, exceptions are logged (see below). Investigate.

2018-06-05 14:04:52,218 WARN  org.eclipse.jetty.server.session -
java.lang.IllegalStateException: class org.apache.ignite.internal.processors.cache.CacheStoppedException: Failed to perform cache operation (cache is stopped): com.enonic.xp.webSessionCache
	at org.apache.ignite.internal.processors.cache.GridCacheGateway.enter(GridCacheGateway.java:164)
	at org.apache.ignite.internal.processors.cache.GatewayProtectedCacheProxy.onEnter(GatewayProtectedCacheProxy.java:1656)
	at org.apache.ignite.internal.processors.cache.GatewayProtectedCacheProxy.remove(GatewayProtectedCacheProxy.java:981)
	at com.enonic.xp.web.session.impl.IgniteSessionDataStore.delete(IgniteSessionDataStore.java:114)
	at org.eclipse.jetty.server.session.AbstractSessionCache.loadSession(AbstractSessionCache.java:470)
	at org.eclipse.jetty.server.session.AbstractSessionCache.get(AbstractSessionCache.java:357)
	at org.eclipse.jetty.server.session.SessionHandler.getSession(SessionHandler.java:940)
	at org.eclipse.jetty.server.session.SessionHandler.getHttpSession(SessionHandler.java:545)
	at org.eclipse.jetty.server.session.SessionHandler.checkRequestedSessionId(SessionHandler.java:1642)
	at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:1543)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:166)
	at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1155)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:141)
	at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:132)
	at com.codahale.metrics.jetty9.InstrumentedHandler.handle(InstrumentedHandler.java:240)
	at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:132)
	at org.eclipse.jetty.server.Server.handle(Server.java:530)
[...]

PR #6386, opened 2018-06-26
Sessions in Ignite - Exception logged on shutdown of server (#6303)

aro:
- Set IGNITE_NO_SHUTDOWN_HOOK to true to avoid Ignite JVM shutdown hook to happen before the controlled OSGi shutdown
- Removed some logging level switches that are available in Ignite 2.5.0 but not yet in [2.3.0](https://ignite.apache.org/releases/2.3.0/javadoc/org/apache/ignite/IgniteSystemProperties.html) that we currently use


GlennRicaud:
Nice. Merged.

Issue #6345, opened 2018-06-14
Content library - findVersions, getActiveVersions & setActiveVersion


GlennRicaud:
Asked by 2 two developers from Bouvet and Tine on Slack.
Then asked again, the day after, by the same developer on Discuss: https://discuss.enonic.com/t/programmatically-work-with-versioning/1542

It should not take too much time. The methods are already there. We just have to expose them in the content library

sigdestad:
6.15?

alansemenov:
Ok

Bellfalasch:
Request for this feature dates all the way back into 2016 =)
https://discuss.enonic.com/search?q=version

Bellfalasch:
Request for this feature dates all the way back into *2016* =)
https://discuss.enonic.com/search?q=version 

Or this, from Sept *2015* where we said "It will come with 7.0 that we plan to release Q1 2016", fun dive into history. https://discuss.enonic.com/t/content-versions/217/2

sigdestad:
If there are no objectsion, lets just put this into 6.15.
@GlennRicaud , I would love to see the signature of the methods before you commit them? Also, I belive they want this for both content and node API's

GlennRicaud:
I was thinking 6.16 or 7.0 but if it is a priority I will work on it today.
Please adapt the version though. We cannot be in RC1 and at the same time not be in feature freeze.

GlennRicaud:
Been asked to stop the development for now.

For future development, start from work done in:
XP branch: issue-6345
App-features branch: issue-56

GlennRicaud:
Closing this issue. See last comment in PR for more info: https://github.com/enonic/xp/pull/6454

PR #6454, opened 2018-07-18
 Content library - findVersions, getActiveVersions & setActiveVersion (#6345)

GlennRicaud:


sigdestad:
I like the "getActiveVersion" in context of a branch... getVersions could work as a method to collect any version across all branches (the result of this could contain branch(es) a version is active in too I guess).

GlennRicaud:
It is fixed.
But, unlike the NodeVersions, I do not think that it is wise to expose the ContentVersions as they are defined today.
So I will review and merge the issue #6459, but close this one.

My problem is the definition of a ContentVersion
  I initially thought that ContentVersions were filtered by distinct modifiedTime but it is not true. That is done only at admin level
  Unlike NodeVersion it is not just version meta data but also 3 properties from the blob. (so a blob retrieval on each hit)
  The ContentVersion also lose the timestamp information (how are you supposed to order them? Not by modifiedTime since you will get multiple versions with a same modifiedTime)

PR #6455, opened 2018-07-19
Issue details dialog - users without publish rights should be removed from the Assignees selector #6348

pmi:


GlennRicaud:
@pmi: I wanted to check before changing.
Are you sure about this line in SecurityResource? Is that really what you want to do.
``unfilteredCount.set(` from );``

Here if I search for example starting from 100, I will get a result with "unfilteredSize"  >= 100 no matter the number of users.

Otherwise I will just remove this line. Then it will returned the correct unfiltered count of users.

Issue #6503, opened 2018-08-16
Page template property not indexed


GlennRicaud:

GlennRicaud:
Was a bit confusing because it was present in old data. It was correctly indexed in Enonic XP <6.6 and not since.

PR #6504, opened 2018-08-16
Page template property not indexed (#6503)

GlennRicaud:


sigdestad:
Glenn, what is this issue? All nodes are indexed - what am I missing?

GlennRicaud:
The property "page.template" not the page template itself

Issue #6512, opened 2018-08-21
Error when try to click on 'Save as Template' menu item and '_templates' folder does not exist


ashklianko:
@sgauruseu commented on [Tue May 29 2018](https://github.com/enonic/app-contentstudio/issues/161)

This issue appears when  '_templates' folder was **deleted** or site was **duplicated** without the folder
`Steps`: 
1 Open the site 
2. Press the `Save as Template` button

![image](https://user-images.githubusercontent.com/3728712/40649461-a73ca8d8-6339-11e8-8ca0-a514dd5a773b.png)

---

@alansemenov commented on [Tue Aug 21 2018](https://github.com/enonic/app-contentstudio/issues/161#issuecomment-414576707)

Catch the error and show our standard floating error message: "Failed to save the template"

PR #6513, opened 2018-08-21
Error when try to click on 'Save as Template' menu item and '_templat…

ashklianko:
…es' folder does not exist #6512

-Making template errors to be returned as 5xx errors that will be translated to Errors on ui, rather than 4xx Errors that are translated to Warnings on ui


ashklianko:
Also https://github.com/enonic/app-contentstudio/pull/375 required

PR #6522, opened 2018-08-23
Always include internal outbound references when duplicating content

vbradnitski:


GlennRicaud:
There seems to be a confusion.
This is not just a PR about the duplication dependency resolution. This is a raw merge of the app-translations prototype branch.
Please create a proper bug fix branch (based of this code of course).

About the resolution interface, why is contentIds a Map and not ContentIds in ResolveDuplicateDependenciesParams.
If you need to keep track of a mapping by content ID, then create the map inside ResolveDuplicateDependenciesCommand instead.
But there might be a reason I did not see.

Good refactoring for the builders and hashCode

Issue #6543, opened 2018-09-06
Content query with mustBeReferencedById parameter incorrectly resolves outbound dependencies


alansemenov:
When content is saved in the Wizard, we're supposed to refresh previews (both in Grid and Wizard) of those items that have direct outbound references to the updated content. It was implemented [here](https://github.com/enonic/lib-admin-ui/issues/623) and [here](https://github.com/enonic/app-contentstudio/issues/243).

However, it looks that when ``api.content.ContentQueryRequest`` is called with ``mustBeReferencedBy`` parameter, it will sometimes returns a hit even if the content in ``mustBeReferencedBy`` is not a *direct* outgoing dependency (it's not show in the Dependencies widget at all).

Most likely this happens because the two contents has some indirect relation, but we are only interested in direct reference here.

It results in wizard suddenly being refreshed when a seemingly unrelated content is modified. 

![image](https://user-images.githubusercontent.com/11612490/45148769-52786e80-b1c8-11e8-9f71-2e25a941c9f2.png)

Reported by Tine [here](https://enonic.zendesk.com/agent/tickets/2540)

PR #6557, opened 2018-09-10
Content preview is refreshed when another non-related content is modi…

vbradnitski:
…fied #6543


GlennRicaud:
See comment in https://github.com/enonic/xp/pull/6558

PR #6558, opened 2018-09-10
Content preview is refreshed when another non-related content is modi…

vbradnitski:
…fied #6543


GlennRicaud:
I was first going to complain that the real bug is in ContentQueryNodeQueryTranslator:processReferenceIds.
But after seeing the code, I understand your choice to fix it at ContentResource level and why the bug happened in the first place:
There is a confusion at multiple places in ContentService and its commands. Null and empty lists are not the same thing.
I agree. Too much/risky of a refactoring for 6.15. I will create a Refactoring issue for 7.0

Nice to think about the missing content, but sourceContent will never be null because a ContentNotFoundException will be thrown. Removed.
Also not a big fan of the exception (A content without outbound dependencies is a valid case. But that will disappear in 7.0). Removed the INFO logs tough.

Merged

Issue #6539, opened 2018-09-04
Make it possible to override host/port discovery-settings


runarmyklebust:
It should be possible to set a specific host/port-combination in the discovery-settings.
This has been possible earlier, and must be supported.

in 6.15.1/2 there are two different settings for this:

**In specific cluster type/provider config:**
``discovery.tcp.port`` - Accepts a port or range of ports

**In general cluster config:**
``discovery.tcp.hosts`` - Accepts a range of hosts without ports

It should be possible to e.g specify the discovery.tcp.hosts_port or similar in the specific cluster provider to override this with a <host>:<port> - setting

PR #6581, opened 2018-09-20
Make it possible to override host/port discovery-settings #6539

runarmyklebust:


runarmyklebust:
Lets not do this for 6.15 if we dont have to

PR #6582, opened 2018-09-20
Disable Ignite by default (#6577)

aro:


GlennRicaud:
Really tried to find something to complain about. Without success.
Merged :)

Issue #6583, opened 2018-09-20
ES&Ignite Config - Allow to define socket addresses


GlennRicaud:
#Override discovery.unicast.hosts & discovery.unicast.port
#discovery.unicast.sockets = 127.0.0.1[9300]

PR #6596, opened 2018-09-21
ES&Ignite Config - Allow to define socket addresses #6583

GlennRicaud:


GlennRicaud:
@runarmyklebust :
https://github.com/enonic/xp/pull/6593 was merged but not https://github.com/enonic/xp/pull/6596

Issue #6618, opened 2018-09-28
Define built-in image styles


alansemenov:
We will have 7 predefined image styles in XP:

Alignment:
* _editor-align-left_
* _editor-align-justify_
* _editor-align-center_
* _editor-align-right_

Image processing:
* _editor-style-original_

Size:
* _editor-width-auto_
* _editor-width-custom_

1. Define these in the core's **site/styles.xml**:
```xml
<styles css="assets/styles/styles.css">
  <image name="editor-style-original">
    <display-name i18n="style.editor.original">Original</display-name>
  </image>
</styles>
```

2. Define these in the core's **assets/styles/styles.css**:
```css

  figure img {
    width: 100%;
  }

  figure.editor-width-actual img {
    width: auto;
  }

  figure.editor-align-left {
    float: left;
    margin: 15px;
    width: 40%;
  }

  figure.editor-align-right {
    float: right;
    margin: 15px;
    width: 40%;
  }

  figure.editor-align-center {
    margin: auto;
    width: 60%;
  }
```
3. Add localisation resources (**i18n/phrases.properties**) for the style display names:
_style.editor.original=Original (no image processing)_
_style.editor.align.justify=Justify_
_style.editor.align.left=Align left_
_style.editor.align.center=Center_
_style.editor.align.right=Align right_
_style.editor.width.custom=Custom_

PR #6662, opened 2018-10-12
Define built-in image styles (#6618)

aro:


GlennRicaud:
Closing this PR and creating a new one. The change of target branch makes Github output an incorrect diff

Issue #6670, opened 2018-10-18
Event listeners may block event dispatching


ComLock:
Events are dispatched sequentially with a single thread. That makes that an event listener function may block all event dispatching in the system.

Original issue:

----
Event listeners in main.js and in tasks are affected by sleep in main.js

The code below should produce the following log:
send
got
send
got
send
got
send
got
send
got

It actually produces:
send
send
send
send
send
got
got
got
got
got

```javascript
lib.event.listener({
	type: 'custom.*',
	callback: (event) => {
		log.info(toStr({gotEvent: event}));
	}
});
for (let i = 0; i < 5; i += 1) {
	lib.task.sleep(2000);
	const event = {
		type: 'whatever',
		distributed: false, // Send only to current cluster node.
		data: {
			a: 1
		}
	};
	log.info(toStr({sendingEvent: event}));
	lib.event.send(event);
}
```

I will try a workaround by making the listener start inside a task.

ComLock:
I tried both workarounds.

Running the sleep+send as a task worked beautifully.

However

Running the listener in a task, while the sleep+send in the main.js still affected the listener.

ComLock:
So current lesson is: "Do not sleep in main.js thread"

ComLock:
Notice that when sleeping in main, Content Studio doesn't show new content before reload.

aro:
I reproduced the issue. Looks like a sleep in an application's main.js blocks all the events sent in the system, even though the events should be asynchronous.

alansemenov:
@aro @GlennRicaud what happened to this one? there's a PR but the issue was moved from "Review" to "In progress"

GlennRicaud:
Well it was good but after discussion we wanted to do more. Thus the sprint focus about events.

alansemenov:
Let's merge this and then create additional tasks later

GlennRicaud:
Reassigning to @rymsha as he has been working on related problems and has planned to refactor it in the future.

rymsha:
`sleep` must be avoided by any means. It doesn't fit in JS world. I'm discussing deprecation of this library with @sigdestad

GlennRicaud:
Not really about the sleep. More about event listener blocking the entire event propagation I think.

rymsha:
#8021 gives little bit more input which areas are affected

rymsha:
Current event flow diagram
[...]

PR #6684, opened 2018-10-23
Event listeners may block event dispatching (#6670)

aro:


GlennRicaud:
Can you add a max pool size.
I will merge then

GlennRicaud:
I will not merge this PR for now.

First because of the test failing https://travis-ci.org/enonic/xp/builds/460230666
And that the fix seems to have been : https://github.com/enonic/xp/pull/6684/commits/5ca1c1a4e5c87fb065b8248d3ad2b142b572569e
This seems to be just postponing the problem and should be investigated. We cannot fail if the number is too high. And we cannot create X threads for X events as in https://github.com/enonic/xp/pull/6684/commits/f4b5639655033b654dee9bb0eba28fd7c09f6e19. So to be investigated

Also, do we guarantee that event are delivered sequentally? If we do, this could have an impact.

PR #6699, opened 2018-10-29
Change getDependencies REST API to include dependency info

ashklianko:
…#6698

-sending back inbound/outbound dependency contents instead of just dependency items couns


alansemenov:
@aro the idea was to show the list of dependencies of specific content type in a tooltip but since it's now going to be very simple to open the list of such dependencies we can drop the tooltip.

@ashklianko can you make changes to the [PR in app-contentstudio](https://github.com/enonic/app-contentstudio/pulls) and remove the tooltip? this PR can then be closed.

ashklianko:
On it

Issue #6692, opened 2018-10-25
Content - processedReferences field & ContentProcessor


GlennRicaud:
The goal is that the links present in HtmlAreas should be handled as references.
The chosen solution is to have a new Content field "processedReferences" that will be translated to Reference at node level. This field will be populated by a ContentProcessor that will visit the HtmlArea properties and add references to processedReferences

* Create a new field "processedReferences" of type ContentIds in com.enonic.xp.content.Content
  * This field will not be transmitted to the rest api or the controllers (Do not add this to ContentMapper or ContentSummaryJson)
* Create a new ContentProcessor (in portal-impl), that will
  * Visit the HtmlArea properties of the content
  * Retrieve the references contained (A regexp exist in HtmlLinkProcessor.CONTENT_PATTERN) 
  * Add the references to "processedReferences"

sigdestad:
Do we have a single field, or one per editor? Is the name processedReferences good? 

Also, are there any problems related to updating content with this approach?

PR #6701, opened 2018-10-30
Content - processedReferences field & ContentProcessor #6692

vbradnitski:


GlennRicaud:
Two problems here: 
- The processor goes through all the string properties. The purpose here is to go through all the HtmlArea only. You will need to retrieve the form for this data/siteConfig/x-data and visit the properties for these HtmlAreas
- Do not use processedContents as a field. Declare it inside processCreate and processUpdate

Issue #6711, opened 2018-10-31
Vacuum - Delete unused segments


GlennRicaud:
* Add the possibility to delete segment
* Create new vacuum task, removing segments for deleted repositories

* Use this issue to refactor the Repository<->Segment conversions

PR #6735, opened 2018-11-09
Vacuum - Delete unused segments (#6711)

GlennRicaud:


GlennRicaud:
It was coherent with the rest of the vacuum behaviour. Updated with error handling now as requested.

To note: There is no error message possibility here, so then I had to log the cause of the exception for each one.

Issue #6697, opened 2018-10-29
Update serialized page structure


aro:
Update the serialization of the content page, putting all the components at the same level instead of in a hierarchical way.

Changes in the data structure:
- List of elements as an array, under "components" property at top level
- The component path starts with slash
- Page is the first component in the list, it will have slash as path (path="/")
- The type property will be any of: "page", "part", "layout", "image", "text", "fragment"
- Each component item has: path and type
- Specific properties for each component will be on a set with name = _\<component-type\>_  . See example JSON. (e.g. "part": {... } 
- The config will be in the specific part, and nested under an object with the app name(replacing dots with dashes) and nested an object with the part/layout name. 
- Rename "controller" property to "descriptor"
- Image will have a property id, pointing to the image content-id
- Text will have a property value, with the text string
- Fragment components will have a property id, pointing to the content-id of the content with type portal:fragment
- In the page of a fragment (content-type='portal:fragment'), the component at the root will have path="/"

Example of serialization for storage of a page:

```json
{
  "components": [
    {
      "path": "/",
      "type": "page",
      "page": {
        "descriptor": "com.enonic.app.superhero:mypage",
        "template": "234234234234",
        "config": {
          "com-enonic-app-superhero": {
            "mypage": {
              "page-config": false
            }
          }
        }
      }
    },
    {
      "path": "/main/1/right/0",
      "type": "part",
      "part": {
        "descriptor": "com.enonic.app.superhero:search-form",
        "config": {
          "com-enonic-app-superhero": {
            "search-form": {
              "showPostCount": "yes",
              "showHierarchy": 42
            }
          }
        }
      }
    },
    {
      "path": "/main/1/right/1",
      "type": "part",
      "part": {
        "descriptor": "com.enonic.app.superhero:categories",
        "config": {
          "com-enonic-app-superhero": {
            "categories": {
              "showPostCount": false,
              "showHierarchy": true
            }
          }
        }
      }
    },
    {
      "path": "/main/1/right/2",
      "type": "image",
      "image": {
        "id": "423423423423"
      }
    },
    {
      "path": "/main/1/right/2",
      "type": "text",
      "text": {
        "value": "Hello"
      }
    },
    {
      "path": "/main/1/right/3",
      "type": "fragment",
      "fragment": {
        "id": "2342342342"
      }
    }
  ]
}
```

Example of serialization for storage of a fragment (content-type='portal:fragment'):

```json
{
  "components": [
    {
      "path": "/",
[...]

PR #6741, opened 2018-11-13
Update serialized page structure #6697

ashklianko:


ashklianko:
1. Need to update apps that use content lib to fetch newer version of content lib (I used features app for testing)
2. Need to update ImageComponent on frontend to not use config ( ImageComponent is not supposed to have it , at least now) and send image's Caption as a separate param, same as image content's id is sent

ashklianko:
Thanks Alex! on it;

Issue #6743, opened 2018-11-14
Delete Content dialog - Progress bar is not updated during delete operation


vbradnitski:
@alansemenov commented on [Tue Nov 06 2018](https://github.com/enonic/app-contentstudio/issues/537)

1. Open Content Studio
2. Select some content with lots of children - "Large Tree" from the Features app is good (208 items total)
3. Open Delete Content dialog and press Delete

Progress bar appears and is visible for several seconds but it's not being updated while items are being deleted.

PR #6745, opened 2018-11-15
Delete Content dialog - Progress bar is not updated during delete ope…

vbradnitski:
…ration #6743

### DIFF
--- a/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/content/DeleteContentProgressListener.java
+++ b/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/content/DeleteContentProgressListener.java
@@ -8,6 +8,8 @@ public final class DeleteContentProgressListener
 {
     private final ProgressReporter progressReporter;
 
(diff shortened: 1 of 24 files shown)


GlennRicaud:
``progressReporter.progress( progressCount, progressCount )``
Wow. How has this progress ever worked. Good catch.
I guess there was some UI trickery by keeping the total at some point.
But from a backend point of view this has basically never worked.

Agreed with the solution of setTotal and to have a NodeListener.

But 2 remarks:
1 - You can put size 0 in the DeleteRunnableTask.countChildrenToDelete

But most importantly:
2 - Not a fan of passing the NodeListener to NodeStorage, BranchService, StorageDao.
I think this is really a NodeService (+commands) concept.

2.1 Could you instead do a batching in AbstractDeleteNodeCommand.deleteNodeWithChildren. And that would be the progress reported.
2.2 Also you have the total in the same method by using nodesToBeDeleted.getSize(), so you do not need to do this additional job in DeleteRunnableTask.countContentsToDelete.
This total will be propagated through the listeners to DeleteRunnableTask.

Issue #6651, opened 2018-10-09
Update indexing of page for flattened page


aro:
Update the index configuration in `PageRegionsConfigProcessor` according to the changes in the page components structure.

Remove the double wild cards "**" since it is not necessary anymore with the flat structure.

Set the following index paths

- Image components:
  - `"components.image.id"`

- Fragment components:
  - `"components.fragment.id"`

- Text components:
  - `"components.text.value"`

- Layout components:
  - `"components.layout.descriptor"`
  - `"components.layout.config.<app-key>.*"`

- Part components:
  - `"components.part.descriptor"`
  - `"components.part.config.<app-key>.*"`

- Page:
  - `"components.page.descriptor"`
  - `"components.page.template"`
  - `"components.page.customized"`
  - `"components.page.config.<app-key>.*"`

sigdestad:
This one is particularily interesting, as we had to do some special tricks to index page content in 6.x. Is there any logical improvement we can do to how we store data to optimize indexing/search of pages?

PR #6753, opened 2018-11-15
Update indexing of page for flattened page #6651

ashklianko:


ashklianko:
No problem Alex, my fault, I had a question about that I forgot to ask:
Won't it be a problem for indexing that components are in array? So path like "components.fragment.id" will work?

PR #6759, opened 2018-11-21
Update indexing of page for flattened page #6651

ashklianko:


ashklianko:
One more question: in PageConfigProcessor we used to index page's config only if it is not empty, do we need same behaviour for layout and part(currently indexing their configs anyway)?

PR #6767, opened 2018-11-21
Delete Content dialog - Progress bar is not updated during delete ope…

vbradnitski:
…ration #6743


GlennRicaud:
Better. I will merge the issue.
But there are still some problems.

1) The progress is still faulty in some cases: Delete of published contents for example.

2) DeleteContentListener passed inside a PushContentParams.
There is a mixup here. You should not need to pass a DeleteContentListener to ContentService.publish
Outside of ContentService you should only see that a content has been pushed. You should not have to care about the deletion/push really happening below

3) DeleteRunnableTask: You still have the DeleteRunnableTask.countContentsToDelete and DeleteRunnableTask.countChildrenToDelete

I will create an issue for these problems above.

4) I will also create a separate refactoring task.
We start to have a lot of duplicate methods in NodeService that could be united by having: DeleteNodeParams, PushNodeParams, ...
Now (7.0) is actually the perfect occasion to do that.

PR #6772, opened 2018-11-22
Custom styles in HTML Editor

alansemenov:


GlennRicaud:
Did not test it but the code looks good.
Would have been better to make image style an extension of generic style both in the code and the schema but we can always refactor that later.

Issue #6540, opened 2018-09-04
Pasting full image data:image/jpeg type url into an ImageSelector in Content Studio leads to stackoverflow exception and in one case a server crash


RF0:
A student during the developer 101 course was able to get an exception and crash the server by typing a data:image type url from a google image search to an image on the web. This gives a red "null" message in Content Studio.

This should not work, but it should not be able to crash a server, as a non experienced publisher may try something like this.

1. Do a google search for an image (preferably large / high-res)
2. Right click and select "Copy image address"
3. Paste the resulting "data:image/jpeg;base64,/9j/4" .. string into an ImageSelector in a contenttype

Log from the students pc:
[exception.txt](https://github.com/enonic/xp/files/2348581/exception.txt)

Stackoverflow exception when using a data/image type url from google image search (see below)
[exception2-stackoverflow.txt](https://github.com/enonic/xp/files/2348586/exception2-stackoverflow.txt)

<img width="515" alt="screen shot 2018-09-04 at 14 38 49" src="https://user-images.githubusercontent.com/951293/45031847-53869000-b050-11e8-9671-bc655788ceec.png">
<img width="341" alt="screen shot 2018-09-04 at 14 37 24" src="https://user-images.githubusercontent.com/951293/45031848-53869000-b050-11e8-80f2-0f6e4d24eb39.png">

vbradnitski:
We use 'jparsec' lib version 2.1 to parse queries, which have a problems with long strings. It was fixed in version 3.0, but unfortunatelly it compatable only with java 9+.

https://github.com/jparsec/jparsec/commit/f925994868b14c2598a1f7d423d72d0dee144b81

Xp java version upgrading is in progress now, so the best way is just to wait till it will be finished and upgrade lib version then.

PR #6774, opened 2018-11-23
Pasting full image data:image/jpeg type url into an ImageSelector in …

vbradnitski:
…Content Studio leads to stackoverflow exception and in one case a server crash #6540


GlennRicaud:
Good.

But I would not just catch and return an empty list with a status 200.
There is an Index exception here that could have multiple cause. I would return an exception with a more specific message  "Failed to find contents" and wrapping the original IndexException.

vbradnitski:
changed.

Issue #6744, opened 2018-11-14
UUID Version ID


GlennRicaud:
https://github.com/enonic/enonic/blob/master/xp/7.0/improved-version-storage.md#epic2-uuid-version-id

GlennRicaud:
Things to review:
* DeleteVersionCommand seems to not be used and not do anything a check

PR #6781, opened 2018-11-26
UUID Version ID (#6744)

GlennRicaud:
To make it easier to review, here are the PRs of the two parts:
https://github.com/enonic/xp/pull/6770
https://github.com/enonic/xp/pull/6780


GlennRicaud:
@runarmyklebust: https://github.com/enonic/xp/pull/6781/commits/2f30ae0ead0eff8117d8bcbc6b59b06b49c29023

Can you notify me before you merge so we can warn on slack

Issue #6555, opened 2018-09-07
valueLib.binary() throws Exception


ashklianko:
Trying to invoke valueLib.binary(name, stream) fails with Exception:
`java.lang.ClassCastException: Cannot cast com.google.common.io.Files$FileByteSource to com.google.common.io.ByteSource`

That's weird, especially when FileByteSource extends ByteSource

Code to try:
```
var tempFile = Java.type('java.io.File').createTempFile("perm-report-", ".csv");
var Files = Java.type('com.google.common.io.Files');
var byteSource = Files.asByteSource(tempFile);
var binary = valueLib.binary('report', byteSource);
```

PR #6782, opened 2018-11-27
valueLib.binary() throws Exception #6555

vbradnitski:


GlennRicaud:
Not sure about this one.
You overwrite loadClass and skip the locking mechanism.

A solution might be to use the bundle class loader in ApplicationBuilder instead of creating our own here.
``this.bundle.adapt( BundleWiring.class ).getClassLoader()``

Issue #6785, opened 2018-11-28
Implement new node event type for update permissions action


vbradnitski:
Create 'permissionsUpdated' node event type and fire it instead of basic 'update' event in NodeService.applyPermissions method.

PR #6789, opened 2018-11-29
Implement new node event type - permissionsUpdated #6785

vbradnitski:


GlennRicaud:
Well it looks cleaner now but it changes the behaviour of the ApplyPermissions.

Before
Applying permissions would do an update of the "parent" node (not on the children).
-> This would change the modifier and the modifiedTime of this "parent" node

With these new changes
The modifier and modifiedTime are never changed (there is no update, only a modification of permissions).
This does not shock me, but there was maybe a reason to do this update

@runarmyklebust or @sigdestad ?

GlennRicaud:
Confirmation from Thomas that the content should not be "updated" if it is only a permissions changed.

Approved

Issue #6763, opened 2018-11-21
API - Progress - TaskResource


GlennRicaud:
* Create a new JaxRsComponent "TaskResource" with @path("/api/task")
* This resource will have an endpoint "info" taking a task ID as parameter and returning the task information

GlennRicaud:
See response in https://github.com/enonic/xp/issues/6762#issuecomment-440970080

PR #6800, opened 2018-12-05
API - Progress - TaskResource #6763

vbradnitski:


GlennRicaud:
Can you throw an exception if the task is not found: 
``throw JaxRsExceptions.notFound( String.format( "Task [%s] was not found", idParam ) );``

Issue #6803, opened 2018-12-07
Application Service - Check application system requirement before starting


GlennRicaud:
Add a check in application service: The system should not start an application with an invalid system requirement (required XP version).

PR #6808, opened 2018-12-10
Application Service - Check application system requirement before sta…

vbradnitski:
…rting #6803

### DIFF
--- a/modules/core/core-api/src/main/java/com/enonic/xp/app/Application.java
+++ b/modules/core/core-api/src/main/java/com/enonic/xp/app/Application.java
@@ -26,6 +26,8 @@ public interface Application
 
     String getMinSystemVersion();
 
+    boolean includesSystemVersion( final Version version );
+
     String getUrl();
 
     String getVendorName();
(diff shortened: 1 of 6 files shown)


GlennRicaud:
I like that null or empty will return true. It will avoid problems with applications that do not use the latest gradle plugin.

But I think the following like could cause problems.
``this.systemVersion.getLeft().equals( version )``

Left cannot be null. So I would simply let ``systemVersion.includes( version )`` if systemVersion is not null or empty

vbradnitski:
It based on behavior of VersionRange class. Parsing for exact version string like '5.0' won't throw an error, but will create an object with LEFT=5.0, RIGHT=null. Then there is strange things start:
1. isExact() returns false.
2. includes(version) returns true for any version bigger than LEFT.

I was not sure who controls "X-System-Version" header writing for bundle, maybe user can't access it and it's an extra check. @GlennRicaud ?

vbradnitski:
changed.

GlennRicaud:
Merge in https://github.com/enonic/xp/issues/6938

Issue #6804, opened 2018-12-07
API - Status Resource


GlennRicaud:
Create a new JaxRsComponent "StatusResource" with @path("/api/status")
This resource will have an endpoint "server" returning the server information (Same information as what ServerReporter is returning)

sigdestad:
I think this might be a misunderstanding?
Plan was for status to run on a separate port, @runarmyklebust ?

GlennRicaud:
This is not about the separate port at all.
This is about the CLI getting the server version while allowing to extend in the future.

We can change the name "status" for something else of course

PR #6810, opened 2018-12-11
Status Resource #6804

vbradnitski:


GlennRicaud:
As discussed, can you modify to let JAXRS handle the exception since we are in an /api context here.

vbradnitski:
changed.

Issue #6653, opened 2018-10-09
Flattened page structure - Dump migration


aro:
Create an upgrade script to upgrade content nodes created with 6.x that used the tree structure for the page. They should be converted to the new flattened page structure.

TODO: Describe how to do this @GlennRicaud

GlennRicaud:
Explanations in: https://github.com/enonic/xp/issues/6697
Code present both in epic-6649 and issue-6697

GlennRicaud:
For the following page
![screen shot 2018-11-28 at 10 10 48](https://user-images.githubusercontent.com/12030113/49143815-24548780-f2fc-11e8-82aa-a84d7eb030ea.png)

Here is the Node data part of the VersionNode
[nodedata-aftermod.json.txt](https://github.com/enonic/xp/files/2623837/nodedata-aftermod.json.txt)
[nodedata-beforemod.json.txt](https://github.com/enonic/xp/files/2623838/nodedata-beforemod.json.txt)

GlennRicaud:
@aro:
Some question regarding the new serialization:

1) The property "config" of Image components has been dropped. It has been replaced by a property "caption". Is that on purpose?
2) The property "config" of Image components has been dropped. Is that on purpose?

GlennRicaud:
If all of this is on purpose, then I will start the dump migration tomorrow.
But we will not be able to merge all of this before https://github.com/enonic/xp/pull/6781 is merged, since dump migration is sequential.

![screen shot 2018-11-28 at 16 28 13](https://user-images.githubusercontent.com/12030113/49162252-c2f7dd00-f32a-11e8-8bdc-af67894a2c88.png)
![screen shot 2018-11-28 at 16 28 19](https://user-images.githubusercontent.com/12030113/49162249-c2f7dd00-f32a-11e8-9723-f3edbcaf932c.png)

ashklianko:
> @aro, @ashklianko :
> Some question regarding the new serialization:
> 
> 1. The property "config" of Image components has been dropped. Is that on purpose?
>    It has been replaced by a property "caption" with as value the property "caption" of the Image component config.
> 2. The property "config" of Fragment components has been dropped. Is that on purpose?

Hi Glenn, I removed it on a purpose in spite of it is not mentioned in a task; Config is not needed neither for fragment or image, it is used with components that have a descriptor (layout and part) and a page itself;

sigdestad:
What happens if the fragment is a part or layout?

aro:
[...]

PR #6815, opened 2018-12-14
 Flattened page structure - Dump migration (#6653)

GlennRicaud:


GlennRicaud:
@ashklianko: 
Most files changed here are test resources that had to be upgraded to be relevant
Most interesting class to check here is: FlattenedPageDataUpgrader.

ashklianko:
Seems OK, but I'll add some tests first for FlattenedPageDataUpgrader and create PR for those.

PR #6818, opened 2018-12-17
Admin rest - Add a preview endpoint for media contents

vbradnitski:


GlennRicaud:
Another solution would be to use the existing endpoint and add a "download" parameter (default to true I guess) and remove the test with "application/msword" and "application/vnd." ( I know it comes from the original application).
That would match the existing behaviour with portal: http://repo.enonic.com/public/com/enonic/xp/docs/6.15.1/docs-6.15.1-libdoc.zip!/module-portal.html#.attachmentUrl
Because "preview" does not make sense here I think.

I approved. I let you decide which solution you choose.

vbradnitski:
refactored.

Issue #6764, opened 2018-11-21
API - Progress - Operations as tasks


GlennRicaud:
Adapt the following endpoints:
``/api/system/dump``
``/api/system/load``
``/api/system/vacuum``
``/api/system/upgrade``
``/api/repo/export``
``/api/repo/import``

* Execute them inside tasks and return the task ID.
* Use the listener mechanisms available to report the progress

PR #6819, opened 2018-12-17
API - Progress - Operations as tasks #6764

vbradnitski:


GlennRicaud:
Will wait for the CLI modifications to test and merge

GlennRicaud:
We should log exception thrown. Will create a separate issue to solve this

Issue #6870, opened 2019-01-18
Add connection size config for port connectors


vbradnitski:
Add jetty config for connectors:

xp : 8080
management : 4848
monitor : 2609

vbradnitski:
Add jetty config for connectors:
1. xp : 8080
2. management : 4848
3. monitor : 2609

GlennRicaud:
See comment in https://github.com/enonic/xp/pull/6874

PR #6874, opened 2019-01-18
Add connection size config for port connectors #6870

vbradnitski:


GlennRicaud:
This is not really what Thomas was thinking about I think
The problem is more the thread pool of the server and not the connector acceptors I think.
This thread pool is shared by the entire server so no real possibility to specify a number of threads per connector (to be confirmed).

What we could do is specify the max number of threads for the entire server.

I will close this issue

PR #6922, opened 2019-02-06
ContentImageResource: add "crop" parameter, remove "scaleWidth"

alansemenov:


sigdestad:
I'm unsure about this one? Thought source refs binary label?

alansemenov:
I don't understand the question. "source" is still there, it was wrong message in the commit.

Issue #6918, opened 2019-02-06
Do not resolve services and assets under site


GlennRicaud:
For backward compatibility services and assets present under /site/services and /site/assets are still resolved.

Only services and assets present under /services and /assets should be resolved

PR #6932, opened 2019-02-13
Do not resolve services and assets under site #6918

ashklianko:


ashklianko:
There's also "site/services" path used in ServiceDescriptorServiceImpl and ServiceDescriptor, is it also to be updated @GlennRicaud ?

sigdestad:
Yes, services must be placed in /resources - should not resolve any of this under /site anymore

ashklianko:
Ok!

ashklianko:
Updated

PR #6933, opened 2019-02-13
HTML Area Styles - Process HTML

alansemenov:


jsi:
What happened to Codacy?  Did we remove it from this project?  
The analysis came quickly for this, and it is positive (neutral).  The code quallity did not change at all with this commit: https://app.codacy.com/app/enonic/xp/commit?bid=4267815&cid=323003429

GlennRicaud:
Some problems with this commit: https://github.com/enonic/xp/pull/6933/commits/0b0b2655d900e9c74a57e9d5e3964cbc951a3807
The parameter "scaleWidth" was put back (which is good, there is no need to remove a parameter. You could even leave the default value it had before and just adapt its use in content-studio).

But in any case, you do not seem to use it and hardcode 'true' as the value. Please pass the parameter value to the appropriate private functions

Issue #6916, opened 2019-02-06
Serve webapps on /webapp


GlennRicaud:
In a previous issue, '/portal' and '/admin/portal' haven been renamed to '/site' and '/admin/site' respectively.
The goal of this issue is to rename '/app' to '/webapp'

sigdestad:
One omportant note here. When traffic hits /webapp,  it should automatically redirect to /webapp/ before handling control over to webapp.js

When using vhosts, this is not a problem. 

Verify solution with Alan

GlennRicaud:
This redirection improvement does not seem consistent with the behaviour of the rest of the platform.

What is the purpose of this modification: PWA scopes?
In any case, it does not seem linked to this issue. Let's discuss it separately before we take any decision.

sigdestad:
Agree it is a separate task. But it is required for PWA to work smoothly without hacking inside apps. We should really do the same for site roots. Ask Alan - I thought this was already implemented, but wanted to be absolutely sure.

GlennRicaud:
What has been implemented in a previous version is that "/myapp/" is now handled the same way as "/myapp" (It was not handled at all before).
But no there is no redirection no.

And not sure that it is a good idea to change the current behaviour of the entire platform just because of this particular PWA scope issue that can be solved by a simple redirection "/app" -> "/app/"

sigdestad:
It is not the behaviour of the entire platform.. it is just how the webapp endpoint works - and it should redirect when not using vhosts. Our current behaviour is not good in general.

PR #6946, opened 2019-02-19
Serve webapps on /webapp #6916

vbradnitski:


GlennRicaud:
Please revert the changes in the class InstallApplicationCommand.java.
It is communicatiing with com.enonic.xp.impl.server.rest.ApplicationResource which has not changed (and that is correct).

vbradnitski:
fixed.

Issue #6953, opened 2019-02-20
Component config - Group by descriptor key - Serialization


GlennRicaud:
Adapt the serialization/deserialization to changes described in https://github.com/enonic/xp/issues/6952

ashklianko:
Ok!

PR #6964, opened 2019-02-21
Component config - Group by descriptor key - Serialization #6953

ashklianko:
-Now storing component configs by descriptor key (applicationKey + descriptor name)


ashklianko:
Few moments, forgot to add test.

PR #6988, opened 2019-03-07
Check for empty config in PageDataSerializer

ashklianko:
-Check if config is non empty before serializing config


GlennRicaud:
I am trying to understand how can this have any impact outside of the serialization.
To decide to serialize an empty config as an empty property set or not serialize it should not have any impact.

But I am not against fixing the config serialization. There are clearly problems here
- For example before issue#6697, DescriptorBasedComponent::hasConfig was always returning true. So clearly better.
But now we have cases like ComponentMapper:71 that gets the config and check if null (and this will always be true...) instead of reusing hasConfig.
- Page does not necessarily always have a config not null unlike DescriptorBasedComponent.
- This check added inside PageDataSerializer would be better inside Page::hasConfig, to be coherent with DescriptorBasedComponent::hasConfig).

So totally for a refactoring:
- "config" set in the constructor of Page the same way as DescriptorBasedComponent
- Use hasConfig instead of (getConfig != null) when using DescriptorBasedComponent (The way it is done for Page)
- Move this new check inside Page::hasConfig

Issue #6989, opened 2019-03-08
Disable cluster provider during restoration


GlennRicaud:
* Close all indexes before starting the restoration
* Disable Cluster providers if closed indexes
* Wait for closed indices before starting restoration

GlennRicaud:
A design flaw more visible now is that there is no check that all the services used by libraries are activated before the applications are started. For the core libraries, these are the services used:

AdminToolDescriptorService
ContentService
ContentTypeService
EventPublisher
HtmlSanitizer
IndexService
LocaleService
MailService
MixinService
MultipartService
NodeService
PortalUrlService
RepositoryService
ResourceService
SecurityService
ScriptEventManager
TaskDescriptorService
TaskService
ViewFunctionService
WebSocketManager
XDataService

PR #6992, opened 2019-03-11
 Disable cluster provider during restoration (#6989)

GlennRicaud:


modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/elasticsearch/snapshot/SnapshotRestoreExecutor.java
```diff
@@ -61,6 +63,57 @@ private void addResult( final RestoreResult.Builder builder, RestoreResult resul
...
+    {
+        final RestoreResult.Builder builder = RestoreResult.create();
+        final RepositoryIds existingRepositoryIds = getRepositories( true );
+        RepositoryIds currentRepositoryIds = null;
+
+        try
+        {
+            //Closes  current indices and wait for ES provider to be disabled
+            closeIndices( existingRepositoryIds );
+            while ( !clusterManager.isHealthy() )
```
runarmyklebust: while !healthy -> wait?

modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/elasticsearch/snapshot/SnapshotRestoreExecutor.java
```diff
@@ -61,6 +63,57 @@ private void addResult( final RestoreResult.Builder builder, RestoreResult resul
...
+            {
+                Thread.sleep( 1000l );
```
runarmyklebust: Would be nice to log here maybe?

modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/elasticsearch/snapshot/SnapshotRestoreExecutor.java
GlennRicaud: You are only reviewing the first commit

Issue #7085, opened 2019-04-16
Default values not set on sub components


vbradnitski:
@alansemenov commented on [Thu Apr 11 2019](https://github.com/enonic/app-contentstudio/issues/424)

1. Open a site / Page Editor
2. Open the Page Components view
3. Add a part inside the Page Components view (via context menu)
4. Select descriptor for this part (inside the Inspect panel)

The page auto-saves, the part remains selected on the page and inside the Inspect panel but it's no longer selected inside the Page Components view

GlennRicaud:
I do not reproduce the issue.
Was it rather a content studio bug that has been fixed and this is just a refactoring/fixing something else?

vbradnitski:
Problem is reproduced only with parts contained inputs with default values inside layouts. Also reproduced it for default values of layouts, pr is updated.

PR #7086, opened 2019-04-16
Page component loses selection in the Page Components view after desc…

vbradnitski:
…riptor is selected #7085


GlennRicaud:
Good work on 2nd commit.
First commit would have overwritten values I suppose

Issue #7097, opened 2019-04-30
Handle spaces in server script parameters


pmi:
@sigdestad commented on [Thu Apr 25 2019](https://github.com/enonic/cli-enonic/issues/127)

At least on macs, spaces are not working properly for sandboxes.
Apparently, sandbox is created fine, but XP_HOME is set incorrectly?

See screenshot below:
![Screenshot 2019-04-25 at 12 44 48](https://user-images.githubusercontent.com/482228/56730373-265e2d80-6758-11e9-95a8-ae92ad7ec8eb.png)

Fix and verify on both Mac and Windows..

PR #7098, opened 2019-04-30
Spaces not supported for sandbox on mac #7097

pmi:


GlennRicaud:
Not a valid solution. This will block us from calling the script with multiple parameters.
I guess the CLI is trying to call the server with something like this ``server.sh -Dxp.home="/myhome withaspace"``
Try to find a fix for the server scripts that would allow us to do both

PR #7119, opened 2019-05-22
Check if node exists when checking if content has unpublished children

edloidas:
* Updated `HasUnpublishedChildrenCommand` to return `false`, if the parent node is absent.
* Updated unpublish command message when content is deleted.


GlennRicaud:
Mixing different things in this PR.

The original error is that calls to /admin/rest/content/hasUnpublishedChildren with a node ID for an unexisting node will fail.
It should return false to be coherent with the rest.
That is fixed in commit 0adca64a49b527dca52956ea4aa0a12873e41dc9

I tried the entire branch, but then get strange behaviour with "Processing failed: Item "..." could not be published".

I will merge commit 0adca64a49b527dca52956ea4aa0a12873e41dc9 and set this issue to done.
We can always open another issue if we need to discuss and modify the way the unpublish works

Issue #7143, opened 2019-06-06
Skip creating extra cache files for gif images


vbradnitski:
New cache image is created for any new combination of read params, but it contains exact original source, so it can be skipped.

vbradnitski:
New cache image is created for any new combination of read params, but it contains exact original source, so it can be skipped.

GlennRicaud:
Closing issue
See comment in https://github.com/enonic/xp/pull/7145#issuecomment-501616099

PR #7145, opened 2019-06-06
Skip creating extra cache files for gif images #7143

vbradnitski:


GlennRicaud:
Multiple problems here. Reformat of images, remove possibility to process gifs, mismatch with the check at other places, ... 
Closing this issue.

If the goal is to always serve Gifs (without reformat) as attachment, then it should be discussed and really refactored

Issue #7001, opened 2019-03-11
Commit Resource


GlennRicaud:
Create in admin-impl com.enonic.xp.admin.impl.rest.resource.commit.CommitResource
  - It will have access to the service NodeService and CommitService
  - Create an endpoint/method "getCommit" taking GetCommitParameters as input

GetCommitParameters will have the following parameters:
  - commitId: NodeCommitId
  - nodeId: NodeId

CommitResource::getCommit will work as following
  - If no commitId, retrieve it based on the nodeId (using NodeService::getActiveVersions)
  - Retrieve and return the commit (CommitService::get)

GlennRicaud:
https://github.com/enonic/xp/pull/7147#issuecomment-509101334

PR #7147, opened 2019-06-11
Commit Resource #7001

vbradnitski:


GlennRicaud:
Different solution picked: https://github.com/enonic/xp/issues/7198

Issue #7237, opened 2019-07-15
IssueResource - Support statistics by issue type


edloidas:
This task depends on [this backend task](https://github.com/enonic/xp/issues/7183) to be completed.

* Implement support of the issue `type` in `find` request.
* Update the issues `stats` request to return statistics by issue type.

PR #7238, opened 2019-07-15
Add support of the issue type to the statistics request in IssueResou…

edloidas:
…rce #7237


GlennRicaud:
Only one remark.
In IssueResource, can you make the POST getStats a new method and deprecate the old one (to avoid breaking change for Content Studio)

Issue #7244, opened 2019-07-18
Fragment Resource - Add workflow info parameter


ashklianko:
@sgauruseu commented on [Mon Jul 15 2019](https://github.com/enonic/app-contentstudio/issues/642)

1. Open new site wizard, select Features, select Page-controller.
2. Insert new text component(for example) then select `Save as Fragment` menu item:

![image](https://user-images.githubusercontent.com/3728712/61215572-fe0d1c00-a713-11e9-9fb3-b78a95cfeba4.png)

3. New fragment wizard is opened.

BUG: Workflow state of the fragment should `In Progress`, because parent site is 'In Progress'

![image](https://user-images.githubusercontent.com/3728712/61215739-59d7a500-a714-11e9-99b8-0bb224570165.png)

GlennRicaud:
Why not in progress anymore.
What is the status?

ashklianko:
Had to switch to Content Layers epic, so postponed for now

PR #7245, opened 2019-07-18
New fragment should be created in the same workflow state as the cont…

ashklianko:
…ent it was created from #7244

-Creating fragment with 'In Progress' state because it's parent will have same state after creating fragment


GlennRicaud:
Not correct.
The goal is that the "New fragment should be created in the same workflow state as the content it was created from"

Issue #7255, opened 2019-07-24
Content lib - Modify-method changes values to strings in the site config


arneroen:
Hi,

I think I've found a bug where contentLib.modify will change booleans in the siteConfig to strings (eg true -> "true").
(Enonic XP version 7.0.1)
To reproduce it:

```
<?xml version="1.0" encoding="UTF-8"?>
<site>
  <form>
    <input type="CheckBox" name="someConfigValue">
      <label>label</label>
      <occurrences minimum="0" maximum="1"/>
      <default>checked</default>
    </input>
  </form>
</site>

```

And the function that modifies it is just
```
  contentLib.modify({
    key: '/path',
    branch: 'draft',
    requireValid: false, //to see the changes in the data structures. 
    editor: site => {
      return site
    }
  })
```

Using data toolbox we can inspect the config values before and after
Before:
```
"data": {
  "siteConfig": {
    "applicationKey": "com.example",
    "config": {
      "someConfigValue": true
    }
  }
}
```
After: 
```
"data": {
  "siteConfig": {
    "applicationKey": "com.example",
    "config": {
      "someConfigValue": "true"
    }
  }
}
```

GlennRicaud:
@arneroen Thank you for reporting this bug. We will fix it in 7.0.3

GlennRicaud:
@vbradnitski

The current fix is good and should be included in the final version.

But it does not solve the wider problem:
SiteConfig forms are not used when modifying contents...
One solution would be to inline the site forms similar to mixins when calling FormJsonToPropertyTreeTranslator

To test the solution try the same thing but with the following site config

```
<?xml version="1.0" encoding="UTF-8"?>
<site>
  <form>
    <input name="myDate" type="Date">
      <label>Date</label>
      <occurrences minimum="0" maximum="0"/>
    </input>
  </form>
</site>
```

PR #7260, opened 2019-07-29
content lib - Modify-method changes booleans to strings in the siteCo…

vbradnitski:
…nfig #7255

### DIFF
--- a/modules/core/core-api/src/main/java/com/enonic/xp/data/Value.java
+++ b/modules/core/core-api/src/main/java/com/enonic/xp/data/Value.java
@@ -73,6 +73,11 @@ public abstract class Value
         return type == ValueTypes.STRING || type == ValueTypes.XML;
     }
 
+    public boolean isBoolean()
+    {
+        return type == ValueTypes.BOOLEAN;
+    }
+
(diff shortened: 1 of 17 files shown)


GlennRicaud:
This is not really the problem.
The problem is that the site config is not used to translate the site config (has it ever worked? :S).
Closing this PR and will make a new one with proposition of fix.

GlennRicaud:
Getting NPEs while testing. Please be careful with this issue. Especially since you have created a PR to 7.0 branch. 
I see at least one problem with the check at BaseCreateOrModifyContextHandler::169
Also createJson is called twice in BaseCreateOrModifyContextHandler:createPropertyTree


```
Caused by: java.lang.NullPointerException: null
	at com.enonic.xp.lib.content.BaseCreateOrModifyContextHandler.createPropertyTree(BaseCreateOrModifyContextHandler.java:156)
	at com.enonic.xp.lib.content.ModifyContentHandler.updateContent(ModifyContentHandler.java:94)
	at com.enonic.xp.lib.content.ModifyContentHandler.lambda$newContentEditor$0(ModifyContentHandler.java:72)
```

Issue #7286, opened 2019-08-08
Set default layer language for a new content


ashklianko:
@alansemenov commented on [Mon Jul 22 2019](https://github.com/enonic/app-contentstudio/issues/673)

When a new content is created, set `language` property of the new content to default language of the current layer (now it's set to `null`).

GlennRicaud:
Modification done in PR #7287 is not correct.

This new behaviour is overriding the param (not behaving as a default value).
Also the place to do this kind of thing already exists: Please adapt CreateContentCommand::getDefaultLanguage instead

Moving this issue back to InProgress

GlennRicaud:
Pass back to done. Will be fixed in https://github.com/enonic/xp/issues/7305

PR #7287, opened 2019-08-08
Set default layer language for a new content #7286

ashklianko:


alansemenov:
@GlennRicaud I'll merge this since we need it for the feature to function but will leave the PR open, so that we could fix things If anything is not right.

Issue #7291, opened 2019-08-12
Implement API to fetch layer info for provided contentId


alansemenov:
We need an API that will return the array of layers (and content properties in each of these layers) which have an item with specified contentId and property `inherited` set to `false`. The list of layers should be correctly sorted from top to bottom.

Each array item should have the following properties:
1) layer parent id
2) layer id
3) layer icon
4) displayName (in the layer)
5) name/path (in the layer)

In the UI this API will be used to render the chain of inheritance for the specific item in the specific layer:

![image](https://user-images.githubusercontent.com/11612490/62877698-eb443200-bd27-11e9-9abf-6d43a57d775f.png)

PR #7297, opened 2019-08-14
Implement API to fetch layer info for provided contentId #7291

vbradnitski:


sigdestad:
FYI: Here is a UI that we will implement, and we might also show "child" layer versions of the same item.
Screenshot 2019-08-14 at 16 32 49

So, basically, getting access to all layers where this version exists, and if it is a "local copy" or just inherited would be what we need to begin with. We can then access the content details in separate requests i assume?

sigdestad:
Do you have an example of api signature and response?

alansemenov:
@vbradnitski Pls add content language to response. We also need compare/publish status but I guess that could be fetched in a separate request from the client.

GlennRicaud:
This was discussed and it was supposed to be done in the widget, when rendering server side.
Not as an additional admin API endpoint in the runtime.

I have been told we are short on time and we should merge.
Will merge since it is a prototype, but we should do it differently in the final solution.

Issue #7248, opened 2019-07-22
Audit Log - Initializer


GlennRicaud:
* To be done in branch epic-auditlog
* Create an initializer com.enonic.xp.core.impl.auditlog.AuditLogRepoInitializer extends ExternalInitializer
  * It will create a repo "system.auditlog"
  * Root permissions are:
    * Full access for "system.admin"
* Create com.enonic.xp.auditlog.AuditLogService (core-api)
* Create com.enonic.xp.core.impl.auditlog.AuditLogServiceImpl (core-auditlog)
  * Initialize on active method (similar to IssueServiceImpl, ContentServiceImpl, ...)

PR #7298, opened 2019-08-15
Audit Log - Initializer (#7248)

gbbirkisson:
Note: Tests and Codacy problems fixed in the last pull request: #7295


GlennRicaud:
Good. I change the initialization of the repository to be done no matter if the config is enabled or not

Issue #7249, opened 2019-07-22
Audit Log - AuditLog creation/retrieval


GlennRicaud:
* To be done in branch epic-auditlog
* Create com.enonic.xp.auditlog.AuditLogId (core-api)
  * Similar to ContentId
* Create com.enonic.xp.auditlog.AuditLog (core-api) with the following fields
  * id: AuditLogId //Non-null
  * type: String //Non-null
  * time: Instant //Non-null
  * source: String //Non-null
  * user: PrincipalKey //Non-null
  * message: String //Non-null
  * objectUris: Set<java.net.URI> //Non-null, Immutable
  * data: PropertyTree //Non-null
* Create com.enonic.xp.auditlog.AuditLogs extends AbstractImmutableEntitySet<AuditLog>
* Create com.enonic.xp.auditlog.LogAuditLogParams (core-api) with the following fields
  * type: String //Mandatory
  * time: Instant //Default: Current time
  * source: String //Default: Current bundle name
  * user: PrincipalKey //Default: Current User key
  * message: String //Default: Empty string
  * objectUris: Set<java.net.URI> //Default: Empty immutable set
  * data: PropertyTree //Default: Empty string
* Add following method on AuditLogService
  * log(LogAuditLogParams): AuditLog
  * get(AuditLogId): AuditLog
* Implement methods in AuditLogServiceImpl
  * AuditLogs will be store at the root of the repo "system.auditlog" (only one branch: master)

PR #7299, opened 2019-08-15
 Audit Log - AuditLog creation/retrieval #7249

gbbirkisson:
Review pull request #7298 before this one.

Note: Tests and Codacy problems fixed in the last pull request: #7295

Check out TODOs:
* [AuditLogParams // TODO: Get bundle name](https://github.com/enonic/xp/blob/issue-7249/modules/core/core-api/src/main/java/com/enonic/xp/auditlog/AuditLogParams.java#L52)


GlennRicaud:
You are right for the bundle name, we cannot do it at that level

Issue #7324, opened 2019-08-26
lib-mail fails to send an email with attachment


alansemenov:
Originally reported by @hjelmevold. Verify and fix if reproduced. When/if fixed, let Bjørnar know so that he could release a new version of the form builder app.

Example can be found [here](https://developer.enonic.com/docs/xp/stable/api/lib-mail) (scroll down to _Send an e-mail with attachments_)

> In 7.0 with lib-mail, I can't seem to send e-mail with attachments (but plain text e-mails are sent properly). I get the following stack trace:

```
org.eclipse.jetty.server.HttpChannel - /admin/site/preview/default/draft/superhero/form-test/_/component/main/0/left/0
java.lang.LinkageError: loader constraint violation: when resolving method "javax.mail.internet.MimeBodyPart.setDataHandler(Ljavax/activation/DataHandler;)V" the class loader org.apache.felix.framework.BundleWiringImpl$BundleClassLoader @6309e4b4 (instance of org.apache.felix.framework.BundleWiringImpl$BundleClassLoader, child of 'bootstrap') of the current class, com/enonic/xp/lib/mail/SendMailHandler, and the class loader 'app' (instance of jdk.internal.loader.ClassLoaders$AppClassLoader) for the method's defining class, javax/mail/internet/MimeBodyPart, have different Class objects for the type javax/activation/DataHandler used in the signature
    at com.enonic.xp.lib.mail.SendMailHandler.compose(SendMailHandler.java:172)
    at com.enonic.xp.mail.impl.MailServiceImpl.send(MailServiceImpl.java:48)
    at com.enonic.xp.lib.mail.SendMailHandler.send(SendMailHandler.java:117)
    at jdk.scripting.nashorn.scripts/jdk.nashorn.internal.scripts.Script$Recompilation$3159$2347A$mail.L:1#send(com.enonic.app.formbuilder:/lib/xp/mail.js:66)
```

hjelmevold:
Tested app-formbuilder with Enonic XP 7.1.0-B2
Attachments are now sent properly, like in 6.15.
I can verify that this issue has been fixed, and this issue can be closed.

pvmerlo:
I'm still having errors when sending email with attachments. Plain text works fine.
[...]

PR #7326, opened 2019-08-27
lib-mail fails to send an email with attachment #7324

vbradnitski:


vbradnitski:
'javax.mail:mail' already has 'javax.activation' as a dependency and I can't find any other usage of it but maybe missed something. @GlennRicaud ?

GlennRicaud:
Well it solves it for javax.mail:mail but it is because then it uses javax.activation 1.1.0 from the Framework bunble rather than javax.activation 1.2.0 from com.sun.activation:javax.activation:1.2.0

But there are 5 other bundles that were using this package. Could you check the impact
Screenshot 2019-09-04 at 12 40 35
Screenshot 2019-09-04 at 12 34 25

GlennRicaud:
JAXB is used by Tika.
But luckily JAXB 2.2 is using java.activation 1.1

So it should be alright

Issue #7333, opened 2019-08-28
Handle localization format errors


vbradnitski:
Localization message format allows only number value inside `{}` braces. Adding text value leads to `MessageFormat` parser failing and crash FormJson serialization process without any exceptions.

PR #7336, opened 2019-08-28
FormJson.java serialization fails because of wrong localization messa…

vbradnitski:
…ge value #7333


jsi:
Hi Slava!
Codacy have raised an issue here, that a JUnit test method does not follow the naming standard.  Is this correct?  Then please fix this.  Otherwise, please let me know, and I can adjust the rule.

PR #7341, opened 2019-08-30
ConfigLoader FileReader is not properly closed #7340

### LINKED ISSUE #7340: ConfigLoader - FileReader is not properly closed
#7368 introduces JUnit5 and TmpDir annotation which exposed a bug: `com.enonic.xp.server.internal.config.ConfigLoader ` creates `java.io.FileReader` but never closes it.

Junit5 tests fail on Windows due to this: JUnit5 tries to delete contents of temporary folder but fails due to file still being open.

Fix is easy: wrap FileReader with try-with-resources block.

rymsha:
fix #7340


GlennRicaud:
Nice catch

PR #7363, opened 2019-09-05
Fix jsdoc for "attachmentUrl" in portal lib

tajakobsen:
The key `params.type` was used on both line 79 and 81. It appears that this is a typo, and the correct key on line 79 should be `param.label`.


GlennRicaud:
Merged.
Thank you for the PR :+1:

Issue #7296, opened 2019-08-13
Dump failing on missing blob


runarmyklebust:
If dumping a repository with missing blob, the whole dump will fail with exception.
The dumper should not quit, but rather finish the dump and log / report on failing nodes

```
2019-08-13 12:58:13,743 ERROR c.e.x.j.i.e.JsonExceptionMapper - Cannot get blob with blobKey: 43c7e0a9533b1e10ce87fead5c0c9260d6b68405: blob is null
java.lang.IllegalArgumentException: Cannot get blob with blobKey: 43c7e0a9533b1e10ce87fead5c0c9260d6b68405: blob is null
	at com.enonic.xp.repo.impl.node.dao.NodeVersionServiceImpl.getFromBlob(NodeVersionServiceImpl.java:84)
	at com.enonic.xp.repo.impl.node.dao.NodeVersionServiceImpl.doGetNodeVersion(NodeVersionServiceImpl.java:72)
	at com.enonic.xp.repo.impl.node.dao.NodeVersionServiceImpl.get(NodeVersionServiceImpl.java:55)
	at com.enonic.xp.repo.impl.storage.NodeStorageServiceImpl.get(NodeStorageServiceImpl.java:267)
	at com.enonic.xp.repo.impl.node.NodeServiceImpl.getByNodeVersion(NodeServiceImpl.java:644)
	at com.enonic.xp.repo.impl.dump.RepoDumper.doStoreVersion(RepoDumper.java:189)
	at com.enonic.xp.repo.impl.dump.RepoDumper.lambda$dumpVersions$2(RepoDumper.java:174)
	at java.util.HashMap$KeySpliterator.forEachRemaining(HashMap.java:1553)
	at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:580)
	at com.enonic.xp.repo.impl.dump.RepoDumper.dumpVersions(RepoDumper.java:166)
	at com.enonic.xp.repo.impl.dump.RepoDumper.lambda$execute$1(RepoDumper.java:94)
	at com.enonic.xp.context.ContextImpl.runWith(ContextImpl.java:85)
	at com.enonic.xp.repo.impl.dump.RepoDumper.execute(RepoDumper.java:94)
	at com.enonic.xp.repo.impl.dump.DumpServiceImpl.dump(DumpServiceImpl.java:107)
	at com.enonic.xp.impl.server.rest.SystemResource.systemDump(SystemResource.java:88)
[...]

PR #7366, opened 2019-09-05
Dump failing on missing blob #7296

anatol-sialitski:
- fixed


GlennRicaud:
I got your first remark now (where to put the errors in the result).
It is true that there is no place for the error messages in RepoDumpResult

* You will have to update RepoDumpResult
  * Add a field: private final List<DumpError> versionsErrors;
* Adapt the serialization

* Adapt the code you wrote to add the exception as an error instead of logging it (as it is done in RepoDumper::doDumpNode)

Issue #7370, opened 2019-09-06
Tests are running on older version on OSGI


rymsha:
core-app tests using Felix 5.2
launcher-impl declares 6.0.1

PR #7371, opened 2019-09-06
Tests are running on older version on OSGI #7370

rymsha:


GlennRicaud:
About ApplicationServiceImplTest.java , it is a bit strange that it takes the felixVersion instead of the real version which is 7.1.0-SNAPSHOT
Could you look into that.

Also if really it is not possible, then can you update the other version ranges (Replace 5.2 by 6.0 and the other random values by appropriate values)

rymsha:
Tests deploy bundles  top level "Bundle" which is Felix itself. So, system version is version of Felix.
There is little need to update that, and in addition we can verify that we didn't use old version of Felix.

Issue #7379, opened 2019-09-06
Minimize use of LinkedList.


rymsha:
In theory it is slower than, i.e. ArrayList
https://stackoverflow.com/questions/322715/when-to-use-linkedlist-over-arraylist-in-java

There are also places where it is better to be replaced with ImmutableSet.Builder or similar.

PR #7380, opened 2019-09-06
Minimize use of LinkedList. #7379

rymsha:


GlennRicaud:
Can you revert the change of return type of VirtualFilePath:getElements
Will merge after that.

rymsha:
I guess, this is due to the fact it is part of API...
Well, sure.

Issue #7362, opened 2019-09-05
Repository Data - Data field


GlennRicaud:
To be PRed to branch "next"

* Add field on com.enonic.xp.repository.Repository
  * private final PropertyTree data;
* Adapt serialization/unserialization

sigdestad:
Question:
Why do we need to serialize the propertyTree? Aren't we storing this data in the system repo?

rymsha:
Yes we will do store `data` in system-repo.

Your initial questions will remain unanswered for now... I need to talk to Glenn.

rymsha:
Questions to Glenn:
- What does "Adapt serialization/unserialization"  mean?
- How do we deal with Dump/Restore. Especially if we plan to save an Attachment it this repo?

GlennRicaud:
By serialization/deserialization, I meant the translation to/from node.
These classes are often suffixed with Serializer in the code: ContentDataSerializer, IssueDataSerializer, ...
But in this case RepositoryNodeTranslator as you found out.
So no problem here.

And yes we will have to change again in a second issue by including BinaryAttachments to CreateRepositoryParams.
Will review and create a second issue.

PR #7394, opened 2019-09-11
Repository Data - Data field #7362

rymsha:


rymsha:
Work in progress still. This PR is not assigned for review.
So, better ask a question in the related issue #7362

PR #7401, opened 2019-09-12
New fragment should be created in the same workflow state as the cont…

ashklianko:
…ent it was created from #7244

-Setting fragment parent's workflow state to a newly created fragment


GlennRicaud:
Could you add the following field in CreateFragmentParams
 private final WorkflowInfo workflowInfo;

Similarly to CreateContentParams

ashklianko:
Ok!

PR #7403, opened 2019-09-12
Code cleanup: simplify ContentPaths code #7402

rymsha:


rymsha:
Don't merge. It breaks API. I will restore it first

Issue #7408, opened 2019-09-13
Repository Data - BinaryAttachments


GlennRicaud:
To be PRed to epic-repositorydata

* Make the create/update methods in RepositoryEntryService take *Params classes as parameter
  * Keep the existing methods for backward compatibility
* For the Create and update parameters, allow so specify an additional field
  * BinaryAttachments attachments (Look at NodeServiceImpl to check how it is done=
* Adapt RepositoryService methods

PR #7419, opened 2019-09-16
Repository Data - BinaryAttachments #7408

rymsha:


rymsha:
I think it must be reverted. Test doesn't pass...

PR #7422, opened 2019-09-16
Minimize use of LinkedList. #7379

rymsha:
Differes from #7380 now does not break API


modules/core/core-api/src/main/java/com/enonic/xp/vfs/VirtualFilePathImpl.java
```diff
@@ -110,7 +111,7 @@ String join()
     @Override
     public String getName()
     {
-        return this.elements.getLast();
+        return this.elements.get(this.elements.size() - 1);
```
rymsha: Need to restore getLast to restore performance

rymsha: well, it is almost same.  because linked list is smart enough to iterate from tail in this case.

PR #7441, opened 2019-09-23
Make binary extractor max length configurable #7294

anatol-sialitski:
Fixed remarks after review


GlennRicaud:
Why the enable. From what I see it has no impact.
I think we can remove it

Issue #7409, opened 2019-09-13
Repository Data - Repo library


GlennRicaud:
* Adapt the repo library
* Add an update function to update the repository data

PR #7453, opened 2019-09-24
Repository Data #7409

rymsha:


rymsha:
There is big inconvenience with AttachedBinary DTO : it used a lot like Map.Entry but has no Map . 

Don't merge, I need to review this moment.

Issue #7414, opened 2019-09-16
update Guava to version 33.4.0


rymsha:
27.1 is the first version declared to support Java 11
https://github.com/google/guava/issues/3272

rymsha:
https://github.com/google/guava/issues/3302

rymsha:
- Guava is part of XP API (sad but true).
- Many libs are compiled  Guava 26 dependency.
- Many libs and apps use Guava classes internally.
- Gradle plugin we provide for making apps/libs doesn't bundle Guava classes into OSGi (com.google.common becomes provided, not privite)
- Guava update is major (so OSGi requires versions 26.x)

This makes guava update a breaking change. Wait for XP 8.

manibee:
Does JDK 11 compatibility issues fixed in the latest 28.2?
Still "unsupported" warnings appear, is it really a issue?

rymsha:
> Does JDK 11 compatibility issues fixed in the latest 28.2?

Guava 28.2 improved support of Java 11 **App Engine** which we don't support anyway.
 
> Still "unsupported" jdeps warnings appear, is it really an issue?

This warning does not go away with the latest version either. Java 11 we currently use still has sun.misc.Unsafe . Also read https://github.com/google/guava/issues/3272#issuecomment-525430084

PR #7461, opened 2019-09-25
Guava updated to 28.1 #7414

rymsha:


rymsha:
this is breaking anyway.

Issue #7384, opened 2019-09-09
Enforced code format


rymsha:
Currently there is no check for 
- correct order of imports
- unused imports
- curly braces
- etc

Tools for rule enforcement (integrated with IDEs) should be used
We need to introduce https://editorconfig.org/ and
https://checkstyle.sourceforge.io/

rymsha:
https://intellij-support.jetbrains.com/hc/en-us/articles/206544839

PR #7482, opened 2019-10-01
Enforced code format #7384

rymsha:


alansemenov:
@GlennRicaud why is this closed? shouldn't we encourage same coding style? plus it's a plus if Codacy will verify commits and warn about unused/illegal imports.

GlennRicaud:
I guess I misunderstood. I thought this was the codestyle that we are already using.
Should we add it in all projects we have?

rymsha:
Keeping things in sync is another story. But code style is something very static. Last time https://github.com/enonic/dev-tools/blob/master/codestyles/idea/Enonic.xml it was updated was Jan 16, 2017
So it should not be that painful. 
But yeah, I would like to change code styles slightly, do to the fact that they are slightly wrong (will be another PR)

rymsha:
Let's try it. It is not hard to revert.

Issue #7420, opened 2019-09-16
Apache Tika Parsers jar hell


rymsha:
It doesn't hit us hard due to OSGi, but certainly brings up a lot of unnecessary dependencies
 
https://issues.apache.org/jira/projects/TIKA/issues/TIKA-2882

It would be nice to find a lightweight alternative so we don't have uncontrolled dependencies on
- CXF
- Joda time
- Jackson
- etc...

rymsha:
Solution would be to only have Tika parsers as an OSGi bunfle, and not as compile dependency.

PR #7510, opened 2019-10-07
Apache Tika Parsers jar hell #7420

rymsha:


GlennRicaud:
Does that mean that it removes the parsing of PDFs?

Issue #7523, opened 2019-10-09
WebDispatcherImpl implementation issues


rymsha:
- [x] WebHandler.MIN_ORDER and WebHandler.MAX_ORDER are not respected

- [x] Every add/remove creates tons of garbage due to CopyOnWriteArrayList sorting

- [x] Needless method synchronization

- [x] Copying of list on every dispatch

PR #7528, opened 2019-10-10
WebDispatcherImpl implementation issues #7523

rymsha:
- keep elements with same order. So we need a list, not a set with comparator (initially I tried SkipListSet and it didn't work well)
- avoid synchronized methods. AtomicReference with CAS works great for occasionally happening add/remove, so CopyOnWrite is not needed.
- avoid copying on every dispatch (this is why ImmutableList instead of copyOnwSrite)
- Don't expose mutable iterator. ImmutableList helps here as well.
- fix comparator so now it correctly sorts big integers like MAX_INT and MIN_INT
- also funny the implementation had race condition: add method for a short while was introducing handler at the end of chain regardless of its real order and this change was visible in dispatch. So Order was not always respected.


matkasatta:
Great

PR #7535, opened 2019-10-11
Update Guava to 28.1 #7414

rymsha:
This PR awaits decision of haw to deal with OSGi bundle we already provide for version 27.1


rymsha:
Waiting for XP 8

Issue #7283, opened 2019-08-05
Content Project - Admin Rest


GlennRicaud:
Epic #7280
To be PullRequested in branch "epic-project"

- For backward compatibility, we have to keep all the /admin/rest/content/* endpoint
  - /admin/rest/content
  - /admin/rest/content/page
  - /admin/rest/content/page/template
  - /admin/rest/content/page/fragment
  - /admin/rest/content/icon
  - /admin/rest/content/media
  - /admin/rest/content/image
- And make them listen additionaly to respectively:
  - /admin/rest/cms/[project]/[layer]
  - /admin/rest/cms/[project]/[layer]/page
  - /admin/rest/cms/[project]/[layer]/page/template
  - /admin/rest/cms/[project]/[layer]/page/fragment
  - /admin/rest/cms/[project]/[layer]/icon
  - /admin/rest/cms/[project]/[layer]/media
  - /admin/rest/cms/[project]/[layer]/image
- We will create a JAX-RS filter (Look at CmsResourceFilter below)
  - It will listen to /admin/rest/cms/[project]/[layer] and set the context to use the corresponding repository (ignore layer for now)
- Add new Admin Rest Resource to on "/admin/rest/project" to allow to "list","get","create","modify","delete" projects

You can look at the branch epic-contentlayer where a similar work has been done:
* https://github.com/enonic/xp/blob/epic-contentlayer/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/content/ContentResource.java
* https://github.com/enonic/xp/blob/epic-contentlayer/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/layer/ContentLayerResource.java
* https://github.com/enonic/xp/blob/epic-contentlayer/modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/content/CmsResourceFilter.java

PR #7549, opened 2019-10-15
Content Project - Admin Rest #7283

vbradnitski:


modules/core/core-api/src/main/java/com/enonic/xp/project/ProjectService.java
```diff
@@ -13,5 +13,5 @@
 
     Project get( final ProjectName projectName );
 
-    void delete( final ProjectName projectName );
+    ProjectName delete( final ProjectName projectName );
```
rymsha: Usually, true/false is what API should return for delete.

modules/core/core-api/src/main/java/com/enonic/xp/project/Projects.java
```diff
@@ -65,6 +65,11 @@ public Builder addAll( Collection<Project> projects )
             return this;
         }
 
+        public Builder addAll( Project... projects )
```
rymsha: Don't introduce varargs. They are error prone (worst, they accept null)
One can use much safer addAll(List.of(...))

modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/content/CmsResourceDynamicFeature.java
```diff
@@ -0,0 +1,35 @@
...
+import com.enonic.xp.admin.impl.rest.resource.schema.xdata.XDataResource;
+import com.enonic.xp.jaxrs.JaxRsComponent;
+
+
+@Provider
+public final class CmsResourceDynamicFeature
+    implements DynamicFeature, JaxRsComponent
+{
+    private final List<Class<? extends JaxRsComponent>> SUPPORTED_RESOURCES =
+        List.of( ContentResource.class, XDataResource.class, PageTemplateResource.class, ContentImageResource.class,
```
rymsha: Looks strange...
Why capital case, but not static?
But I agree it should not be static :)
It is also Set, not list, because there are no duplicates

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/project/ProjectServiceImpl.java
```diff
@@ -127,18 +128,22 @@ private Project doGet( final ProjectName projectName )
...
+        return callWithContext( () -> {
+            final Boolean result = doDelete( projectName );
             LOG.info( "Project deleted: " + projectName );
+
+            return result;
         } );
     }
 
-    private void doDelete( final ProjectName projectName )
+    private Boolean doDelete( final ProjectName projectName )
```
rymsha: use boolean instead of Boolean

rymsha: boolean instead of Boolean

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/project/ProjectServiceImpl.java
```diff
@@ -127,18 +128,22 @@ private Project doGet( final ProjectName projectName )
     }
 
     @Override
-    public void delete( ProjectName projectName )
+    public Boolean delete( ProjectName projectName )
```
rymsha: boolean instead of Boolean

Issue #7556, opened 2019-10-18
ErrorProne low hanging issues


rymsha:
https://errorprone.info can do static analysis.
Fix ones which are easy

PR #7557, opened 2019-10-18
ErrorProne low hanging issues #7556

rymsha:


rymsha:
next time when Override annotations fix applied globally.

Issue #7574, opened 2019-10-25
Add getByIdAndVersionId() to ContentResource


pmi:
Use ContentService.getByIdAndVersionId()

PR #7581, opened 2019-10-29
Add getByIdAndVersionId() to ContentResource #7574

pmi:


sigdestad:
This method name is pretty obscure. Why not simply call it getVersion(nodeId, versionId)?

GlennRicaud:
Already discussed it. It is not named getByIdAndVersionId in the end

PR #7587, opened 2019-10-30
Code Cleanup: guava Sets replace with Java #7583

rymsha:


GlennRicaud:
Approved. I let you fix the conflict

PR #7551, opened 2019-10-16
Distributed Session Management #7529 #7594

rymsha:
- ignite was updated to latest version. To make sure I don't use deprecated API
- worked around ignite bug where it uses wrong classloader for serialisation of ignite internal classes (this is due to conflict between ignite and OSGi which classloader to use)
- removed ignite XP "Cluster" abstraction, because not all nodes will be ignite cluster (if any)
- javax cache updated to version 1.1.1 (now Apache license). Removed rebundle as it is now correct bundle by itself. It is potentially a breaking change, but very unlikely because code is supposed to import 1.x version of javax.cache package. 
- fixed SessionWrapper: it had race conditions for session creation and was storing reference to potentially expired session
- fixed JettyActivator to wait for all dispatchers to be ready. Before it was potential race.
- introduced IgniteSessionStoreFactory (OSGi service) and made IgniteSessionStore OSGi independent. This make multiple dispatchers work again - before they were sharing SessionStore singleton and fail on startup
- worked around Jetty vs Ignite conflict which deserializer to use
- added package import so Auth info deserialization works (OSGi prevents not imported classes to be deserialized).

PR #7597, opened 2019-11-05
Distributed sessions compare Ignite and Hazelcast

rymsha:
This PR to port all bells and whistles we made for Ignite into Hazelcast ones.


rymsha:
moved to #7551

Issue #7577, opened 2019-10-28
Content lib - getOutboundDependencies


GlennRicaud:
https://discuss.enonic.com/t/proper-way-of-getting-outbound-dependencies/2011/2

Expose getOutboundDependencies in content library

Example of usage:
```
const contentIds = contentLib.getOutboundDependencies({
    key: '/path/to/mycontent'
});
````

alansemenov:
The result should be the same as in the "outbound" part of the Dependencies widget when selecting the content item:

![image](https://user-images.githubusercontent.com/11612490/68215602-d18d5080-ffdf-11e9-89a3-631bbb83df52.png)

PR #7606, opened 2019-11-07
Content lib - getOutboundDependencies #7577

anatol-sialitski:
- implemented


GlennRicaud:
Could you modify it to return the array directly.

```
const contentIds = contentLib.getOutboundDependencies({
    key: '/path/to/mycontent'
});
```

Issue #7674, opened 2019-11-26
Portal Lib - URL generation for content with special characters


GlennRicaud:
Have a content with a norwegian character.
Generate an idprovider URL from this content (test with assetUrl also).
These URL should not contain the content path but they do.

Reproduce and investigate why

Fix in 6.15 also

Example:

1) Content has path "_features_":

IdProviderUrl: `/site/default/draft/_/idprovider/system`

2) Content has path "_featuresåøæ_":

IdProviderUrl: `/site/default/draft/features%C3%A5%C3%B8%C3%A6/_/idprovider/system` (the `features%C3%A5%C3%B8%C3%A6` part should not be in the URL)

alansemenov:
@GlennRicaud What should there be if not path? From what I see it's the same for all urls generated with portalLib

![image](https://user-images.githubusercontent.com/11612490/69709484-c2d91b80-10fd-11ea-9891-527ea9b79b94.png)

GlennRicaud:
Asset URL is wrong there

PR #7684, opened 2019-11-28
Portal Lib - URL generation for content with special characters #7674

vbradnitski:


sigdestad:
Whoo what is this? Need to know what we are changing before any changes are made. URL handling has IMHO been fine for a long time...

vbradnitski:
We always add normalized contentPath to url during construction. But then we need to remove it for some cases. And normalized contentPath was compared with non-normalized contentPath from portal request. And it fails if there are any characters which have to be normalized.

alansemenov:
> Whoo what is this? Need to know what we are changing before any changes are made. URL handling has IMHO been fine for a long time...

@sigdestad Details are in the [original issue](https://github.com/enonic/xp/issues/7674). And here's the [support case](https://enonic.zendesk.com/agent/tickets/3324).

GlennRicaud:
Perfect :+1:

Issue #7673, opened 2019-11-26
Incorrect progress reporting in content reprocess task


pmi:
@pmi commented on [Thu Nov 21 2019](https://github.com/enonic/cli-enonic/issues/182)

reprocess command was not async prior to XP 7.2

PR #7691, opened 2019-12-02
Fix the incorrect progress reporting in content reprocess task #7673

pmi:


GlennRicaud:
The path taken by the API is a content path, not a node path

GlennRicaud:
Discussed with @pmi.
Closing this PR until next commit

pmi:
Updated PR

GlennRicaud:
You have to check that it is not ContentPath.Root
draft:/ and draft:/content will give you the same content path here I think

GlennRicaud:
It still is incorrect if you send draft:/content
Use params.getSourceBranchPath().getContentPath() instead

pmi:
Right ! Updated !
It doesn't look right that both **/** and **/content** return content named **/content** though ;)

PR #7698, opened 2019-12-05
Lokalise: Translations update

jsi:
This pull request was initiated by Lokalise (user Jørgen Sivesind) at 2019-12-05 18:40:04


jsi:
Incorrect use of projects in lokalise

Issue #7608, opened 2019-11-07
Create new version on move/rename


GlennRicaud:
When renaming/moving a node, its children will be renamed/move.
But there is no new version created for the children.

With the modifications made in 7.0, we can now, in a fast way, create a new version (by only creating the version document in ES: the version will have the same blob keys, only the path will change).

PR #7714, opened 2019-12-10
No version is created for children when parent content is renamed #7608

GlennRicaud:


sigdestad:
This change will make moving and renaming content a potential nightmare (have to republish entire structures and causing conflicts with items in progress etc)

GlennRicaud:
But I mean you already have to republish the entire structure. What does it change here?

Issue #7737, opened 2019-12-15
Guava Sets and Lists replace with Java (part 2)


rymsha:
continue on #7583 and #7410

PR #7738, opened 2019-12-15
Guava Sets and Lists replace with Java (part 2) #7737

rymsha:


GlennRicaud:
Approved. I let you fix the conflicts and merge

Issue #7777, opened 2019-12-25
Javax Mail and Activation should be bundles


rymsha:
Currently they are unnecessarily treated as part of system.
Activation jar is not an OSGi bundle, so this is a first step of update supposed to be done in #7619

rymsha:
Licences for Jakarta Activation and Mail are complicated, but OSS compatible. Mentioning them will be done in scope of #7761

PR #7778, opened 2019-12-25
Javax Mail and Activation should be bundles #7777

rymsha:


GlennRicaud:
Approved. I let you solve the conflicts.

Issue #7761, opened 2019-12-20
Add dependency report plugin in gradle build


jsi:
We need some gradle plugin that can give us the complete list of dependencies that should be listed in NOTICE.txt.  -  Manually inspecting the distro does not show everything.

PR #7781, opened 2019-12-27
#7761 - Correct configuration of the gradle license plugin.

jsi:
I tested the two other plugins that produced HTML.  Had problems with both.  The documentation is a lot about Android, and I wonder if they were made to work with Android and that Java is just and afterthought, so either way, I do think this is the best plugin, even if the report generated is not all that easy to work with.


GlennRicaud:
If there had been a PR for https://github.com/enonic/xp/commit/4713643332ccd64af1352ad8172c3ab4d93a10bf, this would have been detected before the push to origin/master.

Approved and merged

PR #7864, opened 2020-02-06
Issue 7863

ashklianko:


GlennRicaud:
Still one question
I like what you did with adding the /content, /media, /schema and /issue in the original commit..
But I see an incoherence in XDataResource.java (/content/schema). Is that on purpose?

ashklianko:
> Still one question
> I like what you did with adding the /content, /media, /schema and /issue in the original commit..
> But I see an incoherence in XDataResource.java (/content/schema). Is that on purpose?

My bad, it requires to be changed also, will do now

ashklianko:
Fixed

Issue #7847, opened 2020-02-03
Remove JSDoc generation on build


rymsha:
Every build zip with JSDoc is generated.
It was used for documentation auto-generation, but starting from XP 7 it is not.

JSDoc generation task requires grunt (obsolete)
and node-gradle-plugin (doe not seem to be supported)

It also prevent an update to Gradle 6.x #7851

rymsha:
No go for now. https://github.com/enonic/xp/pull/7884#issuecomment-588101304

Maybe try this?
https://github.com/node-gradle/gradle-node-plugin

PR #7884, opened 2020-02-19
Remove JSDoc generation on build #7847

rymsha:


GlennRicaud:
I would not delete this yet. A lot of people, including myself, are still using it.

GlennRicaud:
The main reason is that this one is always up to date even during development while the documentation is lagging behind and there has been some process issues with the documentation in the past.

Issue #7845, opened 2020-01-30
Turn on Gradle build scans on Github Actions


rymsha:
It is free
https://guides.gradle.org/creating-build-scans/

It is supported by Github Actions
https://github.com/marketplace/actions/gradle-command

rymsha:
@jsi you need to read and terms of service and see if they are OK for us https://gradle.com/terms-of-service

PR #7908, opened 2020-02-24
Turn on Gradle build scans on Github Actions #7845

rymsha:
https://github.com/enonic/xp/issues/7845#issuecomment-590336697


rymsha:
It is not hard, but overkill. Travis is a "duplicate" build system for XP repository. (Maybe currently Travis is used by more but personally prefer Actions)

Issue #7978, opened 2020-03-27
Application fails to update properly under load


rymsha:
When application gets updated/reinstalled under load it may fail to run properly
```
Caused by: java.lang.NullPointerException: null
at com.enonic.xp.script.impl.service.ServiceRefImpl.findService(ServiceRefImpl.java:33)
at com.enonic.xp.script.impl.service.ServiceRefImpl.get(ServiceRefImpl.java:22)
```

`ScriptExecutorManager` grabs an application bundle which is invalid (~~for unknown reason~~) and uses it to get BundelContext (which is null for not started/invalid bundles)

It is easy to fix to validate BundleContext presence, but it actually doesn't solve the problem. Application must wait for the bundle to fully start.

rymsha:
First part:

 - Reinstallation (or update) of an application calls OSGi Bundle uninstall, which first tries to stop the bundle and send a STOPPING event.

 - STOPPING event triggers a remove of the application from app-registry

 - Millisecond after that request comes from frontend and recreates the app in app-registry (unfortunately with stop-in-progress bundle) creates javascript engine executor and awaits for configuration (lock L1 )before returning it (holding the script-engine registry lock L2)

 - Bundle gets fully uninstalled.

 - New app-registry invalidation issued which tries to remove javascript engine executor from script-engine registry. It awaits lock L2

- After 10 seconds configuration lock L1 times out and installation continues normally.

rymsha:
Second Part:

There are many variations of this issue. For instance configuration may and up being provided, but bundle for an Application becomes uninstalled. In this case application will never get into normal state and will keep throwing NullPointerException on `ServiceRefImpl.findService` because `BundleContext.getServiceReference` always return null for invalid bundle.
Or Bundle gets uninstalled before  javascript engine executor gets created. In this case Bundle Classlader will keep throwing IllegalStateException: The bundle is uninstalled (this one is not permanent as it may only appear before step 5 )

PR #7991, opened 2020-04-01
Application fails to update properly under load #7978

rymsha:
`ApplicationListener` got deprecated as it was never used. `ApplicationListenerHub` responsible for `ApplicationListener`s notification is kept but no longer able to catch all application bundle events because it is now controlled by `ApplicationServiceImpl` which has no fully synchronous control over bundle/application binding . It was the main cause of issue: `ApplicationListenerHub` could invalidate Application instance too early (while it was still in STOPPING state), causing other threads to re-capture reference to a dying bundle.

`ApplicationService` invalidate method got deprecated. It is unreliable and can cause deadlocks.

Fallback to ConfigAdmin is kept is branch 7.2 to simplify PR readability, but will be removed in master.

`ScriptExecutorManager` now verifies `BundleContext` before creating an ScriptExecutor. Without such a check any frontend request could create an executor with `null` `BundleContext`, even if new Application Bundle was correctly captured (`applicationService.getInstalledApplication` may return Application even if bundle is just INSTALLED, but in such state bundles don't have context yet)

### DIFF
--- a/modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationRegistry.java
+++ b/modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationRegistry.java
@@ -72,9 +72,10 @@ final class ApplicationRegistry
 
     public void invalidate( final ApplicationKey key, final ApplicationInvalidationLevel level )
     {
+        LOG.debug( "Invalidating app {}", key );
         applications.computeIfPresent( key, ( k, v ) -> {
             callInvalidators( k, level );
-            v.unlatch();
+            v.setConfig( null );
             return null;
         } );
     }
@@ -96,8 +97,7 @@ final class ApplicationRegistry
                     callInvalidators( k, ApplicationInvalidationLevel.CACHE );
                 }
                 LOG.debug( "Configure and unlatch app {}", key );
-                v.application.setConfig( configuration );
-                v.unlatch();
+                v.setConfig( configuration );
                 return v;
             }
         } );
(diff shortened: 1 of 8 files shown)


modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationRegistry.java
```diff
@@ -96,8 +97,7 @@ public void setConfiguration( final ApplicationKey key, final Configuration conf
                     callInvalidators( k, ApplicationInvalidationLevel.CACHE );
                 }
                 LOG.debug( "Configure and unlatch app {}", key );
-                v.application.setConfig( configuration );
-                v.unlatch();
+                v.setConfig( configuration );
```
rymsha, line 100: Config happens after invalidation. Super-fast frontend request may grab old config for ScriptExecutor

rymsha: It is less critical. And it used to be like this before. Submitted new issue #7994

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationServiceImpl.java
```diff
@@ -579,7 +582,9 @@ private void uninstallBundle( final ApplicationKey applicationKey )
 
             if ( bundle != null )
             {
+                LOG.debug( "Uninstalling application {} bundle {}", applicationKey, bundle.getBundleId() );
```
anatol-sialitski, line 585: Why this log is used twice?

Issue #8004, opened 2020-04-06
GIF images should not be scaled


sgauruseu:
GIF images should be served as original binaries regardless of image service parameters. Now they are served correctly (as originals) in content thumbnail, media preview and image selector but attempted to be scaled inside image component and HTML Area which occasionally gives NPE because of high compression.

alansemenov:
![vegardsgif](https://user-images.githubusercontent.com/11612490/78542301-46ff8600-77f7-11ea-862e-42ca6350d686.gif)

rymsha:
```
Caused by: java.lang.ArrayIndexOutOfBoundsException: Index 4096 out of bounds for length 4096
        at java.desktop/com.sun.imageio.plugins.gif.GIFImageReader.read(GIFImageReader.java:992)
        at java.desktop/javax.imageio.ImageIO.read(ImageIO.java:1468)
        at java.desktop/javax.imageio.ImageIO.read(ImageIO.java:1363)
        at com.enonic.xp.image.ImageHelper.toBufferedImage(ImageHelper.java:54)
        at com.enonic.xp.core.impl.image.ImageServiceImpl.retrieveBufferedImage(ImageServiceImpl.java:233)


```
in reality issue is https://bugs.openjdk.java.net/browse/JDK-7132728

PR #8006, opened 2020-04-06
Errors after inserting a gif file in Live Edit and HtmlArea

alansemenov:


alansemenov:
@rymsha An animated gif gives NPE when it's being scaled or cropped server-side, so the best option is to serve the original blob and apply scaling client-side.

alansemenov:
But if you know how to implement scaling then please suggest a fix. The gif in question is attached to the original issue.

sigdestad:
Since when did scaling of animated gif's crash? This has been working for a long time afaik, and users only need to specify "use original" if they want to keep the animation.

alansemenov:
Could it be some new(er) format that we don't support, like WebP? It does seem to work on older Gifs, but not on Vegards.

rymsha:
> Since when did scaling of animated gif's crash?

Since people started to create enormously big GIF images with greatest compression tricks :)

> Could it be some new(er) format that we don't support, like WebP?

Well, it is "just" a bug in Java
https://github.com/enonic/xp/issues/8004#issuecomment-609807015

rymsha:
Rising another question: but is there, but turning off GIF scaling due to some special cases, is it worth it?

sigdestad:
I suggest we don't change how stuff works. We already have a solution where user can choose "use original image" to handle such special cases. If this is a Java bug, it will eventually be fixed - or we can start using a better library.

alansemenov:
Well, this is a showstopper for web editors. Not sure if you read the [original issue](https://github.com/enonic/app-contentstudio/issues/1664) from Vegard but he got an endless spinner after uploading his GIF because server crashed trying to serve the image preview (image preview is scaled to 768px by default). User doesn't get a chance to select "Original" because preview crashes rights away after the image is selected.

I guess we could change the preview logic, handle the error, show some placeholder and still allow selecting scaling options even if the preview failed, but this seems strange to me. Another alternative is to set "Original" by default (but this can result in serving a huge image by default). Again, this won't always help because in image component scaling is not up to user at all.

We definitely cannot wait for a fix of the bug that was reported 8 years ago and still hasn't been fixed.

sigdestad:
Proper solution is to handle this error better in the client imho. Then user can choose original and things will work for now. Also just happens for special images right now.

alansemenov:
It's not always up to user. Add an image component to the page -> select Vegard's gif from the selector -> crash. There you cannot choose original.

PR #8015, opened 2020-04-09
GIF images should not be scaled #8004

vbradnitski:


modules/core/core-image/src/main/java/com/enonic/xp/core/impl/image/ImageServiceImpl.java
```diff
@@ -53,67 +51,47 @@
     public ByteSource readImage( final ReadImageParams readImageParams )
         throws IOException
     {
-        final Path cachedImagePath = getCachedImagePath( readImageParams );
-        ByteSource imageByteSource = ImmutableFilesHelper.computeIfAbsent( cachedImagePath, () -> createImage( readImageParams ) );
-        return imageByteSource;
+        final String format = doGetFormatByMimeType( readImageParams.getMimeType() );
+
+        if ( renderAsASource( format ) )
```
anatol-sialitski, line 56: I am not sure that `A` needs here, I think `renderAsSource` will be better

modules/core/core-image/src/main/java/com/enonic/xp/core/impl/image/ImageServiceImpl.java
```diff
@@ -53,67 +51,47 @@
     public ByteSource readImage( final ReadImageParams readImageParams )
         throws IOException
     {
-        final Path cachedImagePath = getCachedImagePath( readImageParams );
-        ByteSource imageByteSource = ImmutableFilesHelper.computeIfAbsent( cachedImagePath, () -> createImage( readImageParams ) );
-        return imageByteSource;
+        final String format = doGetFormatByMimeType( readImageParams.getMimeType() );
```
rymsha, line 54: It seems unnecessary to extract format from MIME type here as we only ignore processing very well defined `image/gif` MIME. So we can remove `renderAsASource` altogether and replace it with plain `"image/gif".equals(readImageParams.getMimeType())`

modules/core/core-image/src/main/java/com/enonic/xp/core/impl/image/ImageServiceImpl.java
```diff
@@ -157,9 +135,6 @@ else if ( readImageParams.isScaleWidth() )
         //Filter string value
         final String filter = readImageParams.getFilterParam() != null ? readImageParams.getFilterParam() : "no-filter";
 
-        //Format string value
-        final String format = readImageParams.getFormat();
```
rymsha, line 161: Move calculation of "format" here, so it is also guarded by cache.

modules/core/core-image/src/test/java/com/enonic/xp/core/impl/image/ImageServiceImplTest.java
```diff
@@ -62,7 +60,7 @@ public void setUp()
         imageService.setImageFilterBuilder( imageFilterBuilder );
     }
 
-    @Test
+    /*@Test
```
rymsha, line 63: dead code. remove

Issue #8080, opened 2020-05-13
Resolving image/fragment component name fails when attached content is not found


ashklianko:
ContentNotFound exception is thrown

@sgauruseu commented on [Wed Feb 05 2020](https://github.com/enonic/app-contentstudio/issues/1472)

Case 1 :
1. Open Features , open Page Component View and create new fragment from existing text component. Close the wizard.
2. Expand Features and delete just created fragment.
3. Select Features and click on Edit button.
**Bug: Wizard page is not loaded, error in browser console.**

Case 2:
1. Open Features , open Page Component View and insert new image component
2. Upload an image and close the site
3. Expand Features and select just uploaded image.
4. Click on `Delete` button then click on `Delete now`.
![image](https://user-images.githubusercontent.com/3728712/73848985-31430400-483a-11ea-8198-7136821de7e7.png)

5. Select Features and click on Edit button.

**BUG:** 
-  Wizard page is not loaded, errors in console appear

![image](https://user-images.githubusercontent.com/3728712/73849261-b0383c80-483a-11ea-90bc-f5962fcf0c96.png)

- click on New button -  New content Dialog does not load and the same error in console .

**The same issue after deleting a fragment in the site.**

PR #8081, opened 2020-05-13
Resolving component name fails when attached content is not found #8080

ashklianko:
- Returning component type name when attached content removed
- Using try-catch clause to define if attached content is there or not (not using contentExists since it fetches node with try-catch itself)

### DIFF
--- a/modules/admin/admin-impl/src/test/java/com/enonic/xp/admin/impl/rest/resource/content/ComponentNameResolverImplTest.java
+++ b/modules/admin/admin-impl/src/test/java/com/enonic/xp/admin/impl/rest/resource/content/ComponentNameResolverImplTest.java
@@ -1,0 +1,125 @@
+package com.enonic.xp.admin.impl.rest.resource.content;
+
+import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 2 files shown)


modules/admin/admin-impl/src/test/java/com/enonic/xp/admin/impl/rest/resource/content/ComponentNameResolverImplTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+        Mockito.when( contentService.getById( imageComponent.getImage() ) ).thenThrow(
+            new ContentNotFoundException( imageComponentId, null ) );
+
+        final ComponentName result = componentNameResolver.resolve( imageComponent );
+
+        assertEquals( imageComponent.getName(), result );
+    }
+
+    public void testResolveEmptyFragmentComponent()
```
rymsha, line 73: Missing test annotation

ashklianko: Updated

Issue #8085, opened 2020-05-14
Mock API requires resource extension 


anatol-sialitski:
**CURRENT SITUATION**

In order to mock parts of API we should register internal services in test. That is very difficult for end-user, because he has to know internal logic of services.

For example, we want to mock content lib.

```
t.mock('/lib/xp/content', {
  get: function(params) {
    return { ... }
  }
});
```
In this case we have to add mock for services which are called in the `get` method

```
private ContentTypeService contentTypeService;
private MixinService mixinService;
private XDataService xDataService;

@Override
public void initialize()
    throws Exception
{
    super.initialize();

    this.contentTypeService = Mockito.mock( ContentTypeService.class );
    this.mixinService = Mockito.mock( MixinService.class );
    this.xDataService = Mockito.mock( XDataService.class );

    addService( ContentTypeService.class, this.contentTypeService );
    addService( MixinService.class, this.mixinService );
    addService( XDataService.class, this.xDataService );

    Content content = ... // build content here

    Mockito.when(contentService.getById( Mockito.any( ContentId.class ) ) ).thenReturn( content );
    Mockito.when(contentService.getByPath( Mockito.any( ContentPath.class ) ) ).thenReturn( content );
}
```
**SHOULD BE**

When some method is mocked, then a call of real method of service/lib should be avoided and replaced with a method decorator.

Details of implementation should be discussed.

rymsha:
rebranded to a bug, because mock was supposed to work similarly to require

PR #8104, opened 2020-05-20
Improve mock method for Test Framework #8085

anatol-sialitski:


modules/tools/testing/build.gradle
```diff
@@ -10,4 +10,7 @@ dependencies {
     compile "org.mockito:mockito-core:${mockitoVersion}"
     compile 'org.hamcrest:hamcrest-library:1.3'
     compile "org.slf4j:slf4j-simple:${slf4jVersion}"
+    testImplementation "com.enonic.xp:lib-content:${version}"
```
rymsha, line 13: This is circular dependency. It is simple to avoid if you put js files you want to mock in resources.

Issue #8079, opened 2020-05-13
Audit logs appear to always be created as "anonymous" user


sigdestad:
https://enonic-community.slack.com/archives/C73JLM8E7/p1587987894074000

Investigations seem to confirm this. This is a major error and should be corrected as soon as possible.

sigdestad:
Apparently, user seems to be correct for latest 7.3 build. Verify if this has been fixed for 7.3

alansemenov:
Relevant for 7.3-B1 at least

![image](https://user-images.githubusercontent.com/11612490/82561506-cb258700-9b73-11ea-8700-7862f1e1fb71.png)

PR #8109, opened 2020-05-22
Audit logs appear to always be created as "anonymous" user #8079

anatol-sialitski:


modules/core/core-content/src/test/java/com/enonic/xp/core/impl/content/ContentAuditLogSupportImplTest.java
```diff
@@ -0,0 +1,101 @@
...
+        final Context context = ContextBuilder.create().
+            branch( ContentConstants.BRANCH_DRAFT ).
+            repositoryId( RepositoryId.from( "test-repository" ) ).
+            authInfo( authInfo ).
+            build();
+
+        // test
+        context.runWith( () -> support.createContent( params, content ) );
+
+        executor.shutdown();
```
rymsha, line 89: Use CompletableFuture and it will look great!
(Tip: CF::runAsync has same signature as Executor::execute )

Issue #8132, opened 2020-06-05
Link to issue details is missing project context


ashklianko:
@alansemenov commented on [Fri Jun 05 2020](https://github.com/enonic/app-contentstudio/issues/1918)

When an issue is created we send out an email with issue details which contains a link to open the issue in the browser. This link is now missing the project context and user has to choose one after clicking the link.

rymsha:
Isn't it breaking change? Because now old CS won't find issues. Also all previously sent emails are invalid...

sigdestad:
Which is why we call it CS 3 (and everyone should use it from xp 7.3 and newer

PR #8133, opened 2020-06-05
Link to issue details is missing project context #8132

ashklianko:


modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/issue/IssueMailMessageGenerator.java
```diff
@@ -125,6 +127,8 @@ private String generateMessageBody()
             params.getLocaleMessageResolver().localizeMessage( "issue.email.showDetailsCaption", "Show Details..." );
         final String latestCommentTitle =
             params.getLocaleMessageResolver().localizeMessage( "issue.email.latestCommentTitle", "Latest comment" );
+        final String project =
+            ContextAccessor.current().getRepositoryId().toString().replace( ProjectConstants.PROJECT_REPO_ID_PREFIX, "" );
```
rymsha, line 131: ProjectName.from(...) should be used instead.

PR #8135, opened 2020-06-05
Component names: regression bug

vbradnitski:


modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/json/content/page/region/ImageComponentJson.java
```diff
@@ -21,7 +21,8 @@
     private final List<PropertyArrayJson> config;
 
     @JsonCreator
-    public ImageComponentJson( @JsonProperty("config") final List<PropertyArrayJson> config, @JsonProperty("image") final String image )
+    public ImageComponentJson( @JsonProperty("config") final List<PropertyArrayJson> config,
+                               @Deprecated @JsonProperty("image") final String image, @JsonProperty("name") final String name )
```
anatol-sialitski, line 25: @vbradnitski Why `image` is deprecated?

Issue #8159, opened 2020-06-16
Endless loop when calling application market


rymsha:
How to reproduce:

Running Any OS, any browser
Start clean sandbox with ~~7.3.0~~ 7.2.4 also affected
"login" to admin
Complete wizard by installing the tree apps
Wait ~8 minutes minutes

Result:
 100% CPU

rymsha:
```
"HttpClient-1-Worker-5" id=268 state=RUNNABLE
    at java.base@11.0.7/java.util.AbstractList.listIterator(AbstractList.java:311)
    at java.base@11.0.7/java.util.AbstractSequentialList.iterator(AbstractSequentialList.java:238)
    at java.base@11.0.7/java.util.Collections$SynchronizedCollection.iterator(Collections.java:2036)
    at platform/java.net.http@11.0.7/jdk.internal.net.http.common.SSLFlowDelegate.cleanList(SSLFlowDelegate.java:942)
    at platform/java.net.http@11.0.7/jdk.internal.net.http.common.SSLFlowDelegate$Writer.processData(SSLFlowDelegate.java:768)
    at platform/java.net.http@11.0.7/jdk.internal.net.http.common.SSLFlowDelegate$Writer$WriterDownstreamPusher.run(SSLFlowDelegate.java:646)
    at platform/java.net.http@11.0.7/jdk.internal.net.http.common.SequentialScheduler$CompleteRestartableTask.run(SequentialScheduler.java:147)
    at platform/java.net.http@11.0.7/jdk.internal.net.http.common.SequentialScheduler$SchedulableTask.run(SequentialScheduler.java:198)
    at platform/java.net.http@11.0.7/jdk.internal.net.http.common.SequentialScheduler.runOrSchedule(SequentialScheduler.java:271)
    at platform/java.net.http@11.0.7/jdk.internal.net.http.common.SequentialScheduler.runOrSchedule(SequentialScheduler.java:224)
    at platform/java.net.http@11.0.7/jdk.internal.net.http.common.SSLFlowDelegate$Writer.triggerWrite(SSLFlowDelegate.java:723)
    at platform/java.net.http@11.0.7/jdk.internal.net.http.common.SSLFlowDelegate.doHandshake(SSLFlowDelegate.java:1038)
    at platform/java.net.http@11.0.7/jdk.internal.net.http.common.SSLFlowDelegate.doClosure(SSLFlowDelegate.java:1124)
[...]

PR #8162, opened 2020-06-17
Endless loop when calling application market #8159

rymsha:
Endless loop caused by a bug in JDK 11.0.7 TLSv1.3 implementation
adding a file `security.properties` with `jdk.tls.disabledAlgorithms=TLSv1.3` and referencing it via `-Djava.security.properties="%DIRNAME%/security.properties"` turns off TLSv1.3


rymsha:
Not going to merge, as JDK 11.0.8 (hopefully) fixes the problem

Issue #8176, opened 2020-06-26
set system.hazelcast.initial.min.cluster.size to 2


rymsha:
Hazelcast is only used in clustered environment.
`system.hazelcast.initial.min.cluster.size=2` will ensure that "blind" cluster update from 7.3 to 7.4 (without opening hazelcast port between nodes) will issue messages in logs, and XP cluster won't start in split-brain condition.

For clusters with 6 nodes `system.hazelcast.initial.min.cluster.size` should be at least 4 but since we support clusters with 3 nodes - 2 is the default value we should use.

PR #8178, opened 2020-06-26
set system.hazelcast.initial.min.cluster.size to 2 #8176

rymsha:


sigdestad:
Hi. Quick question. What does this value mean? I.e. imagine I have a "minimal cluster" 3 master nodes, and 2 data nodes. XP should still be running even if I stop one of these nodes.

rymsha:
From our docs:

> system.hazelcast.initial.min.cluster.size
Initial expected cluster size to wait before the node to start completely.

Issue #8191, opened 2020-07-01
Project meta-data gets lost after dump restore


alansemenov:
After taking a dump and importing it into another instance, project meta-data gets lost on the way. Repositories and data get restored fine, but project meta-data is missing, so projects are not shown anywhere in the UI.

Before dump (target server):
![image](https://user-images.githubusercontent.com/11612490/86243479-c1496780-bba6-11ea-8e2b-c1ca1e6dadde.png)

After dump (source server):
![image](https://user-images.githubusercontent.com/11612490/86243417-a4149900-bba6-11ea-9ed1-2a6ab10139b1.png)

sigdestad:
Wow, how is this possible? This is just a regular node?

alansemenov:
Good question... It's either not included in the dump or (most likely) isn't restored correctly.

rymsha:
> Wow, how is this possible? This is just a regular node?

Because dumps-load uses high level Repository API to deal with creation of repositories to avoid reading metadata directly from the dump.


Some facts:
- Dumps taken on 7.3.0 are not corrupted. Problem appears on load.
- If Projects feature was not used, users won't notice the problem.

PR #8202, opened 2020-07-07
Project meta-data gets lost after dump restore #8191

vbradnitski:


modules/core/core-api/src/main/java/com/enonic/xp/repository/CreateRepositoryParams.java
```diff
@@ -13,6 +15,10 @@
 
     private final RepositorySettings repositorySettings;
 
+    private final PropertyTree data;
+
+    private final AttachedBinaries attachedBinaries;
```
rymsha, line 20: It might be a problem here.
BinaryAttachments is what users can post (reference + binary)
AttachedBinaries is what users can get (reference + blobKey)

So this API is suitable only for corner case when binary is already attached and we want to reattach it.

Issue #8217, opened 2020-07-10
POC: inherited filed should affect data inheritance, but not path


rymsha:
`inherited` should only control content data inheritance. Changing path of a content item must not reset `inherited` flag.

Implementation details:
Rename `inherited` flag to `inherit`. It lists what an item should inherit from a parent project's item. Any combination of `data` and `path` should be possible.

`[path, data]`- inherit path and data (equal to current `inherited = true`)
`[data]` - inherit only `data`, keep existing path the child project
`[]` - nothing is inherited (equal to current `inherited = false`) 
`[path]` - currently there is no use-case for this but should be supported.

When only file part (rename) of path gets changed in child project - inherit should unset both `path` and `data`
Otherwise (move) - inherit should unset only `data`

rymsha:
It could be implemented as 
- separate content fields ("inheritData", "inheritPath)
- as Integer with bitwise operations.
- as array of inherit types

Last one is a compromise between storage consumption and elasticsearch performance in case we need to use queries on this field.

PR #8246, opened 2020-07-22
POC: inherited filed should affect data inheritance, but not path #8217

vbradnitski:


modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/json/content/ContentSummaryJson.java
```diff
@@ -44,7 +48,7 @@ public ContentSummaryJson( final Content content, final ContentIconUrlResolver i
         this.thumbnailJson = content.hasThumbnail() ? new ThumbnailJson( content.getThumbnail() ) : null;
         this.isSite = content.isSite();
         this.isPage = content.hasPage();
-        this.inherited = content.isInherited();
+        this.inherit = new ArrayList( new TreeSet( content.getInherit() ) );
```
rymsha, line 51: Use stream.distinct

modules/core/core-api/src/main/java/com/enonic/xp/content/CreateContentParams.java
```diff
@@ -378,9 +380,12 @@ public Builder createSiteTemplateFolder( final boolean createSiteTemplateFolder
             return this;
         }
 
-        public Builder inherited( final boolean inherited )
+        public Builder inherit( final Set<ContentInheritType> inherit )
```
rymsha, line 383: Collection instead of Set

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/serializer/ContentDataSerializer.java
```diff
@@ -242,6 +246,17 @@ private void addWorkflowInfo( final PropertySet contentAsData, final WorkflowInf
         }
     }
 
+    private void addInherit( final PropertySet contentAsData, final Set<ContentInheritType> inherit )
+    {
+        if ( inherit != null )
+        {
+            contentAsData.ifNotNull().addStrings( INHERIT, inherit.
+                stream().
+                map( Enum::toString ).
```
rymsha, line 255: use Enum::name as it is guaranteed to be stable

modules/core/core-api/src/main/java/com/enonic/xp/content/Content.java
```diff
@@ -416,7 +418,7 @@ public int hashCode()
 
         protected boolean hasChildren;
 
-        protected boolean inherited;
+        protected ImmutableSet.Builder<ContentInheritType> inherit = ImmutableSet.builder();
```
rymsha, line 421: Immutable EnumSet

modules/core/core-api/src/main/java/com/enonic/xp/content/CreateContentParams.java
```diff
@@ -378,9 +381,12 @@ public Builder createSiteTemplateFolder( final boolean createSiteTemplateFolder
             return this;
         }
 
-        public Builder inherited( final boolean inherited )
+        public Builder inherit( final Collection<ContentInheritType> inherit )
         {
-            this.inherited = inherited;
+            if ( inherit != null )
```
rymsha, line 386: Why do you need null check?

modules/core/core-api/src/main/java/com/enonic/xp/content/CreateContentTranslatorParams.java
```diff
@@ -387,9 +389,12 @@ public Builder workflowInfo( final WorkflowInfo workflowInfo )
             return this;
         }
 
-        public Builder inherited( final boolean inherited )
+        public Builder inherit( final Set<ContentInheritType> inherit )
```
rymsha, line 392: Allow Collection

modules/core/core-api/src/main/java/com/enonic/xp/content/CreateContentTranslatorParams.java
```diff
@@ -387,9 +389,12 @@ public Builder workflowInfo( final WorkflowInfo workflowInfo )
...
         {
-            this.inherited = inherited;
+            if ( inherit != null )
```
rymsha, line 394: Why would you need null check?

modules/core/core-api/src/main/java/com/enonic/xp/content/UpdateContentParams.java
```diff
@@ -26,7 +27,7 @@
 
     private boolean requireValid;
 
-    private boolean inherited = false;
+    private Set<ContentInheritType> inherit = Set.of();
```
rymsha, line 30: It differs from CreateContentParams.
There inherit is an immutable final field set from a builder.

modules/core/core-api/src/main/java/com/enonic/xp/content/CreateContentParams.java
vbradnitski: to avoid NPE if someone will send it here(CreateContentParams is PublicApi)

Issue #8232, opened 2020-07-15
POC: filter by nodeid in project sync tree


rymsha:
It should be possible to provide nodeid "filter" when fetching project sync tree structure.

Only Projects containing node with provided id should be returned.

PR #8253, opened 2020-07-24
POC: filter by nodeid in project sync tree #8232

vbradnitski:


modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/project/ProjectResource.java
```diff
@@ -274,6 +304,12 @@ private TaskResultJson doApplyReadAccess( final ProjectName projectName, final P
             execute();
     }
 
+    private AuthenticationInfo adminAuthInfo()
```
rymsha, line 307: Why is this needed?
 I thought if user doesn't have rights to read contents it should not see that repository has such contentId.

sigdestad:
If this is for the "layers widget", I have not yet designed what should happen if user does not have access to layer (or content). I believe we should show that it exists in a layer (but do something special when presenting it in these cases.

Issue #8190, opened 2020-06-30
PublishInfo is missing in `getVersionsForView` API


alansemenov:
We currently return the following fields in [publishInfo](https://github.com/enonic/xp/blob/02c27721fbab69b89480173110c282469afc2723/modules/core/core-api/src/main/java/com/enonic/xp/content/ContentVersionPublishInfo.java#L8) object of `getVersionsForView` API: `message`, `publisher`, `publisherDisplayName` and `timestamp`.
We need these two fields added (from [ContentPublishInfo](https://github.com/enonic/xp/blob/master/modules/core/core-api/src/main/java/com/enonic/xp/content/ContentPublishInfo.java)): `publishFrom`, `publishTo`

PR #8305, opened 2020-08-27
Add `publishFrom` and `publishTo` fields to `getVersionsForView` API #8190

anatol-sialitski:


modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/json/content/ContentPublishInfoJson.java
```diff
@@ -36,4 +37,26 @@ public Instant getFirst()
     {
         return first;
     }
+
+    @Override
```
rymsha, line 41: It is often unnecessary  and even error-prone to write equals hashcode for DTOs.

rymsha (commented):
What

rymsha (approved):

vbradnitski (approved):

Issue #8355, opened 2020-09-11
Update javax.xml.bind dependency


rymsha:
~~None of internal apps depend on `javax.xml.bind`~~
Actually quite a few, see below

and 
~~this dependency problematic one in OSGi~~
Fixed in jaxb 2.3.3 version

rymsha:
https://issues.apache.org/jira/browse/TIKA-3094 is quite revealing. Tika uses Jaccess to parse metadata of microsoft access files. Jaccess uses Jaxb, not a great choice for Java 11 ant Tika is not going to bundle Jaxb either.

rymsha:
More info:
javax.ws.rs-api depends no it, so does jackson-jaxrs-json-provider

Tika parsers also depends directly and a lot of libraries in Tika bundle

rymsha:
This change became important as latest Tika depends on newer JAXB, so does Jackson-JAXB

PR #8356, opened 2020-09-11
Remove javax.xml.bind dependency #8355

rymsha:


rymsha:
https://issues.apache.org/jira/browse/TIKA-3094 is quite revealing. Tika uses Jaccess to parse metadata of microsoft access files. Jaccess uses Jaxb, not a great choice for Java 11 ant Tika is not going to bundle Jaxb either.

PR #8410, opened 2020-09-30
Highlighting in query doesn't work with count=-1 #8409

vbradnitski:


modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/elasticsearch/SearchRequestBuilderFactory.java
```diff
@@ -163,8 +187,22 @@ public Builder client( final Client client )
...
 
+        public Builder scrollTime( final TimeValue scrollTime )
+        {
+            this.scrollTime = scrollTime;
+            return this;
+        }
+
+        private void validate()
+        {
+            Preconditions.checkNotNull( query, "query must be set." );
```
anatol-sialitski, line 198: @vbradnitski I suggest to use `Objects.requireNonNull()` instead of `Preconditions.checkNotNull` because we are trying to avoid uses guava library.

rymsha: Better do it in one shot, not gradually. Leave it for now.

rymsha (approved):

anatol-sialitski (approved):

Issue #8460, opened 2020-11-09
Extra inherit flag in duplicated content


vbradnitski:
ContentInheritType.NAME should be removed from created/duplicated content if it's name was changed during sync.

PR #8461, opened 2020-11-09
Extra inherit flag in duplicated content #8460

vbradnitski:


rymsha (approved):

sigdestad:
What extra flag? I was not aware we needed more?

vbradnitski:
no, it's about `NAME` which shouldn't be in renamed content.

sigdestad:
Not sure if I get it? Is it a change of behaviour or did we add a new inherit flag?

rymsha:
Neither.
Just not very clear issue name.

Issue #8462, opened 2020-11-09
highlight fragmentSize and numberOfFragments fails with parseInt


ComLock:
Normal int works, and int via parseInt fails with:

class java.lang.Double cannot be cast to class java.lang.Integer (java.lang.Double and java.lang.Integer are in module java.base of loader 'bootstrap')

Reproduced in all versions from 7.2.0 unto 7.5.0

rymsha:
In a few places Java code expects Integer where JS can easily provide Double. parseInt is one of the examples where JS returns Double (despite method name)

rymsha:
Bugs are in lib-node / lib-content 
So applications must be recompiled with newer lib versions in order to get problem fixed

PR #8463, opened 2020-11-09
highlight fragmentSize and numberOfFragments fails with parseInt #8462

vbradnitski:


modules/lib/lib-content/src/main/resources/lib/xp/examples/content/query.js
```diff
@@ -112,7 +112,7 @@ var result = contentLib.query({
             displayName: {
                 fragmenter: 'span',
                 fragmentSize: 6,
-                noMatchSize: 6,
+                noMatchSize: parseInt('6', 6),
```
rymsha, line 115: Base 6?

rymsha (changes requested):
LGTM. Except parseInt base (must be 10)

rymsha (approved):

rymsha (approved):

Issue #8451, opened 2020-11-03
Extend Management with App events


rymsha:
A continuation of #8118
Implement SSE endpoint to receive global application events (started/stopped/installed/uninstalled)

endpoint should be `/app/events`

Prefer to reuse what Jetty already provides. Or, JAX.RS (Resteasy), if it speeds up development.

anatol-sialitski:
@gbbirkisson Could you please to specify the list of properties which should be returned in the DTO when a client was subscribed to the SSE endpoint (app list)?

gbbirkisson:
> @gbbirkisson Could you please to specify the list of properties which should be returned in the DTO when a client was subscribed to the SSE endpoint (app list)?

As you told me on slack, this has been decided. Im happy with the result.

PR #8465, opened 2020-11-09
  Extend Management with App events #8451

anatol-sialitski:


modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
```diff
@@ -0,0 +1,169 @@
...
+import com.enonic.xp.app.ApplicationNotFoundException;
+import com.enonic.xp.app.ApplicationService;
+import com.enonic.xp.event.Event;
+import com.enonic.xp.event.EventListener;
+import com.enonic.xp.impl.server.rest.model.ApplicationInfoJson;
+import com.enonic.xp.impl.server.rest.model.ApplicationUninstalledJson;
+import com.enonic.xp.impl.server.rest.model.ListApplicationJson;
+import com.enonic.xp.jaxrs.JaxRsComponent;
+
+@Path("sse")
```
rymsha, line 34: Should be in apps

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
```diff
@@ -0,0 +1,169 @@
...
+    @Context
+    public void setSse( final Sse sse )
+    {
+        this.sse = sse;
+        this.sseBroadcaster = sse.newBroadcaster();
+    }
+
+    @GET
+    @Path("subscribe-app-events")
+    @Produces(MediaType.SERVER_SENT_EVENTS)
```
rymsha, line 88: Should be events

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
```diff
@@ -0,0 +1,169 @@
...
+    public void subscribe( @Context SseEventSink sseEventSink )
+    {
+        this.sseBroadcaster.register( sseEventSink );
+
+        final OutboundSseEvent sseEvent = sse.newEventBuilder().
+            name( "List of applications" ).
```
rymsha, line 94: Just `list`

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
```diff
@@ -0,0 +1,169 @@
...
+                collect( Collectors.toList() ) ) ).
+            comment( "List of installed applications" ).build();
+
+        sseEventSink.send( sseEvent );
+    }
+
+    @Override
+    public void onEvent( final Event event )
+    {
+        if ( EVENT_TYPE.equals( event.getType() ) )
```
rymsha, line 108: We maybe doing something wrong here. These events are global, but we probably need local (from bundle events)
Need to discuss.

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
```diff
@@ -0,0 +1,169 @@
...
+                            LOG.debug( "Ignoring {} {}", EVENT_TYPE, eventSubType );
+                            break;
+                    }
+                } );
+        }
+    }
+
+    private void handleEvent( final Event event, final String eventSubType )
+    {
+        if ( sseBroadcaster == null )
```
rymsha, line 131: Isn't it so that broadcaster always initialized?

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
anatol-sialitski: Done

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
anatol-sialitski: Done

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
anatol-sialitski: Done

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
```diff
@@ -0,0 +1,169 @@
...
+import com.enonic.xp.app.ApplicationService;
+import com.enonic.xp.event.Event;
+import com.enonic.xp.event.EventListener;
+import com.enonic.xp.impl.server.rest.model.ApplicationInfoJson;
+import com.enonic.xp.impl.server.rest.model.ApplicationUninstalledJson;
+import com.enonic.xp.impl.server.rest.model.ListApplicationJson;
+import com.enonic.xp.jaxrs.JaxRsComponent;
+
+@Path("apps")
+@Component(immediate = true, property = "group=api")
```
rymsha, line 35: Must be admins only

anatol-sialitski: Done

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
anatol-sialitski: Done

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/event/ApplicationClusterEvents.java
```diff
@@ -34,6 +34,7 @@ public static Event installed( final Node applicationNode )
             distributed( true ).
             value( EVENT_TYPE_KEY, INSTALLED ).
             value( NODE_ID_PARAM, applicationNode.id() ).
+            value( APPLICATION_KEY_PARAM, applicationNode.name().toString() ).
```
rymsha, line 37: I think toString is not needed

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
```diff
@@ -0,0 +1,186 @@
...
+            if ( UNINSTALLED.equals( eventSubType ) )
+            {
+                eventBuilder.data( ApplicationUninstalledJson.class,
+                                   new ApplicationUninstalledJson( event.getValue( APPLICATION_KEY_PARAM ).get().toString() ) );
+            }
+            else
+            {
+                final ApplicationKey applicationKey = ApplicationKey.from( event.getValue( APPLICATION_KEY_PARAM ).get().toString() );
+
+                final Application application = applicationService.getInstalledApplication( applicationKey );
```
rymsha, line 156: It looks like overkill for state event (on start/stop) to fetch app details? But maybe @gbbirkisson has a use for it?

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
```diff
@@ -0,0 +1,186 @@
...
+
+                if ( application == null )
+                {
+                    throw new ApplicationNotFoundException( applicationKey );
```
rymsha, line 160: There are no users of this exception. A WARN log should be enough.

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
gbbirkisson: I personally like the model K8s uses, that is send event type and the whole object related to that event. If it is not a huge performance problem, I think we should keep it. It will probably come in handy.

gbbirkisson (approved):

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
rymsha: In current implementation it is open for race conditions. That is my main concern.
But 
with installed event it is a race anyway
and race chance is low.

rymsha (approved):

PR #8471, opened 2020-11-11
Codacy fixes #8469

### LINKED ISSUE #8469: Some Codacy fixes

rymsha:


vbradnitski (approved):

modules/core/core-api/src/main/java/com/enonic/xp/node/NodePath.java
```diff
@@ -300,7 +300,7 @@ public boolean equals( final Object o )
 
             final Element element = (Element) o;
 
-            return Objects.equals( this.name.toLowerCase(), element.name.toLowerCase() );
+            return this.name.equalsIgnoreCase( element.name );
```
anatol-sialitski, line 303: name can be null here

rymsha: In the only place where Element gets created there is a null guard
https://github.com/enonic/xp/pull/8471/files#diff-53e21a0ee070cb8f2eeaa93e069082f67f5f4ebdbd4a6ec89de1ecaf5ec37c9dR399

anatol-sialitski (approved):

PR #8472, opened 2020-11-11
Distributed tasks

rymsha:


modules/core/core-task/src/main/java/com/enonic/xp/impl/task/MemberAttributesApplier.java
```diff
@@ -0,0 +1,72 @@
...
+    public void activate( final TaskConfig config )
+    {
+        localMember.setBooleanAttribute( TASKS_ENABLED_ATTRIBUTE_KEY, config.offload_acceptInbound() );
+        super.open();
+    }
+
+    @Deactivate
+    public void deactivate()
+    {
+        super.close();
```
rymsha, line 45: Check that all attributes removed

anatol-sialitski (approved):

modules/core/core-task/src/main/java/com/enonic/xp/impl/task/MemberAttributesApplier.java
rymsha: yes, removedBundle called on close

vbradnitski (approved):

Issue #8492, opened 2020-11-20
serviced support


anatol-sialitski:
@rymsha commented on [Tue Oct 15 2019](https://github.com/enonic/xp-distro/issues/87)

For instance, setting up a service on Fedora 30 is not that simple.

---

@jsi commented on [Wed Nov 06 2019](https://github.com/enonic/xp-distro/issues/87#issuecomment-550355435)

When this issue is implemented, it should be followed by documentation.  We do not have very good operations doc for 7.x, but it would improve greatly if this guide is written: https://github.com/enonic/doc-xp/issues/94

PR #8493, opened 2020-11-20
serviced support #8492

anatol-sialitski:


rymsha:
@gbbirkisson any suggestions to improve this?

modules/runtime/src/service/systemd/xp.service
```diff
@@ -0,0 +1,40 @@
...
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+Type=simple
+PrivateTmp=true
+Environment=XP_INSTALL=/opt/enonic/xp
+Environment=XP_JAVA_HOME=${XP_INSTALL}/jdk
+
+WorkingDirectory=/opt/enonic/xp
```
rymsha, line 13: `/opt` Is it commonly used directory nowadays?

modules/runtime/src/service/systemd/xp.service
```diff
@@ -0,0 +1,40 @@
...
+
+LimitFSIZE=infinity
+
+KillSignal=SIGTERM
+
+KillMode=process
+
+SendSIGKILL=no
+
+SuccessExitStatus=143
```
rymsha, line 37: does this apply to sh script? we don't directly call java, but does script propagate same java exit code (143 means java exited normally after SIGTERM)

modules/runtime/src/service/systemd/xp.service
gbbirkisson:
```
/opt – Optional Packages

The /opt directory contains subdirectories for optional software packages. It’s commonly used by proprietary software that doesn’t obey the standard file system hierarchy – for example, a proprietary program might dump its files in /opt/application when you install it.
```

So if the XP distro is installed there this makes sense.

modules/runtime/src/service/systemd/xp.service
rymsha: @anatol-sialitski can you verify this, please

gbbirkisson: Doing a test on my machine:

```console
$XP_HOME=$HOME/.enonic/sandboxes/jwt/home ./bin/server.sh

....

CTRL+C

$echo $?
130
```

gbbirkisson: Killing with sigterm results in 143.

rymsha (approved):

Issue #8485, opened 2020-11-16
Return full list of ids of invalid/in progress items


alansemenov:
Add two new fields to response of `resolvePublishContent` API containing arrays of all ids of _invalid_ and _in progress_
items. Keep `containsInvalid` and `containsNotReady` fields for backwards compatibility.

PR #8499, opened 2020-11-24
Return full list of ids of invalid/in progress items in resolvePublis…

ashklianko:
…hContent API #8485


rymsha (changes requested):
Please amend commit with a shorter description, so issue number is not lost

Issue #8484, opened 2020-11-16
Enable batch exclusion in resolvePublishContent API


alansemenov:
Currently `resolvePublishContent` API takes the list of items to exclude from publishing in `excludedIds` parameter. We should keep this for exclusion of specific items, and implement a new parameter for excluding ALL items (both primary and dependent) with specific state, which is as of moment can be "`invalid`" or "`in progress`" or both.

So, for example, calling `resolvePublishContent` with `excludeState = ['INVALID']` should exclude all of the items from the original list with `isValid=false`, and return a new resolved list. Sending `excludeState = ['IN_PROGRESS']` will exclude items with `workflow.state="IN_PROGRESS"`.

NB! Items required for publishing of the primary items should not be excluded.

PR #8502, opened 2020-11-24
Enable batch exclusion in resolvePublishContent API #8484

vbradnitski:


modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/content/json/ResolvePublishDependenciesJson.java
```diff
@@ -41,4 +45,16 @@ public void setExcludeChildrenIds( final Set<String> excludeChildrenIds )
...
     }
+
+    public Set<PublishDependencyExcludeState> getExcludeStates()
+    {
+        return excludeStates;
+    }
+
+    public void setExcludeStates( final Set<String> excludeStates )
+    {
+        this.excludeStates = excludeStates != null
```
rymsha, line 56: do you ever set null there?

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/ContentAuditLogSupportImpl.java
```diff
@@ -303,6 +304,15 @@ private void doPublish( final PushContentParams params, final PublishContentResu
             paramsSet.addStrings( "excludeChildrenIds", params.getExcludeChildrenIds().stream().
                 map( ContentId::toString ).collect( Collectors.toList() ) );
         }
+
+        if ( params.getExcludeWorkflowStates() != null )
```
rymsha, line 308: This makes me think that it is possible to publish invalid or in-progress content
Don't we need a guard for this is content service anyway?

modules/core/core-api/src/main/java/com/enonic/xp/content/ResolvePublishDependenciesParams.java
```diff
@@ -85,6 +112,18 @@ public Builder target( Branch target )
             return this;
         }
 
+        public Builder excludeInvalid( boolean excludeInvalid )
+        {
+            this.excludeInvalid = excludeInvalid;
+            return this;
+        }
+
+        public Builder addExcludeWorkflowState( WorkflowState workflowState )
```
rymsha, line 121: Maybe Change to setExcludeWorkflowStates()?

rymsha (approved):

anatol-sialitski (approved):

Issue #8504, opened 2020-11-25
Add `originProject` field to /content and /list API response


alansemenov:

rymsha:
What is it for?

alansemenov:
https://github.com/enonic/app-contentstudio/issues/2550

PR #8505, opened 2020-11-25
Add `originProject` field to /content and /list API response #8504

ashklianko:


anatol-sialitski (approved):

rymsha (changes requested):
test missing where originProject is not null

ashklianko:
> test missing where originProject is not null

Updated

rymsha (approved):

vbradnitski (approved):

Issue #8482, opened 2020-11-16
Content Wizard - Custom icon is not reverted after clicking on Reset button


sgauruseu:
1. Create a layer(En language) in Default project 
2. Go to layer's context, open `Features` and upload custom icon, save it .
3. `Revert` button gets visible in the toolbar, click on it.
 
**BUG** - custom icon remains in the wizard after resetting  changes.

![image](https://user-images.githubusercontent.com/3728712/99229504-2a9ce080-27ff-11eb-830e-3f8b8ac671ec.png)

![image](https://user-images.githubusercontent.com/3728712/99229609-4c966300-27ff-11eb-842b-3872291ec967.png)

PR #8516, opened 2020-11-30
Content icon is not reverted after inheritance reset #8482

vbradnitski:


rymsha (approved):

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/UpdatedEventSyncCommand.java
```diff
@@ -87,27 +88,34 @@ private void doSyncMedia( final ContentEventSyncCommandParams params, final Upda
 
     private void doSyncThumbnail( final ContentEventSyncCommandParams params, final UpdateContentParams updateParams )
     {
-        if ( params.getSourceContent().hasThumbnail() &&
-            !params.getSourceContent().getThumbnail().equals( params.getTargetContent().getThumbnail() ) )
+        if ( !Objects.equals( params.getSourceContent().getThumbnail(), params.getTargetContent().getThumbnail() ) )
         {
             final Thumbnail sourceThumbnail = params.getSourceContent().getThumbnail();
+            final Thumbnail targetThumbnail = params.getTargetContent().getThumbnail();
```
anatol-sialitski, line 94: Please to declare the `targetThumbnail` variable in the `else` block, if `sourceThumbnail` is not null then that variable not needed.

anatol-sialitski (approved):

Issue #8518, opened 2020-11-30
Management SSE for apps fails when Accept-Encoding is sent


gbbirkisson:
You can make the SSE endpoint for apps return a `200` but still fail by adding `Accept-Encoding: gzip, deflate` to the request headers.

```console
$ curl -v --user 'su:bla' -N \
    -H 'Accept: text/event-stream' \
    -H 'Accept-Encoding: gzip, deflate' \
    http://localhost:4848/app/events

*   Trying 127.0.0.1:4848...
* Connected to localhost (127.0.0.1) port 4848 (#0)
* Server auth using Basic with user 'su'
> GET /app/events HTTP/1.1
> Host: localhost:4848
> Authorization: Basic blabla
> User-Agent: curl/7.73.0
> Accept: text/event-stream
> Accept-Encoding: gzip, deflate
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Date: Mon, 30 Nov 2020 15:30:18 GMT
< Set-Cookie: JSESSIONID=s0rk7cjder9q1ev2utc5xkp5u66; Path=/; HttpOnly
< Expires: Thu, 01 Jan 1970 00:00:00 GMT
< Content-Type: text/event-stream
< Vary: Accept-Encoding
< Content-Encoding: gzip
< Transfer-Encoding: chunked
< 
Warning: Binary output can mess up your terminal. Use "--output -" to tell 
Warning: curl to output it to your terminal anyway, or consider "--output 
Warning: <FILE>" to save to a file.
* Failure writing output to destination
* Failed reading the chunked-encoded stream
* Closing connection 0
```

Remove the `Accept-Encoding` header and everything is fine.

anatol-sialitski:
@gbbirkisson try to use `-H 'Content-Type: text/event-stream' -H 'Content-Encoding: gzip, deflate'` instead

rymsha:
@anatol-sialitski 
Well, your option will work, but only because it won't have Accept-Encoding hear.
Apache Httpclient sends Accept-Encoding, and for yet unknown reason it is problematic.
I'm on this issue

rymsha:
Jetty and Resteasy (at least 3.x) SSE implementation don't work well together when client wants gzip (advertises via Accept-Encoding)

Easiest way for now is to disable compression for `text/event-stream` responses.

rymsha:
https://github.com/eclipse/jetty.project/issues/4317

PR #8520, opened 2020-12-01
Management SSE for apps fails when Accept-Encoding is sent #8518

rymsha:
Main fix: disabled gzip for SSE. At least for now. see https://github.com/enonic/xp/issues/8518#issuecomment-736538234

A few other polishes:
- For some reason Resteasy injects Context twice. 
https://github.com/resteasy/Resteasy/blob/master/resteasy-core/src/main/java/org/jboss/resteasy/core/ResourceMethodRegistry.java#L228
https://github.com/resteasy/Resteasy/blob/master/resteasy-core/src/main/java/org/jboss/resteasy/core/ResourceMethodRegistry.java#L246
~~TODO: Create an issue for Resteasy?~~ https://issues.redhat.com/browse/RESTEASY-2779

That causes race conditions for us, so sse Context variable is set, I ignore others

- In case of not yet set contextHolder we should throw an error (well, it never happens, but still...), otherwise, if no error is thrown, client just hangs and never receives any messages

- Last fix is to send "list" message before any others can potentially arrive.

### DIFF
--- a/modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
+++ b/modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SseEntryPoint.java
@@ -76,7 +76,11 @@ public class SseEntryPoint
     @Context
     public void setSse( final Sse sse )
     {
-        this.contextHolder = new SseContextHolder( sse );
+        final SseContextHolder ctx = contextHolder;
+        if ( ctx == null )
+        {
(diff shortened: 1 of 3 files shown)


anatol-sialitski (approved):

rymsha:
issue about double injection submitted https://issues.redhat.com/browse/RESTEASY-2779

gbbirkisson (approved):

vbradnitski (approved):

Issue #8404, opened 2020-09-24
Duplicate headers in controller creates a crash.


poi33:
A user reported that when they have a part sending in a header that is allready present it crashes. 

A part returns with a header. The part is added twice on the same site. This creates com.enonic.xp.web.WebException: Multiple entries with same key: <key name>” 

The system merges: pageContributions and cookies, but not headers?

rymsha:
New behavior after #8521 is that duplicate header from different parts won't generate an error.
Instead response will have only one header value for each key (last one wins)

HTTP Header specs went through a few RFCs and still problematic to implement correctly 

But generally "duplicate header key" is exotic:
- It may be used for Accept* headers but, those are mostly client ones
- Well known exception Set-Cookie https://tools.ietf.org/html/rfc6265#section-3 But we support cookies natively via cookies response object. 
- It may be used for Link header, but Link header specification always workarounds it with `,` separated values. We may soon support Links natively #8197
- Other headers that need multiple values have an official workaround to "combine" them. https://tools.ietf.org/html/rfc7230#section-3.2.2

Current implementation of Enonic XP Web Response is written so that keys are unique (it is a js/json object), changing it in a non-breaking way would be too much effort for an exotic use-case. If there is a demand for this, we can open a new feature request.

PR #8521, opened 2020-12-02
Duplicate headers in controller creates a crash. #8404

rymsha:
There were a few iterations of HTTP header specs and how they should handled

Firstly they are case-insensitive preferably lowercased
https://tools.ietf.org/html/rfc7540#section-8.1.2

Secondly the order is not significant
https://tools.ietf.org/html/rfc7230#section-3.2.2

So, it is safe to lowercase them and when response is built, headers get populated into case-insensitive sorted map (for backwards compatibility, so `location = headers['Location']` works equally to `location = headers['location']`)

Instead of throwing an Exception, it is possible to replace previous header values with new ones.
This is the easiest fix for now. More info in https://github.com/enonic/xp/issues/8404#issuecomment-737301463

### DIFF
--- a/modules/web/web-api/src/main/java/com/enonic/xp/web/WebResponse.java
+++ b/modules/web/web-api/src/main/java/com/enonic/xp/web/WebResponse.java
@@ -1,7 +1,10 @@
 package com.enonic.xp.web;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 
 import javax.servlet.http.Cookie;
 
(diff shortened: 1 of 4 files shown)


anatol-sialitski (approved):

modules/web/web-api/src/main/java/com/enonic/xp/web/WebResponse.java
```diff
@@ -106,7 +109,9 @@ public WebSocketConfig getWebSocket()
     {
         private Object body;
 
-        private ImmutableMap.Builder<String, String> headers;
+        private Map<String, String> headers = new HashMap<>();
+
+        ;
```
rymsha, line 114:
```suggestion
        
```

modules/web/web-api/src/main/java/com/enonic/xp/web/WebResponse.java
```diff
@@ -140,27 +145,19 @@ public T body( final Object body )
...
-            if ( this.headers == null )
-            {
-                clearHeaders();
-            }
-            this.headers.put( key, value );
+            this.headers.put( key.toLowerCase( Locale.ROOT ), Objects.requireNonNull( value ) );
             return (T) this;
         }
 
         public T clearHeaders()
```
rymsha, line 158: Don't use in constructor

anatol-sialitski (approved):

vbradnitski (approved):

Issue #8344, opened 2020-09-10
lib-websocket should have access to User


rymsha:
Currently there is no way to access authenticated user in websocket callbacks. This makes it hard to authorize call made within a websocket scope.

javax.websocket provides facility to access Principal, but current implementation does not us it, thus it is not possible to share with websocketCallback function.

rymsha:
One can access  user via `authLib.getUser()` already. 

If it is token-based authentication one should call authLib.login with scope REQUEST with a correct idProvider, in order for user object to be populated into context.

So you have to remind me @gbbirkisson @edloidas what is a use case again?

gbbirkisson:
The issue is because of the handler for websockets. The handler is not a closure but only a function reference. So inside the handler/function, there is no context about what user opened up the socket.

PR #8529, opened 2020-12-04
lib-websocket should have access to User #8344

rymsha:


sigdestad:
I do not know enough about WebSockets, but ideally one should have access to same context as in regular http controllers

vbradnitski (approved):

anatol-sialitski (approved):

edloidas (approved):

Issue #8511, opened 2020-11-27
Anonymous audit-log events


poi33:
* Creating content 
* Update content
* Renaming content 
* Changing child order
All the above changes can sometimes create an aduit-log with the anonymous user:

qa.enonic.com:
Update:
<img src="https://user-images.githubusercontent.com/3340349/100455113-746aae00-30be-11eb-9eb4-78f45301a4eb.png" width="500px"/>

Some instances will do this with SU instead: 
<img src="https://user-images.githubusercontent.com/3340349/100455703-7123f200-30bf-11eb-8661-20aa6b95001b.png" width="200px"/>

PR #8531, opened 2020-12-07
Anonymous audit-log events #8511

anatol-sialitski:


modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/content/json/UpdateContentJson.java
```diff
@@ -59,10 +60,14 @@
...
 
+        final PrincipalKey modifier = ContextAccessor.current().getAuthInfo().getUser() != null
+            ? ContextAccessor.current().getAuthInfo().getUser().getKey()
+            : PrincipalKey.ofAnonymous();
+
         this.updateContentParams = new UpdateContentParams().
             requireValid( Boolean.parseBoolean( requireValid ) ).
             contentId( ContentId.from( contentId ) ).
-            modifier( PrincipalKey.ofAnonymous() ).
+            modifier( modifier ).
```
rymsha, line 70: Do we use this field at all? We should not, probably as we should not trust the caller.

rymsha (changes requested):

modules/core/core-api/src/main/java/com/enonic/xp/content/UpdateContentParams.java
```diff
@@ -40,6 +40,7 @@ public UpdateContentParams contentId( final ContentId contentId )
         return this;
     }
 
+    @Deprecated
     public UpdateContentParams modifier( final PrincipalKey modifier )
     {
         this.modifier = modifier;
```
rymsha, line 46:
```suggestion
```

rymsha (changes requested):

modules/core/core-api/src/main/java/com/enonic/xp/content/UpdateContentParams.java
rymsha: And all unused private fields should be deleted as well

rymsha (approved):

vbradnitski (approved):

Issue #8533, opened 2020-12-07
Replace StringEscapeUtils with alternatives


rymsha:
Continuation of  #7648

PR #8534, opened 2020-12-07
Replace StringEscapeUtils with alternatives #8533

rymsha:


modules/core/core-api/src/main/java/com/enonic/xp/util/HtmlHelper.java
```diff
@@ -0,0 +1,132 @@
...
+import org.jsoup.Jsoup;
+import org.jsoup.nodes.Element;
+import org.jsoup.nodes.Node;
+import org.jsoup.nodes.TextNode;
+import org.jsoup.parser.Parser;
+import org.jsoup.parser.Tag;
+import org.jsoup.select.NodeTraversor;
+import org.jsoup.select.NodeVisitor;
+
+public final class HtmlHelper
```
rymsha, line 12: move to internal

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/macro/EmbedIframeMacroProcessor.java
```diff
@@ -29,7 +29,7 @@ public PortalResponse process( final MacroContext macroContext )
             return generateNonIframeResponse( macroContext );
         }
 
-        return generateResponse( StringEscapeUtils.unescapeHtml( macroContext.getBody() ) );
+        return generateResponse( HtmlHelper.unescape( macroContext.getBody() ) );
```
rymsha, line 32: why unescape here?

modules/core/core-macro/src/main/java/com/enonic/xp/impl/macro/MacroParser.java
```diff
@@ -130,7 +129,7 @@ private void parseAttribute()
         match( '=' );
         ws();
         match( '"' );
-        final String value = StringEscapeUtils.unescapeHtml( parseAttributeValue() );
+        final String value = HtmlHelper.unescape( parseAttributeValue() );
```
rymsha, line 132: why unescape here?

modules/core/core-api/src/main/java/com/enonic/xp/index/HtmlStripper.java
```diff
@@ -25,18 +18,15 @@ public Value process( final Value value )
             return value;
         }
 
-        final Matcher matcher = XML_TAG_PATTERN.matcher( value.toString() );
-        final String strippedHtml = matcher.replaceAll( " " );
-        final String unescapedHtml = StringEscapeUtils.unescapeHtml( strippedHtml );
+        final String htmlStripped = HtmlHelper.htmlToTextLegacy( value.asString() );
```
rymsha, line 21: maybe don't use legacy at all. intention is to generate indexed text from html

modules/core/core-macro/src/main/java/com/enonic/xp/impl/macro/MacroParser.java
rymsha: because macro defined in html

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/macro/EmbedIframeMacroProcessor.java
rymsha: see #8535

anatol-sialitski (approved):

vbradnitski (approved):

Issue #8555, opened 2020-12-15
Project sync mixes up contexts due to race condition


vbradnitski:
nodeService.refresh() command is missing, so there is a time gap between new content data update and it's accessibility from search.

PR #8557, opened 2020-12-15
index refresh is missing for UpdateContentCommand #8555

vbradnitski:


rymsha (changes requested):
likely there is no issue with that. Test does not reveal it. Instead Sync component has an ugly race condition.

Issue #8576, opened 2020-12-23
Incorrect language of a local child item created under inherited non-localised parent


ashklianko:
@alansemenov commented on [Tue Dec 22 2020](https://github.com/enonic/app-contentstudio/issues/2681)

When a new child content is created under existing content, we set language of the parent on this child content.
This behaviour should be different when a child content is created under **inherited non-localised** content. In this case we should set language of the current layer on the child content.

ashklianko:
To be fixed in `CreateContentCommand.getDefaultLanguage()`
Also I think we need CreateContentJson to also have 'language' parameter so language could be set from create request; note that `CreateContentHandler` in lib-content has language as param

ashklianko:
If task is OK with everyone then can assign on me

rymsha:
Too easy to break layers sync. Leave it for @vbradnitski

vbradnitski:
> To be fixed in `CreateContentCommand.getDefaultLanguage()`
> Also I think we need CreateContentJson to have 'language' parameter so language could be set from create request; note that `CreateContentHandler` in lib-content has language as param

We can't force all lib users to set language field in every such request.

PR #8585, opened 2021-01-04
Incorrect language of a local child item created under inherited non-…

vbradnitski:
…localised parent #8576

### DIFF
--- a/modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/CreateContentCommand.java
+++ b/modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/CreateContentCommand.java
@@ -1,6 +1,7 @@
 package com.enonic.xp.core.impl.content;
 
 import java.time.Instant;
+import java.util.List;
 import java.util.Locale;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/CreateContentCommand.java
```diff
@@ -331,7 +334,23 @@ private Locale getDefaultLanguage( final CreateContentParams createContentParams
         {
             final Node parent = nodeService.getByPath( ContentNodeHelper.translateContentParentToNodeParentPath( parentPath ) );
 
-            final String language = parent.data().getString( ContentPropertyNames.LANGUAGE );
+            final List<Property> inheritProperties = parent.data().getProperties( ContentPropertyNames.INHERIT );
+            final boolean inherited = inheritProperties.stream().
+                anyMatch( property -> ContentInheritType.CONTENT.name().equals( property.getString() ) );
+
+            String language;
```
rymsha, line 341:
```suggestion
            final String language;
```

rymsha (approved):
Commit message should be shorter

anatol-sialitski (approved):

Issue #4407, opened 2017-02-13
Extend term-aggregation options in api


runarmyklebust:
There are available term-aggregation parameters that could be useful to expose in JS-API's.
This should be done for both node.query and content.query

* minimumDocCount

````
 aggregations: {
        floors: {
            terms: {
                field: "data.number_floor",
                order: "_count asc",
                minimumDocCount: 10
            }
````

anatol-sialitski:
Property should be named 'minDocCount' to be unified with the rest of the code

PR #8590, opened 2021-01-05
Extend term-aggregation options in api #4407

anatol-sialitski:


modules/core/core-api/src/main/java/com/enonic/xp/query/aggregation/TermsAggregationQuery.java
```diff
@@ -50,7 +50,7 @@ public Type getOrderType()
         return orderType;
     }
 
-    public long getMinDocCount()
+    public Long getMinDocCount()
```
rymsha, line 53: This is impossible to change return type in API.

modules/core/core-api/src/main/java/com/enonic/xp/query/aggregation/TermsAggregationQuery.java
```diff
@@ -103,7 +103,7 @@ public Builder size( final Integer size )
             return this;
         }
 
-        public Builder minDoccount( final long minDocCount )
+        public Builder minDocCount( final Long minDocCount )
```
rymsha, line 106: it is impossible to change method name and arg types in API

rymsha (changes requested):

modules/core/core-api/src/main/java/com/enonic/xp/query/aggregation/TermsAggregationQuery.java
anatol-sialitski: Done

modules/core/core-api/src/main/java/com/enonic/xp/query/aggregation/TermsAggregationQuery.java
anatol-sialitski: Done

vbradnitski (approved):

Issue #8575, opened 2020-12-23
Admin API getDescendantsOfContents should find children by id, not path


alansemenov:
`getDescendantsOfContents` in admin API now fetches child items by parent `path`, but it should use `id` instead. Current behaviour is the reason for [this bug](https://github.com/enonic/xp/issues/8435), when paths are different in the draft and master branches.

rymsha:
Related to #2711

rymsha:
new method:
`resolveForUnpublish`
accept set of content ids: `cids`
for each id in `cids` find descendance ids in draft: `dcids` <- (I would skip this step, as it is complex to explain why some unrelated items in muster suddenly get unpublished). 
for each id `cids` and `dcids` find ids and descendance ids in master: `unids`
return `unids`

`resolveForDelete`
is out of some, but the result is more complicated. Some content ids could be purely deleted, but some should be unpublished only. see #8559

rymsha:
Duplicate of #8435

Issue #8435, opened 2020-10-16
Moved content cannot be unpublished


alansemenov:
Reported in [support case](https://enonic.zendesk.com/agent/tickets/3998)

Steps to reproduce:

1. Publish a content
2. Move the content
3. Unpublish the content

Server will respond with "_Nothing to unpublish_".

Correct behaviour is to find the published version at "old" path in the master and unpublish it.

rymsha:
related to #8559

rymsha:
currently it looks like bug origins from CS:
There are two problems:
- For moved items it requests `getDescendantsOfContents` with empty paths
- Even if path would not be empty, moved content descendants must be enquired from master branch by contentId instead.

rymsha:
Duplicate of #8435

rymsha:
waiting for https://github.com/enonic/app-contentstudio/issues/2756

PR #8591, opened 2021-01-05
Admin API getDescendantsOfContents should find children by id, not pa…

vbradnitski:
…th #8575


anatol-sialitski (approved):

rymsha (changes requested):
This should be a pool for #8435

PR #8595, opened 2021-01-06
treeSelectorQuery uses deprecated method after performance fix  #8577

### LINKED ISSUE #8577: treeSelectorQuery uses deprecated method after performance fix
This is a continuation of #8434

It is somehow related to #5688 but both of the fixes didn't take into account that `treeSelectorQuery` method uses deprecated 
 `contentService.findContentPaths`

Firstly, find out why method was deprecated, and secondly don't use this deprecated method in production code.

vbradnitski:


modules/core/core-api/src/main/java/com/enonic/xp/node/NodeHits.java
```diff
@@ -34,6 +35,14 @@ public NodeIds getNodeIds()
         return NodeIds.from( this.stream().map( NodeHit::getNodeId ).collect( Collectors.toList() ) );
     }
 
+    public NodePaths getNodePaths()
+    {
+        return NodePaths.from( this.stream().
+            map( NodeHit::getNodePath ).
+            filter( Objects::nonNull ).
```
rymsha, line 42: I would prefer NPE here instead, because it is incorrect to not request paths, and ask for them later

rymsha (approved):

anatol-sialitski (approved):

modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/search/NodeSearchServiceImpl.java
```diff
@@ -51,7 +52,10 @@ private SearchResult doQuery( final NodeQuery query, final ReturnFields returnFi
         final SearchRequest searchRequest = SearchRequest.create().
             searchSource( source ).
             query( query ).
-            returnFields( returnFields ).
+            returnFields( query.isWithPath() ? ReturnFields.create().
```
rymsha, line 55: I don't currently imagine any other fields but `path` to be requested. Maybe we should deprecate/remove the method and move building of `ReturnFields` higher up

rymsha (changes requested):

modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/search/NodeSearchServiceImpl.java
rymsha: there is also a request for REFERENCE filed. we need to keep it as it is for now

rymsha (approved):

Issue #8608, opened 2021-01-11
importPermissions param is not used in ImportNodeCommand


vbradnitski:
`importPermissions` param used only in `updateNode` operation, not in `createNode`. Should be fixed to be able to skip source permissions.

PR #8611, opened 2021-01-12
importPermissions param is not used in ImportNodeCommand #8608

vbradnitski:


modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/CreatedEventSyncCommand.java
```diff
@@ -96,6 +96,7 @@ private ImportContentParams createImportParams( final ContentEventSyncCommandPar
             inherit( inheritTypes ).
             originProject( ProjectName.from( params.getSourceContext().getRepositoryId() ) ).
             importPermissions( false ).
```
rymsha, line 98: This is confusing (and also default)

rymsha (approved):

anatol-sialitski (approved):

PR #8615, opened 2021-01-14
importPermissions param is not used in ImportNodeCommand #8608

vbradnitski:


modules/core/core-repo/src/test/java/com/enonic/xp/repo/impl/node/ImportNodeCommandTest.java
```diff
@@ -129,7 +130,7 @@ public void permissions_on_root_create()
             repositoryId( RepositoryId.from( "test" ) ).
             build() );
 
-        final ImportNodeResult importNodeResult = ContextBuilder.from( CTX_OTHER ).
+        final ImportNodeResult importNodeResult = ContextBuilder.from( ContextAccessor.current() ).
```
rymsha, line 133: there is a method ctxOther() in parent class

rymsha (changes requested):

rymsha (approved):

Issue #8559, opened 2020-12-15
Deleted content may still remain published 


rymsha:
How to reproduce:
create folder `/a`
create folder `/a/b`
publish `/a` and `/a/b`
create folder `/c` (it must be in NEW state - that's the trick)
move `/a/b` to `/c`
delete `/c`

Result:
node `/a/b` is still be in master

Expected result
node `/a/b` does not exist in master

alansemenov:
![image](https://user-images.githubusercontent.com/11612490/102490364-02f39f00-406f-11eb-8c47-27bc3ac419b9.png)
![image](https://user-images.githubusercontent.com/11612490/102490384-0c7d0700-406f-11eb-8aed-eb964287e5f5.png)
![image](https://user-images.githubusercontent.com/11612490/102490612-5b2aa100-406f-11eb-96ee-a22f4176005f.png)

rymsha:
such case in code has its own description
> // Root node is new, just delete all children

That was node to never put "new" nodes into "pending delete" state and always delete "new" nodes in-place. Unfortunately the assumption that "new" node never has published descendants is wrong.

The solution would be to forcibly delete descendant nodes in both master and draft branch when when deletion is requested for a "new" node.

alansemenov:
Correction: it's not just _new_, but _unpublished_ as well (node that has been published, and then unpublished). The same rules should apply both for _new_ and _unpublished_ items.

rymsha:
The proposed solution appeared to be problematic as it easily creates detached branches in content tree. 

Imagine 
```
master
a -> b -> c

draft (b and moved)
a 
b
c
```

now delete `b`

with proposed solution it would result in 
```
master
a    c

draft
a 
c
```

so, `c` is not part of main tree in master

rymsha:
Next option is to collect nodeIds of deleted nodes in draft, and delete subtrees in master starting from nodes with returned nodeIds.

One thing to note:
delete operation is "unrecoverable" - it removes entire history of deleted node, and should be avoided for content that used to be published.

PR #8617, opened 2021-01-15
Deleted content may still remain published #8559

vbradnitski:


modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/json/content/ResolveForDeleteJson.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class ResolveForDeleteJson
+{
+    private final Set<ContentId> toDelete = new HashSet<>();
+
+    private final Set<ContentId> toUnpublish = new HashSet<>();
+
+    public ResolveForDeleteJson( final ContentIds toDelete, final ContentIds toUnpublish )
+    {
+        this.toDelete.addAll( toDelete.getSet() );
```
rymsha, line 19: Don't copy, just assign since sets are immutable

modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/json/content/ResolveForDeleteJson.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.enonic.xp.content.ContentIds;
+
+public class ResolveForDeleteJson
+{
+    private final ContentIds toDelete;
+
+    private final ContentIds toUnpublish;
+
+    public ResolveForDeleteJson( final ContentIds toDelete, final ContentIds toUnpublish )
```
anatol-sialitski, line 14: I prefer to use `List<ContentId>` or `List<ContentIdJson>` instead of `ContentIds`

anatol-sialitski (approved):

Issue #8117, opened 2020-05-27
Improve assets cache


alansemenov:
- When XP is running in dev mode assetUrl should not be cached. Related issue: https://discuss.enonic.com/t/asseturl-should-not-generate-cache-headers-when-in-dev-mode/2209/7

- in prod mode only assets with correct fingerprint and path should be cached

sigdestad:
Simply drop the cache header (or perhaps Even force no caching)

poi33:
This is a built in feature on all browsers, disable-caching of assets. 
Even if we remove the caching headers the browser can still force caching. 
I cant find any development environment that forces the browser to not cache.

I suggest to add documentation that the developer needs to disable the cache in the browser.

sigdestad:
Disabling browser cache can be done easily via http cache headers

rymsha:
Unfortunately, not @sigdestad Browsers are "smart" and like to cache anyway. https://blog.chromium.org/2017/01/reload-reloaded-faster-and-leaner-page_26.html

I am also not sure developers will be happy to see noticeable performance drop for page loading (not cached .js and .css files must be always parsed and executed). Even when development mode tis turned on "because it is always turned on on developer's machine". It is certainly a trade-off: one bad experience against another bad experience.

None of frameworks I know change `cache-control` headers for "static" files in development mode. Because disabling cache in browser is a way more flexible approach. 
- It allows to check that custom cache-control settings (set in Headers-Filter for instance) work as expected regardless of mode. 
-  It helps developers to see headers sent from XP as they would be in production.

sigdestad:
To see how work in production, the developer can simply run XP without dev mode :-)

rymsha:
> To see how work in production, the developer can simply run XP without dev mode :-)

But I don't see how it is better than toggle checkbox in the browser. IMO toggle checkbox in browser is simpler as developer does not need to stop XP instance, change mode and start instance again.

rymsha:
To sum up
- it always (often unnecessarily) slows down page loading/rendering in dev mode
- it modifies XP behavior between dev and prod mode
- it does not always work
- it requires a bit extra code testing and support
- it has a robust alternative

Better to document dev mode in our reference and mention browser checkbox there enonic/doc-xp#148

sigdestad:
Hmm... Are you saying devs are familiar with disabling cache? I never heard about this before now?

poi33:
Hard refreshing is one of the first thing i teach to new developers. 
[...]

PR #8618, opened 2021-01-15
Improve assets cache #8117

rymsha:
In PROD mode `/assets/` folder recursive hash is used for fingerprint: `/_/asset/com.enonic.web.enonic.com:f78f656d86dbeab8/img/favicon.ico`. 
Before this PR fingerprint was a timestamp when application was started: `/_/asset/com.enonic.web.enonic.com:1611141546/img/favicon.ico`. That has a glitch in clusters. Different nodes may have slightly different app startup time.

In DEV mode hash is not used at all. Instead fingerprint is a system time in milliseconds. But in order to be slightly more fancy system time is stored once in context when web request is initiated. So all assets urls on one html page share same fingerprint, as before. After this PR though, developers may enjoy cache invalidation for their static resources.

ResourceKey got new builder method to bypass `normalizePath` for well-known paths (currently only `/assets/` is known, but it is easy to add more) -this speed things up a little adds extra guarantees and simplifies code.

ResourceService got new method `resourceHash`. But actually allows only keys with strictly `/assets/` path - to avoid cache abuse.  There were 3 alternatives:
- new method `resourceHash` - easiest to implement with caching 
- `hash` method in `Resource` - Looked most accurate, but not that easy to cache hash results and would require to extend Resource to support folders.
- use `processResource` - Would not require API changes, but strictly speaking this method is an overcomplicated hack. Its cache is unbounded. Cache cleanup has a race with Application reinstall. It requires use of resource timestamp, that is hard to archive for folders.

New application instances are now pre-created for RESOLVED bundles. Before they were created by `ApplicationRegistryImpl`. That allows to access application bundle resources directly without need to use ApplicationService. Huge potential for code simplification in future. 
For now decouples ResourceService from ApplicationService - that allowed to cache hash directly in `ApplicationUrlResolver` so when application bundle gets uninstalled, cache gets removed together with it - it is not associated with applicationKey but rather with application bundle instead.

Hash is collision-resistant and very fast. It can be used in future to generate distinct resource fingerprints and ETags. BTW we can change implementation of it, it does not to be always the same between different versions of XP.


modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/AssetUrlBuilder.java
```diff
@@ -19,26 +24,23 @@ protected void buildUrl( final StringBuilder url, final Multimap<String, String>
...
             resolve();
 
-        final Application application = this.applicationService.getInstalledApplication( applicationKey );
-        if ( application == null )
-        {
-            throw new IllegalArgumentException( "Could not find application [" + applicationKey + "]" );
-        }
-        return application;
+        final String assetsHash = this.resourceService.resourcesHash( ResourceKey.assets( applicationKey ) ).map( HashCode::asString ).
+            orElseThrow( () -> new IllegalStateException( "Could not find application [" + applicationKey + "]" ) );
```
rymsha, line 33:
```suggestion
            orElseThrow( () -> new IllegalS
            ArgumentException( "Could not find application [" + applicationKey + "]" ) );
```

modules/core/core-internal/src/main/java/com/enonic/xp/core/internal/HexCoder.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public static String toHex( final long value )
+    {
+        final char[] buffer = new char[Long.BYTES * 2];
+        for ( int i = 0; i < Long.BYTES; i++ )
+        {
+            final byte v = (byte) ( value >>> ( Long.SIZE - ( i + 1 ) * Byte.SIZE ) );
+            buffer[2 * i] = HEX[( v >>> 4 ) & 0x0f];
+            buffer[2 * i + 1] = HEX[v & 0x0f];
+
```
rymsha, line 32:
```suggestion
```

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationFactoryServiceImpl.java
```diff
@@ -0,0 +1,94 @@
...
+            else
+            {
+                return null;
+            }
+        }
+
+        @Override
+        public void modifiedBundle( final Bundle bundle, final BundleEvent event, final ApplicationImpl object )
+        {
+
```
rymsha, line 86:
```suggestion
```

modules/core/core-api/src/main/java/com/enonic/xp/context/LocalScopeImpl.java
```diff
@@ -16,7 +16,7 @@
 
     LocalScopeImpl()
     {
-        this.attributes = new HashMap<>();
+        this.attributes = new HashMap<>( Map.of( "__currentTimeMillis", System.currentTimeMillis() ) );
```
rymsha, line 19: Move to web context?

modules/core/core-api/src/main/java/com/enonic/xp/util/HashCode.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        @Override
+        public String asString()
+        {
+            return HexCoder.toHex( value );
+        }
+
+        @Override
+        public boolean equals( final Object o )
```
rymsha, line 40: Is equality needed?

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationImpl.java
```diff
@@ -20,8 +19,8 @@
 import static com.enonic.xp.core.impl.app.ApplicationHelper.X_VENDOR_NAME;
 import static com.enonic.xp.core.impl.app.ApplicationHelper.X_VENDOR_URL;
 
-final class ApplicationImpl
-    implements Application
+public final class ApplicationImpl
```
rymsha, line 22: Back to package local?

modules/core/core-internal/src/main/java/com/enonic/xp/core/internal/HexCoder.java
```diff
@@ -0,0 +1,45 @@
...
+    public static String toHex( final byte[] value )
+    {
+        final char[] buffer = new char[value.length * 2];
+        for ( int i = 0; i < value.length; i++ )
+        {
+            final byte v = value[i];
+            buffer[2 * i] = HEX[( v >>> 4 ) & 0x0f];
+            buffer[2 * i + 1] = HEX[v & 0x0f];
+        }
+
```
rymsha, line 20:
```suggestion
```

modules/core/core-api/src/main/java/com/enonic/xp/context/LocalScopeImpl.java
rymsha: does not really matter now. And this way it is way simpler to manage

modules/core/core-api/src/main/java/com/enonic/xp/util/HashCode.java
rymsha: Since it is now private and only implementation - there is some freedom in how it is implemented.
And generally good to have (to be able to collect into HashSet)

anatol-sialitski (approved):

vbradnitski (approved):

Issue #8647, opened 2021-02-03
Image Service validate parameters


rymsha:
- [x] scale: eager validate name, dimensions
- [x] filter: parameters validate syntax and max total (out of scope: name, arguments)
- [x] format: allow only web formats supported by java
- [x] format: same behavior for rest and image service
- [x] background color
- [x] quality
- [x] correct gif and svgz handling

PR #8649, opened 2021-02-08
Image Service validate parameters #8647

rymsha:


modules/core/core-image/src/main/java/com/enonic/xp/core/impl/image/FilterCommand.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright 2000-2011 Enonic AS
```
anatol-sialitski, line 2: 2011?

modules/core/core-image/src/main/java/com/enonic/xp/core/impl/image/effect/ImageScales.java
```diff
@@ -0,0 +1,103 @@
...
+    private final int maxSideLength;
+
+    public ImageScales( final int maxSideLength )
+    {
+        this.maxSideLength = maxSideLength;
+    }
+
+    public ImageFunction full( FocalPoint focalPoint, Object... args )
+    {
+        Preconditions.checkArgument( args.length == 0, "Too many arguments %s", args.length );
```
anatol-sialitski, line 20: @rymsha error message can confused a user. Please update messages also for rest of the checks arguments.

rymsha: Message is
Too many arguments: 1

Only 0 arguments is allowed in this function

vbradnitski (approved):

anatol-sialitski (approved):

Issue #8657, opened 2021-02-17
Make html-escaping inside embed macro required


rymsha:
Use of embed macro without encoded html inside is needless - one who can insert unencoded version into macro could instead insert iframe without macro alltogether.

PR #8658, opened 2021-02-17
Make html-escaping inside embed macro required #8657

rymsha:


modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/macro/EmbedIframeMacroProcessor.java
```diff
@@ -24,37 +22,22 @@ public String getName()
...
-        return IFRAME_PATTERN.matcher( body ).matches();
+        return PortalResponse.create().contentType( MediaType.HTML_UTF_8 ).body( body ).build();
     }
 
-    private RenderMode getRenderingMode( final MacroContext macroContext )
+    private static boolean isIframeHtml( final String body )
     {
-        return macroContext.getRequest() == null ? RenderMode.LIVE : macroContext.getRequest().getMode();
+        // embed tag should be always escaped
+        return body.startsWith( "&lt;iframe" ) && body.endsWith( "&lt;/iframe&gt;" );
```
anatol-sialitski, line 41: body can be null. 
1. https://github.com/enonic/xp/blob/master/modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/postprocess/instruction/MacroInstruction.java#L190
2. https://github.com/enonic/xp/blob/master/modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/postprocess/instruction/Instruction.java#L32

rymsha: then `IFRAME_PATTERN.matcher( body ).matches();` would not work either

rymsha: 👍🏻

anatol-sialitski (approved):

vbradnitski (approved):

Issue #8679, opened 2021-02-25
Validate file names


rymsha:
Validate file names of exports, dumps and content attachments

PR #8683, opened 2021-03-01
Validate file names #8679

rymsha:


modules/core/core-internal/src/main/java/com/enonic/xp/core/internal/FileNames.java
```diff
@@ -0,0 +1,100 @@
...
+    // Sorted for binary search
+    public static final int[] INVISIBLE_CHARACTER_TYPES =
+        IntStream.of( Character.UNASSIGNED, Character.SPACE_SEPARATOR, Character.LINE_SEPARATOR, Character.PARAGRAPH_SEPARATOR,
+                      Character.CONTROL, Character.FORMAT, Character.PRIVATE_USE, Character.SURROGATE ).sorted().toArray();
+
+    private FileNames()
+    {
+    }
+
+    public static boolean isSafeFileName( final String fileName )
```
vbradnitski, line 36: Shouldn't it return `false` for null?

vbradnitski (approved):

modules/core/core-export/src/main/java/com/enonic/xp/core/impl/export/writer/NodeExportPathResolver.java
```diff
@@ -1,14 +1,11 @@
 package com.enonic.xp.core.impl.export.writer;
 
-import java.nio.file.Path;
-import java.nio.file.Paths;
-
-import com.enonic.xp.node.NodePath;
-import com.enonic.xp.node.NodeVersionId;
-import com.enonic.xp.util.BinaryReference;
-
 public class NodeExportPathResolver
```
anatol-sialitski, line 3: Maybe make sense to rename this file, for instance to NodeExportPathConstants, because no any resolve method

modules/core/core-internal/src/main/java/com/enonic/xp/core/internal/FileNames.java
rymsha: null is not unsafe name. there is no name. I change code to throw NPE more explicitly

modules/core/core-export/src/main/java/com/enonic/xp/core/impl/export/writer/NodeExportPathResolver.java
rymsha: Will be part of some other cleanup. This PR is already overloaded

anatol-sialitski (approved):

Issue #8599, opened 2021-01-07
Scheduler Library


rymsha:
- create
- modify
- list
- get
- delete

Pure wrapper of ShedulerService

PR #8701, opened 2021-03-15
Scheduler Library #8599

vbradnitski:


modules/itest/itest-core/src/test/java/com/enonic/xp/core/scheduler/SchedulerServiceImplTest.java
```diff
@@ -426,8 +426,8 @@ void get()
 
         assertEquals( name, scheduledJob.getName() );
         assertEquals( descriptor, scheduledJob.getDescriptor() );
-        assertEquals( calendar.getCronValue(), ( (CronCalendar) scheduledJob.getCalendar() ).getCronValue() );
-        assertEquals( calendar.getTimeZone(), ( (CronCalendar) scheduledJob.getCalendar() ).getTimeZone() );
+        assertEquals( calendar.getCronValue(), ( (CronCalendarImpl) scheduledJob.getCalendar() ).getCronValue() );
```
rymsha, line 429: why would one need to cast?

vbradnitski: ScheduledJob.getCalendar() returns ScheduleCalendar which is too common to fetch fields.

modules/core/core-api/src/main/java/com/enonic/xp/scheduler/OneTimeCalendar.java
```diff
@@ -0,0 +1,11 @@
+package com.enonic.xp.scheduler;
+
+import com.enonic.xp.annotation.PublicApi;
+
+@PublicApi
+public interface OneTimeCalendar
+    extends ScheduleCalendar
+{
+    String getValue();
```
rymsha, line 9: Should be Instant, probably?

modules/core/core-api/src/main/java/com/enonic/xp/scheduler/CronCalendar.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.util.TimeZone;
+
+import com.enonic.xp.annotation.PublicApi;
+
+@PublicApi
+public interface CronCalendar
+    extends ScheduleCalendar
+{
+    String getDescription();
```
rymsha, line 11: It may be handy but do we really need it?

modules/core/core-scheduler/src/main/java/com/enonic/xp/impl/scheduler/distributed/CronCalendarImpl.java
```diff
@@ -17,11 +17,11 @@
 import com.cronutils.parser.CronParser;
 import com.google.common.base.Preconditions;
 
-import com.enonic.xp.scheduler.ScheduleCalendar;
+import com.enonic.xp.scheduler.CronCalendar;
 import com.enonic.xp.scheduler.ScheduleCalendarType;
 
-public final class CronCalendar
-    implements ScheduleCalendar
+public final class CronCalendarImpl
```
rymsha, line 23: to avoid confusion such Impl classes should be package local.
They should be other OSGi classloaders but nothing but this module should be able to instantiate them.

rymsha (approved):

Issue #8600, opened 2021-01-07
Scheduler Management endpoint


rymsha:
List schedules with their data.
Hopefully we can show next run time as well

PR #8708, opened 2021-03-17
Scheduler Management endpoint #8600

vbradnitski:


modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SchedulerResource.java
```diff
@@ -0,0 +1,45 @@
...
+    public List<ScheduledJobJson> list()
+    {
+        return schedulerService.list().
+            stream().
+            map( ScheduledJobJson::new ).
+            collect( Collectors.toList() );
+
+    }
+
+    @Reference
```
rymsha, line 40: Use constructor injection

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/model/ScheduledJobJson.java
```diff
@@ -0,0 +1,75 @@
...
+import com.enonic.xp.scheduler.ScheduleCalendar;
+import com.enonic.xp.scheduler.ScheduledJob;
+
+public class ScheduledJobJson
+{
+    private final ScheduledJob job;
+
+    public ScheduledJobJson( final ScheduledJob job )
+    {
+        this.job = job;
```
rymsha, line 18: Copying fields in ctor would look better

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/model/ScheduledJobJson.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    public Instant getLastRun()
+    {
+        return lastRun;
+    }
+
+    public Instant getNextExecution()
+    {
+        return enabled ? calendar.
```
rymsha, line 93: move to ShedulerCalendar

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/model/ScheduledJobJson.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    public boolean isEnabled()
+    {
+        return enabled;
+    }
+
+    public String getDescriptor()
+    {
+        return descriptor.toString();
```
rymsha, line 68: do not `toString` objects that are String-like

rymsha (changes requested):

rymsha (approved):

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/SchedulerResource.java
```diff
@@ -0,0 +1,46 @@
...
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
+
+import com.enonic.xp.impl.server.rest.model.ScheduledJobJson;
+import com.enonic.xp.jaxrs.JaxRsComponent;
+import com.enonic.xp.scheduler.SchedulerService;
+import com.enonic.xp.security.RoleKeys;
+
+@Path("/scheduler")
```
anatol-sialitski, line 21: maybe rename to `schedulers`, the GET method list will be available by default and that is right naming for REST

vbradnitski: but then we have to rename all exist resources

Issue #6794, opened 2018-12-03
missing deny for rootpermissions should be interpreted as empty array


ComLock:
repo.create fails if rootpermissions doesn't have deny:[] with Empty message in exception

rymsha:
I don't think it is a bug. Inconvenience - more likely.

PR #8710, opened 2021-03-19
repo.create fails if permissions doesn't have deny:[] #6794

anatol-sialitski:


rymsha (changes requested):
That would also make `allow` optional. Is it intentional?

anatol-sialitski:
> That would also make `allow` optional. Is it intentional?

@rymsha Yes, the previous implementation supports creation a repo with empty `allow` permission. But, if skip `allow` in params, then will be the same error as with `deny`. So I use the resolvePermissions for both fields.

rymsha (approved):

vbradnitski (approved):

Issue #8704, opened 2021-03-17
free-config to contentType and component schemas


rymsha:
contentType and component (layout, part, page) schemas model should be extended with config tag
config tag accepts free set of tags. Works same way as config for InputType

Known to be used for the following configuration parameters
`allow-new-content` #8720
`list-tilte-expression` #8713

Example:
```
<content-type>
  <config>
    <allow-new-content>true</allow-new-content>
    <list-display-expression>${expr}</list-display-expression>
  </config>
</content-type>
```

PR #8744, opened 2021-04-12
Add free config to contentType and component schemas #8704

rymsha:


vbradnitski (approved):

modules/core/core-api/src/main/java/com/enonic/xp/region/ComponentDescriptor.java
```diff
@@ -32,6 +35,7 @@ protected ComponentDescriptor( final BaseBuilder builder )
         this.description = builder.description;
         this.descriptionI18nKey = builder.descriptionI18nKey;
         this.config = builder.config;
+        this.schemaConfig = builder.schemaConfig.build();
```
anatol-sialitski, line 38: What about check on null?

anatol-sialitski (approved):

Issue #8755, opened 2021-04-16
Scheduled Job improvements 


vbradnitski:
- rename 'payload' to 'config'. 
- rename 'calendar' to 'schedule' in lib.
- remove 'author' field.
- add createdBy and 'modifiedBy' fields.

PR #8757, opened 2021-04-19
Scheduled Job improvements #8755

vbradnitski:


modules/lib/lib-scheduler/src/main/resources/lib/xp/scheduler.js
```diff
@@ -35,22 +35,20 @@ function nullOrValue(value) {
  * @param {string} params.name unique job name.
  * @param {string} params.description job description.
  * @param {string} params.descriptor descriptor of the task to be scheduled.
- * @param {object} params.payload config of the task to be scheduled.
- * @param {object} params.calendar task time run config.
+ * @param {object} params.config config of the task to be scheduled.
+ * @param {object} params.schedule task time run config.
```
rymsha, line 39: schedule object need docs as well

rymsha (changes requested):

Issue #8737, opened 2021-04-08
Default values are not correctly set for inputs with several occurrences on content create


alansemenov:
This content type has a TextLine input with a default value and min 3 occurrences (which means they should be created automatically).

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<content-type>
  <display-name>Test</display-name>
  <super-type>base:structured</super-type>
  <form>
    <input name="field1" type="TextLine">
      <label>Field 1</label>
      <occurrences minimum="3" maximum="3"/>
      <default>Default text</default>
    </input>
  </form>
</content-type>
```
In Content Studio it's rendered correctly (correct number of fields with the default value):

![image](https://user-images.githubusercontent.com/11612490/114048750-209d7c00-988b-11eb-9dea-227e963637f1.png)

But the content is created with wrong data (only one occurrence of `field1`):

```json
  "data": {
    "field1": "Default text"
  },
```

After first save `data` becomes correct:

```json
  "data": {
    "field1": [
      "Default text",
      "Default text",
      "Default text"
    ]
  },
```

This is also a problem with container inputs (like item-sets) that have several occurrences.

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<content-type>
  <display-name>ItemSet</display-name>
  <super-type>base:structured</super-type>
  <form>
    <input name="field1" type="TextLine">
      <label>Field 1</label>
      <occurrences minimum="1" maximum="1"/>
      <default>Default text</default>
    </input>
    <item-set name="field2">
      <label>Field 2</label>
      <items>
        <input name="label" type="TextLine">
          <label>Label</label>
          <occurrences minimum="1" maximum="1"/>
          <default>Default text</default>
        </input>
      </items>
      <occurrences minimum="3" maximum="3"/>
    </item-set>
  </form>
</content-type>
```

Content Studio:

![image](https://user-images.githubusercontent.com/11612490/114045148-19c13a00-9888-11eb-8b0e-46c2f8d7ef3a.png)

Wrong `data` on create (only one occurrence of `field2`):
```json
  "data": {
    "field1": "Default text",
    "field2": {
      "label": "Default text"
    }
  },
```
Correct `data` after save:
```json
  "data": {
    "field1": "Default text",
    "field2": [
      {
        "label": "Default text"
      },
      {
        "label": "Default text"
      },
      {
        "label": "Default text"
      }
    ]
  },
```

PR #8762, opened 2021-04-22
Default values are not correctly set for inputs with several occurrences on content create #8737

anatol-sialitski:


modules/core/core-schema/src/main/java/com/enonic/xp/core/impl/form/FormDefaultValuesProcessorImpl.java
```diff
@@ -32,10 +36,11 @@
     @Override
     public void setDefaultValues( final Form form, final PropertyTree data )
     {
-        processFormItems( form.getFormItems(), data, PropertyPath.from( "" ) );
+        processFormItems( form.getFormItems(), data, PropertyPath.from( "" ), null );
```
rymsha, line 39: maybe instead of null provide normalize it to "default" value `Occurrences.default()` or smth

rymsha (approved):

vbradnitski (approved):

Issue #8666, opened 2021-02-19
getVersionsForView is missing a version created after sorting changes


alansemenov:
1. Create a new content
2. Publish it
3. Change sorting. Content state changes to "_Modified_".
4. Check the Version History widget

The last version shown there is the one that was published, but the one created when the sorting was changed is missing (in API response there's only one version with `publishInfo`).

![image](https://user-images.githubusercontent.com/11612490/108495996-caa85100-72a9-11eb-8735-880a30e68a5b.png)

![image](https://user-images.githubusercontent.com/11612490/108495858-97fe5880-72a9-11eb-8cd4-8fbeb47aa3c1.png)

PR #8779, opened 2021-04-27
getVersionsForView is missing a version created after sorting changes…

vbradnitski:
… #8666


anatol-sialitski (approved):

rymsha (changes requested):
Use equalsverifier

rymsha (approved):

Issue #8786, opened 2021-04-28
Scheduler improvements


vbradnitski:
- rename 'init_job' to 'init-job' prefix.
 - set `Anonymous` as a default user.
 - fix `last_run` to be an Instant.
 - fix 'tasks' param description for vacuum.
 - add users for default system tasks.
 - change user format to be `<idProvider>:<name>`

PR #8787, opened 2021-04-28
Scheduler improvements #8786

vbradnitski:


anatol-sialitski (approved):

modules/core/core-scheduler/src/main/java/com/enonic/xp/impl/scheduler/distributed/SchedulableTaskImpl.java
```diff
@@ -93,26 +91,32 @@ public void run()
...
     private Context taskContext()
     {
-        Context taskContext = ContextAccessor.current();
-
         if ( job.getUser() == null )
         {
-            return taskContext;
+            return ContextBuilder.from( ContextAccessor.current() ).
+                authInfo( AuthenticationInfo.create().
+                    principals( RoleKeys.AUTHENTICATED, RoleKeys.EVERYONE ).
```
rymsha, line 98: Why would it be authenticated?

modules/core/core-scheduler/src/main/java/com/enonic/xp/impl/scheduler/distributed/SchedulableTaskImpl.java
```diff
@@ -93,26 +91,32 @@ public void run()
...
+
+        final AuthenticationInfo authInfo = OsgiSupport.withService( SecurityService.class, securityService -> {
+            final VerifiedUsernameAuthToken token = new VerifiedUsernameAuthToken();
+            token.setIdProvider( job.getUser().getIdProviderKey() );
+            token.setUsername( job.getUser().getId() );
+
+            return securityService.authenticate( token );
+        } );
+
+        if ( !authInfo.isAuthenticated() )
```
rymsha, line 112: Check only if user is specified

rymsha (changes requested):

modules/core/core-scheduler/src/main/java/com/enonic/xp/impl/scheduler/distributed/SchedulableTaskImpl.java
vbradnitski: There is a check "if ( job.getUser() == null )" above

rymsha: second thought: this is not Scheduler responsibility to verify identity, remove this check altogether (but keep `securityService.authenticate( token );`

rymsha (approved):

Issue #8733, opened 2021-03-12
Templates folder is deleted in a layer after deleting the site in the parent project


sgauruseu:
1. Create new site in SuperHero context - select SuperHero in applications, select `Page` in controllers, insert image components
![image](https://user-images.githubusercontent.com/3728712/110939164-0a60d680-8346-11eb-9ead-7135e5934d9a.png)

2. Create new layer in SuperHero project,
3.  Switch to the layer's context and localize inherited site and child images. Do not localize `_templates` folder
![image](https://user-images.githubusercontent.com/3728712/110933501-e9948300-833d-11eb-943b-1ce098057c69.png)

4. Switch to parent context and delete `site1`
5. Switch to layer again 
**Issue 1**   `_templates`   folder is deleted in child layer
![image](https://user-images.githubusercontent.com/3728712/110933720-2fe9e200-833e-11eb-9bff-794a6373d65b.png)

**Issue 2** 
1) Switch to layer. Open the site  and open child images and reset these items . 
2) Select  child image and delete it 
Issue - the site is automatically deleted after removing the child component

![image](https://user-images.githubusercontent.com/3728712/110935560-a38cee80-8340-11eb-9129-559267205265.png)

Whether the content should be deleted after deleting this content in the parent project and clicking on the `Reset` button?

vbradnitski:
issue 2 is not a bug, because image has an inboud dependency from site.

sigdestad:
Are we considering implemeting some special  handling for template folders? Do we need that?

rymsha:
template-folder is already "special". It is the only content that gets created "automagically" together with site, that layers sync process - we had to implement new `content.import` method in order to bypass auto-creation of template.

This bug is very similar to #6844

Proposed fix in PR creates other problems. Back to discussion...

PR #8805, opened 2021-05-06
Templates folder is deleted in a layer after deleting the site in the…

vbradnitski:
… parent project #8733


rymsha (changes requested):
This would make template folder deletion to never propagate to child layer

Issue #8752, opened 2021-04-16
lib-task submitTask/executeFunction method rename


rymsha:
`submitTask` substitutes and deprecates `submitNamed`
`name` (in `submitNamed`)-> `descriptor` (in `submitTask`)

`submit` -> `executeFunction`

rymsha:
Consider `runFunction` instead of `runLocal`

sigdestad:
Will it run any "function", or only "tasks"?

sigdestad:
If it only runs tasks, and if it is part of task API, why not simply call it "run"?

vbradnitski:
We used to call it task, but it's actually just a js-function can be set there:
 `setTask( final Function<Void, Void> taskFunction )`

rymsha:
> Will it run any "function", or only "tasks"?

alansemenov:
@sigdestad this is becoming a blocker. can we agree on the name here?

sigdestad:
So, it will run any function, and it only runs on the local node?

submitTask() submits task for execution in cluster
Here are my suggestions for the Background task/function executor
Detach() 
Async() 
Execute() 
Run()

sigdestad:
Since this is already in the task lib, maybe we could name "submitTask()" simply submit() also?

sigdestad:
And go for run() for the local version?

rymsha:
> submitTask()" simply submit()

rymsha:
> submitTask()" simply submit()

Not possible. `submit` is already occupied with what we are trying to give a better name.

"Local/clustered" - words are confusing, as it is not always an XP cluster
"Detach" - "C++" term for threads, but means a different thing.
"Async" - will be sooo confusing for JS developers, because behavior is totally different.
"Run" - gives headache for Java developers, as they used to see `Thread.run` (and it differs from `Executor.execute`)
"Execute" - might be a good candidate, actually, but why would we call one `execute` and another `submitTask`...?

We also have to consider how Java API looks like (and hopefully we can make two APIs look more similar)

```
public interface TaskService
{
    TaskId submitTask( RunnableTask runnable, String description ); <--- in JS "submit" wraps it, but does a lot of conversion as well.

    TaskId submitTask( DescriptorKey key, PropertyTree config ); <--- in JS "submitNamed" (yey! we rename it to submitTask in JS)
```

This is why `submitFunction` more or less fits:
- it still starts with "submit" word - similar to "submitTask"
[...]

PR #8807, opened 2021-05-06
lib-task submitTask method #8752

vbradnitski:


rymsha:
emergency fix for the build

Issue #8816, opened 2021-05-12
Optional occurrence of a set is auto-created in a new content


alansemenov:
Most likely regression after https://github.com/enonic/xp/issues/8738.

Creating a new content with this schema will now **automatically** create an occurrence of the _optional_ item-set. It didn't use to be like that (tested on QA with 7.6).

We should probably add tests for this case.

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<content-type>
  <display-name>Item Set</display-name>
  <super-type>base:structured</super-type>
  <form>
    <item-set name="item-set">
      <label>Item Set</label>
      <items>
        <input name="myTextLine" type="TextLine">
          <label>My text line</label>
          <occurrences minimum="0" maximum="0"/>
        </input>
      </items>
      <occurrences minimum="0" maximum="0"/>
    </item-set>
  </form>
</content-type>
```

Content object right after creation (note non-empty `data` object):
```json
{
  "_id": "8a4fa6f2-8037-4f78-85ab-86997cc09519",
  "_name": "__unnamed__a41272f0-3731-49bc-95e3-ce3dd631e6ed",
  "_path": "/features/__unnamed__a41272f0-3731-49bc-95e3-ce3dd631e6ed",
  "creator": "user:system:su",
  "modifier": "user:system:su",
  "createdTime": "2021-05-12T07:49:22.711399Z",
  "modifiedTime": "2021-05-12T07:49:22.711399Z",
  "owner": "user:system:su",
  "type": "com.enonic.app.features:nested-form-item-sets",
  "displayName": "",
  "hasChildren": false,
  "valid": false,
  "childOrder": "modifiedtime DESC",
  "data": {
    "level-1": {}
  },
  "x": {},
  "page": {},
  "attachments": {},
  "publish": {}
}
```

Same on 7.6:
```json
{
  "_id": "62f5fcc2-5d83-4eb9-95f9-27a0d085886b",
  "_name": "__unnamed__49e6f071-2ee7-4296-8baf-3b698516f2b1",
  "_path": "/features/__unnamed__49e6f071-2ee7-4296-8baf-3b698516f2b1",
  "creator": "user:system:ase",
  "modifier": "user:system:ase",
  "createdTime": "2021-05-12T07:53:59.183251Z",
  "modifiedTime": "2021-05-12T07:53:59.183251Z",
  "owner": "user:system:ase",
  "type": "com.enonic.app.features:nested-form-item-sets",
  "displayName": "",
[...]

PR #8818, opened 2021-05-12
Optional occurrence of a set is auto-created in a new content #8816

anatol-sialitski:


rymsha (approved):

vbradnitski (approved):

rymsha (changes requested):
Patch coverage must be 100% in this case

Issue #8817, opened 2021-05-12
Task is missing connection to its job


vbradnitski:
add lastTaskId field to ScheduledJob
Should be set together with `latRun` field.

PR #8819, opened 2021-05-13
Task is missing connection to its job #8817

vbradnitski:


modules/core/core-api/src/main/java/com/enonic/xp/scheduler/ScheduledJob.java
```diff
@@ -36,6 +37,8 @@
 
     private final Instant lastRun;
 
+    public final TaskId lastTaskId;
```
rymsha, line 40:
```suggestion
    private final TaskId lastTaskId;
```

rymsha (changes requested):

modules/core/core-scheduler/src/main/java/com/enonic/xp/impl/scheduler/ScheduleAuditLogSupportImpl.java
```diff
@@ -154,6 +154,11 @@ private void addResult( final PropertySet targetSet, final ScheduledJob job )
             targetSet.setInstant( ScheduledJobPropertyNames.LAST_RUN, job.getLastRun() );
         }
 
+        if ( job.getLastTaskId() != null )
```
rymsha, line 157: why would we need it in audit log? (actually lastRun is also confusing)

modules/lib/lib-scheduler/src/main/java/com/enonic/xp/lib/scheduler/GetScheduledJobLastRunStateHandler.java
```diff
@@ -0,0 +1,33 @@
+package com.enonic.xp.lib.scheduler;
+
+import java.util.Optional;
+
+import com.google.common.base.Preconditions;
+
+import com.enonic.xp.lib.scheduler.mapper.ScheduledJobLastRunStateMapper;
+import com.enonic.xp.scheduler.ScheduledJobName;
+
+public final class GetScheduledJobLastRunStateHandler
```
rymsha, line 10: After a second thought I guess this makes js and java API asymmetrical.
either split methods in Java API or join methods in js

rymsha (changes requested):

modules/core/core-scheduler/src/main/java/com/enonic/xp/impl/scheduler/ScheduleAuditLogSupportImpl.java
vbradnitski: why not? It may be useful for somebody

modules/lib/lib-scheduler/src/test/java/com/enonic/xp/lib/scheduler/BaseScheduledJobHandlerTest.java
```diff
@@ -40,9 +41,11 @@
 
     protected PropertyTreeMarshallerService propertyTreeMarshallerService;
 
+    private Map<ScheduledJobName, ScheduledJob> jobs;
+
     private void mockJob()
     {
-        final Map<ScheduledJobName, ScheduledJob> jobs = Maps.newHashMap();
+        jobs = Maps.newHashMap();
```
rymsha, line 48:
```suggestion
        jobs = new HashMap<>()
```

rymsha (approved):

Issue #8838, opened 2021-05-20
Archive POC


rymsha:
new node `/archive`

new Content API Java methods:
`archive(contentId)` 
- move content under `/archive` node together with its subtree.
- rename content to random UUID.
- store original name in `archivedName` property.

`restore(contentId, parentPath)`
- move content under parentPath
- rename content to the value stored in `archivedName`
- remove `archivedName` property

new Admin API enpdoints with the same functionality

rymsha:
Instead of renaming content we decided to create "container" node for each "archive" transaction with random UUID as name. It holds data about archived nodes. Content nodes moved directly under a holder node without renaming.

PR #8877, opened 2021-06-04
Archive POC #8838

vbradnitski:


modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/content/ContentResource.java
```diff
@@ -421,6 +424,32 @@ public TaskResultJson move( final MoveContentJson params )
...
+            .createTaskResult();
+    }
+
+    @POST
+    @Path("restore")
+    public TaskResultJson restore( final RestoreContentJson params )
+    {
+        return RestoreRunnableTask.create()
+            .params( params )
+            .description( "Archive content" )
```
rymsha, line 446:
```suggestion
            .description( "Restore content" )
```

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/RestoreContentCommand.java
```diff
@@ -0,0 +1,134 @@
...
+        final Node contentToRestore = nodeService.getById( NodeId.from( params.getContentId() ) );
+        final Node container = nodeService.getByPath( contentToRestore.parentPath() );
+
+        final NodePath oldParentPath = params.getPath() != null
+            ? ContentNodeHelper.translateContentPathToNodePath( params.getPath() )
+            : NodePath.create( container.data().getString( "oldParentPath" ) ).build();
+
+        final MoveNodeParams.Builder builder = MoveNodeParams.create().
+            nodeId( NodeId.from( params.getContentId() ) ).
+            parentNodePath( oldParentPath ).
```
rymsha, line 78: When old parent is missing restore to root

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/RestoreContentCommand.java
```diff
@@ -0,0 +1,134 @@
...
+            moveListener( this );
+
+        final Node movedNode = nodeService.move( builder.build() );
+
+        // TODO: discuss
```
rymsha, line 83: I'm ok with deleting it, but only if it is guaranteed that move is successful.

modules/core/core-project/src/main/java/com/enonic/xp/core/impl/project/init/ArchiveInitializer.java
```diff
@@ -0,0 +1,84 @@
...
+            parent( NodePath.ROOT ).
+            permissions( Objects.requireNonNullElse( accessControlList, ArchiveConstants.ARCHIVE_ROOT_DEFAULT_ACL ) ).
+            childOrder( ArchiveConstants.DEFAULT_ARCHIVE_REPO_ROOT_ORDER ).
+            build() );
+
+        LOG.info( "Created archive root-node: " + archiveRoot.path() );
+
+        nodeService.refresh( RefreshMode.ALL );
+
+        nodeService.push( NodeIds.from( archiveRoot.id() ), ContentConstants.BRANCH_MASTER );
```
rymsha, line 72: We should somehow prohibit pushing from archive. One way of doing it is to NOT push archive to master. TBD

rymsha (approved):

Issue #8858, opened 2021-05-28
Trailing slash is removed from rawPath


espen42:
In at least one circumstance, neither `path` nor `contextPath` nor `rawPath` can be used to determine whether or not a call to a controller was done with a trailing slash or not - and from what I could see, not any other fields either. 

The circumstance: vhosting active (`localhost/wapp` -> `localhost/webapp/my.app` etc, in _com.enonic.xp.web.vhost.cfg_), and , following a link in a webapp (link made by portal.serviceUrl) to `localhost/wapp/_/service/my.app/myservice`. In the webapp, the trailing slashes were visible. But in the service, when logging the entire request, there was no difference between navigating to `localhost/wapp/_/service/my.app/myservice/stuff` and `localhost/wapp/_/service/my.app/myservice/stuff/`.

Notably it _only_ occurred when `enabled = true` was set in _com.enonic.xp.web.vhost.cfg_. If it was commented out or set to false, then trailing slashes were visible - _but vhosting was still working_, for some reason. Unexpected to me at least)

Not sure if this is a bug, or necessary behavior for something else to work. But this looks like a **blocking problem for lib-static**, both since detecting trailing slashes is a crucial part of falling back to an index.html, and since leaning on the request object for fallback creates recursive redirect loops whenever using lib-static below a service and calling it (or even a sub-uri) without a trailing slash. 

In what other circumstances might this be an issue? Does this need to be mapped out completely for knowing when lib-static can be trusted to detect this (or, can be used at all - which won't look very good IMO)? If so, verifying all the different kinds of controllers (service, webapp, mapped controller, component controller, service URIs created by `serviceUrl` in all of those), orthogonally combined with [no vhosting] vs [vhosting with enabled = true] vs [vhosting with enabled = false], maybe even [vhosting with enabled commented out]?

alansemenov:
Duplicate of https://github.com/enonic/xp/issues/8848?

espen42:
@alansemenov IMO, no, this is bigger (and worse?). Since it applies to _all_ of the fields in `request` and not only `path`, there's no opening for a workaround in lib-static.

rymsha:
related #7087

rymsha:
The issue happens because when vhosts are enabled "VirtualHostFilter" rebuilds path from "source" to "target" path. But the way it is done also strips away all extra slashes and whitespaces - and as we see, they are sometimes essential. The fix is to be less aggressive in path "normalization": 

before the fix:
[...]

PR #8885, opened 2021-06-14
Trailing slash in request URI is sometimes completely hidden from req…

vbradnitski:
…uest? #8858

### DIFF
--- a/modules/web/web-vhost/src/test/java/com/enonic/xp/web/vhost/impl/VirtualHostInternalHelperTest.java
+++ b/modules/web/web-vhost/src/test/java/com/enonic/xp/web/vhost/impl/VirtualHostInternalHelperTest.java
@@ -30,4 +30,20 @@ public class VirtualHostInternalHelperTest
         assertEquals( fullTarget, "/b/c/other/service" );
     }
 
+    @Test
+    public void testGetFullTargetPathWithTrailingSlash()
+    {
+        final VirtualHostMapping virtualHostMapping = new VirtualHostMapping( "host" );
+
+        virtualHostMapping.setHost( "foo.no" );
+        virtualHostMapping.setSource( "/a" );
+        virtualHostMapping.setTarget( "/b/c" );
+
+        final HttpServletRequest req = mock( HttpServletRequest.class );
+        when( req.getRequestURI() ).thenReturn( "/a/other/service/" );
+
+        final String fullTarget = VirtualHostInternalHelper.getFullTargetPath( virtualHostMapping, req );
+
+        assertEquals( fullTarget, "/b/c/other/service/" );
+    }
 }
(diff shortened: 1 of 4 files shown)


modules/web/web-vhost/src/test/java/com/enonic/xp/web/vhost/impl/VirtualHostInternalHelperTest.java
```diff
@@ -13,6 +13,19 @@
 public class VirtualHostInternalHelperTest
 {
 
+    @Test
+    public void testNormalizePath()
+    {
+        final HttpServletRequest req = mock( HttpServletRequest.class );
+        when( req.getRequestURI() ).thenReturn( "/a/other/service" );
+
+        assertEquals( "/", VirtualHostInternalHelper.normalizePath( null ) );
```
rymsha, line 22: verify if null safety is needed

modules/web/web-vhost/src/test/java/com/enonic/xp/web/vhost/impl/VirtualHostInternalHelperTest.java
```diff
@@ -13,6 +13,19 @@
...
+        assertEquals( "/", VirtualHostInternalHelper.normalizePath( "" ) );
```
rymsha, line 23: verify if we ever receive empty path

modules/web/web-vhost/src/test/java/com/enonic/xp/web/vhost/impl/VirtualHostInternalHelperTest.java
```diff
@@ -13,6 +13,19 @@
...
+        assertEquals( "/", VirtualHostInternalHelper.normalizePath( "/" ) );
+        assertEquals( "/a", VirtualHostInternalHelper.normalizePath( "/ a /" ) );
```
rymsha, line 25: that would be suspicious if path contains raw spaces. Maybe the idea was to trim? (because value comes from config)

modules/web/web-vhost/src/test/java/com/enonic/xp/web/vhost/impl/VirtualHostInternalHelperTest.java
```diff
@@ -13,6 +13,19 @@
...
+        assertEquals( "/a", VirtualHostInternalHelper.normalizePath( "/ a //" ) );
```
rymsha, line 26: Don't mix two asserts. Test that `//` is normalized. Maybe the idea is actually to subtract trailing `/` and the rest of normalization is not needed.
Also, find out, maybe sometimes it is better to preserve trailing `/` ?

modules/web/web-vhost/src/test/java/com/enonic/xp/web/vhost/impl/VirtualHostInternalHelperTest.java
vbradnitski: if `source` or `target` field is not set, then null is coming and we handle it like `/`

rymsha: But this work should be done in VirtualHostConfigMap, in order for Vhost API to work correctly

Issue #8839, opened 2021-05-20
Strange error from scheduler


jsi:
This may just be a small hiccup, but I thought I'd mention it:  If I start a clean XP system and then just stop it again, using the CLI, I get this error when I stop it:
```
2021-05-20 16:35:31,749 WARN  c.e.x.i.s.distributed.RescheduleTask - Problem during task scheduling.
java.lang.IllegalStateException: No service found interface com.enonic.xp.scheduler.SchedulerService
	at com.enonic.xp.core.internal.osgi.OsgiSupport.lambda$withService$3(OsgiSupport.java:97)
	at com.enonic.xp.core.internal.osgi.OsgiSupport.withServiceOrElseGet(OsgiSupport.java:59)
	at com.enonic.xp.core.internal.osgi.OsgiSupport.withService(OsgiSupport.java:96)
	at com.enonic.xp.impl.scheduler.distributed.RescheduleTask.doRun(RescheduleTask.java:52)
	at com.enonic.xp.impl.scheduler.distributed.RescheduleTask.run(RescheduleTask.java:42)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

rymsha:
It is not an error. It is a warning

vbradnitski:
There is also `warning` log level for CancelationException, should be changed to `debug`.

```
2021-05-31 17:21:00,004 WARN  c.e.x.i.s.d.SchedulableTaskImpl - Error while running job. java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@1a8576f0[Cancelled]
java.util.concurrent.CancellationException: null
        at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:121)
        at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
[...]

PR #8896, opened 2021-06-17
Strange error from scheduler #8839

vbradnitski:


modules/core/core-scheduler/src/main/java/com/enonic/xp/impl/scheduler/distributed/RescheduleTask.java
```diff
@@ -43,6 +45,14 @@ public void run()
         {
             this.doRun();
         }
+        catch ( IllegalStateException e )
+        {
+            if ( ++failedCount >= 10 )
+            {
+                failedCount = 0;
```
rymsha, line 52: successful submits should also reset to 0

rymsha (approved):

anatol-sialitski (approved):

Issue #8902, opened 2021-06-18
Import content from app resources does not work


rymsha:
```
14:48:06.835 INFO com.enonic.app.superhero - (/main.js) Project “sample-blog” successfully created
14:48:06.836 INFO com.enonic.app.superhero - (/main.js) Importing “sample-blog” data
14:48:06.856 INFO com.enonic.app.superhero - (/main.js) Error: null
```

rymsha:
regression after #8863

rymsha:
Problem is that exportName was still evaluated relative to exports folder even when source from app resources is specified.
And because exportName is always null in such case, NPE is thrown.

PR #8903, opened 2021-06-18
Import content from app resources does not work #8902

rymsha:


anatol-sialitski (approved):

rymsha:
Fix works. need to add tests

vbradnitski (approved):

Issue #8928, opened 2021-06-23
allowContentTypes regression


rymsha:
Enonic version: 7.7

With 7.7 and Content Studio 3.3 it seems the way regex for "allowContentType" of x-data is changed in a way that breaks current behavior.

These patterns all work fin before we upgraded. After upgrade these x-datas are not added at all. This breaking change is not documented, so I kind of assume this is unexpected behavior for you.

```
<x-data name="teasers" allowContentTypes="(link|link-application|page-(article|general))" optional="true" />
<x-data name="menu-item" allowContentTypes="(page-article|page-general|link|menu-heading|base:shortcut)" optional="true" />
```

We tried changing these to this but that also failed:

```
<x-data name="teasers" allowContentTypes="(com.gjensidige.builders.theme:link|com.gjensidige.builders.theme:link-application|com.gjensidige.builders.theme:page-article|com.gjensidige.builders.theme:page-general)" optional="true" />
<x-data name="menu-item" allowContentTypes="(com.gjensidige.builders.theme:page-article|com.gjensidige.builders.theme:page-general|com.gjensidige.builders.theme:link|com.gjensidige.builders.theme:menu-heading|base:shortcut)" optional="true" />
```

After some chatting on Slack with Sergei it seems the regex-ness of these allow patterns is only activated when a * is present. This wasn't the case before. And you never needed to type out the name of the app. The docs for this is very sparse so it's hard to know what is correct way of doing this. Examples for ContentSelectors is far more expanded, but those patterns looks to differ from x-data patterns slightly.

Can this be fixed so the old way keeps on working, or can we get another way that works that basically lets us handpick specific content types from current app (but some times other apps) and mix those with XP bundled types like base:shortcut.?

rymsha:
Previously pattern matching was using "find" instead of "match", and it was certainly problematic. If you had

`<x-data name="feed" allowContentTypes="folder" optional="true" />`

it would match any content types from any applications like "app:my-folder", "app:foldertoo", "app:infolderapp" or even "com.my.folder:content"


We don't plan to restore previous behavior completely, but we are going to revisit how we do matching when ​`*`​ is not part of an expression.

rymsha:
Fix:

- `${app}` replace anywhere in the string, not just at the beginning. Regex-escape the value.
[...]

PR #8931, opened 2021-06-25
allowContentTypes regression #8928

vbradnitski:


anatol-sialitski (approved):

modules/core/core-api/src/main/java/com/enonic/xp/app/ApplicationWildcardMatcher.java
```diff
@@ -30,17 +26,16 @@ public boolean matches( final String wildcard, T input )
...
 
         return t -> predicate.test( toNameFunction.apply( t ) );
     }
 
     private static String resolveAppPlaceholder( final String nameToResolve, final ApplicationKey applicationKey )
     {
-        return nameToResolve.startsWith( APP_WILDCARD_PREFIX )
-            ? applicationKey + nameToResolve.substring( APP_WILDCARD.length() )
-            : nameToResolve;
+        return nameToResolve.replace( APP_WILDCARD, applicationKey.getName().replace( ".", "\\." ) );
```
rymsha, line 39: it may create more issues. Use `Matcher.quoteReplacement` together with `Pattern.quote`

rymsha (changes requested):

PR #8933, opened 2021-06-28
allowContentTypes regression #8928

vbradnitski:


modules/core/core-api/src/main/java/com/enonic/xp/app/ContentTypeParseMode.java
```diff
@@ -0,0 +1,32 @@
+package com.enonic.xp.app;
+
+public enum ContentTypeParseMode
```
rymsha, line 3: make it public internal static class of ApplicationWildcardMatcher and call it Mode

rymsha (changes requested):
Append of application name to application-less patterns is not needed?

modules/core/core-api/src/main/java/com/enonic/xp/app/ApplicationWildcardMatcher.java
```diff
@@ -8,18 +8,18 @@
...
-    private static final String ANY_WILDCARD = "*";
-
-    private static final String APP_WILDCARD_PREFIX = APP_WILDCARD + ":";
-
     private final ApplicationKey applicationKey;
 
     private final Function<T, String> toNameFunction;
 
-    public ApplicationWildcardMatcher( final ApplicationKey applicationKey, final Function<T, String> toNameFunction )
+    private final Mode contentTypeParseMode;
```
rymsha, line 15: just mo
```suggestion
    private final Mode mode;
```

modules/core/core-api/src/main/java/com/enonic/xp/app/ApplicationWildcardMatcher.java
```diff
@@ -29,18 +29,59 @@ public boolean matches( final String wildcard, T input )
...
+        }
+
+        public String value()
+        {
+            return this.value;
+        }
+
+        public static Mode from( final String state )
+        {
+            if ( LEGACY.value().equals( state ) )
```
rymsha, line 75: Use switch case

rymsha (changes requested):
I still don't think that in match mode
"anymal" would match my.app:anymal

rymsha:
`ApplicationWildcardMatcher.Mode` should be specified as "LEGACY" or "MATCH". "MATCH" is default. (notice - always capital letters)
Restored LEGACY mode to be even closer to how it was in 7.6: If string starts with `${app}` or contains `:` or contains `*` - application name will be prepended to expression.

In MATCH mode is none of of `:\^$.|?*+()[{` characters are part of expression, application name will be prepended to expression. (notice that `:` is part of it to support documented expressions like `base:folder` to still work. Also notice that `.` is part of it for completeness, as it is never  part of content-type-name relative part (we should make sure it is true in XP8) and actual regex metacharacter. `{[` - are not regex metacharacters, but it it does not matter because they are never part of content-type-name relative part or app-name

`XmlInputTypeConfigMapper` no longer prepends application name to `*contentType` like tags (logic for service, mixintype and relationshiptype is preserved as is) -> all logic for expression matching is now in one place.

rymsha (approved):

Issue #8968, opened 2021-07-27
Move ContentTypeResource endpoint


pmi:
@pmi commented on [Tue Jul 27 2021](https://github.com/enonic/app-contentstudio/issues/3338)

PR #8969, opened 2021-07-27
Move ContentTypeResource endpoint #8968

pmi:


modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/schema/content/ContentTypeIconUrlResolver.java
```diff
@@ -9,7 +9,7 @@
     extends IconUrlResolver
 {
 
-    public static final String REST_SCHEMA_ICON_URL = "/admin/rest/schema/content/icon/";
+    public static final String REST_SCHEMA_ICON_URL = "/admin/v2/rest/schema/content/icon/";
```
rymsha, line 12: Should it be here?

rymsha (changes requested):

modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/schema/content/ContentTypeIconUrlResolver.java
pmi: Yeah, app resource and some app json classes use ContentTypeIconUrlResolver.java, so it has to remain here.
But since the content type resource was moved the url should be updated.

pmi: Reverted to old url and deprecated ContentTypeIconURLResolver 
https://github.com/enonic/xp/pull/8969/files#diff82b9c16be5a8757ffcd6592c1f7199a35d1410db060e6ca8437a3f50c3716465

rymsha (approved):

Issue #8998, opened 2021-08-05
Move ApplicationResource endpoint


pmi:
@pmi commented on [Thu Aug 05 2021](https://github.com/enonic/app-applications/issues/432)

@pmi commented on [Thu Aug 05 2021](https://github.com/enonic/app-contentstudio/issues/3381)

PR #9008, opened 2021-08-10
Move ApplicationResource endpoint #8998

pmi:


rymsha (approved):

modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/rest/resource/auth/AuthResource.java
```diff
@@ -28,13 +28,13 @@
 
     private SecurityService securityService;
 
+    //TODO: not used
```
rymsha, line 31: should not be part of this PR

modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/schema/IconUrlResolver.java
```diff
@@ -1,4 +1,4 @@
-package com.enonic.xp.admin.impl.rest.resource.schema;
+package com.enonic.xp.admin.impl.schema;
```
rymsha, line 1: why would it be necessary?

modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/widget/WidgetIconUrlResolver.java
```diff
@@ -1,6 +1,6 @@
 package com.enonic.xp.admin.impl.widget;
 
-import com.enonic.xp.admin.impl.rest.resource.schema.IconUrlResolver;
+import com.enonic.xp.admin.impl.schema.IconUrlResolver;
```
rymsha, line 3: why would it be necessary?

rymsha (changes requested):

rymsha (changes requested):
Please, make it one commit

rymsha (approved):

Issue #9035, opened 2021-08-24
Return the list of dependent item ids in getDependencies API response


vbradnitski:
@alansemenov commented on [Wed Jun 30 2021](https://github.com/enonic/app-contentstudio/issues/3359)

[getDependencies API](https://github.com/enonic/app-contentstudio/blob/master/modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/ContentResource.java#L527) at the moment returns _aggregated_ data on dependencies for provided content ids. We need the actual _list_ of dependent ids in addition to that. 

Another alternative is a brand new admin API which will return only dependency ids (inbound and outbound should still be separated in the response) for provided content ids.

![image](https://user-images.githubusercontent.com/11612490/123934056-36ef3b80-d993-11eb-9afe-e15ff90adfad.png)

---

@alansemenov commented on [Tue Aug 17 2021](https://github.com/enonic/app-contentstudio/issues/3359#issuecomment-900291723)

Added to 3.4.0 milestone. We cannot release https://github.com/enonic/app-contentstudio/issues/3260 without this.

PR #9036, opened 2021-08-24
Return the list of dependent item ids in getDependencies API response…

vbradnitski:
… #9035


anatol-sialitski (approved):

rymsha (changes requested):
It is probably possible do do without API changes

Issue #9049, opened 2021-09-03
Creation info is removed after job update


vbradnitski:
`creator` and `createdTime` removed from Scheduled Job after it was updated.

It causes missing audit log and exception during modify 
```
Audit log failed
java.lang.NullPointerException: null
	at com.enonic.xp.impl.scheduler.ScheduleAuditLogSupportImpl.addResult(ScheduleAuditLogSupportImpl.java:152)
	at com.enonic.xp.impl.scheduler.ScheduleAuditLogSupportImpl.doUpdate(ScheduleAuditLogSupportImpl.java:87)
	at com.enonic.xp.impl.scheduler.ScheduleAuditLogSupportImpl.lambda$modify$2(ScheduleAuditLogSupportImpl.java:77)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
```

rymsha:
Fix is to copy those _fileds_ from a job to be modified.

Since versions 7.7.0 to 7.7.2 could have jobs without creator - add a null check into audit log serializer.

PR #9050, opened 2021-09-03
Creation info is removed after job update #9049

vbradnitski:


modules/core/core-scheduler/src/main/java/com/enonic/xp/impl/scheduler/ScheduleAuditLogSupportImpl.java
```diff
@@ -149,7 +149,10 @@ private void addResult( final PropertySet targetSet, final ScheduledJob job )
             targetSet.setString( ScheduledJobPropertyNames.USER, job.getUser().toString() );
         }
 
-        targetSet.setString( ScheduledJobPropertyNames.CREATOR, job.getCreator().toString() );
+        if ( job.getCreator() != null )// 7.7 produced jobs with an empty `creator` and `createdTime` fields
```
rymsha, line 152: Let's move this comment into test.
Version is 7.7.0 to 7.7.2.

vbradnitski: But 7.7.0 also did it

rymsha (approved):

modules/core/core-scheduler/src/test/java/com/enonic/xp/impl/scheduler/ScheduleAuditLogSupportImplTest.java
```diff
@@ -217,6 +217,57 @@ public void testUpdateContent()
                       argumentCaptor.getValue().getData().getSet( "result" ).getSet( "calendar" ).getString( "value" ) );
     }
 
+    @Test
+    public void testUpdateWithoutCreator()// 7.7 produced jobs with an empty `creator` and `createdTime` fields
```
rymsha, line 221:
```suggestion
    public void testUpdateWithoutCreator() // XP 7.7.0 to 7.7.2 produced jobs with an empty `creator` and `createdTime` fields
```

anatol-sialitski (approved):

Issue #9082, opened 2021-09-16
lib-portal img sizes attribute support


poi33:
https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-srcset
"If the srcset attribute is present and has any image candidate strings using a width descriptor, the sizes attribute must also be present" 

Not having sizes attribute makes the html element invalid. Even if most browsers still render it and handle it properly.

rymsha:
without srcset browser assumes 100vw, some sources say (does not seem to be part of standard thought).
We will add optional `imageSizes` attribute - string we add to `img` `sizes`

PR #9084, opened 2021-09-17
lib-portal img sizes attribute support #9082

anatol-sialitski:


modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/HtmlLinkProcessor.java
```diff
@@ -158,6 +159,11 @@ public String process( final String text, final String urlType, final PortalRequ
                             {
                                 replacement.append( " srcset=\"" ).append( srcsetValues ).append( "\"" );
                             }
+
+                            if ( imageSizes != null && !"".equals( imageSizes.trim() ) )
```
rymsha, line 163:
```suggestion
                            if ( imageSizes != null && !imageSizes.isBlank()  )
```

rymsha (approved):

vbradnitski (approved):

Issue #9100, opened 2021-09-27
Filter causes duplication of query parameters


alansemenov:
`site.xml`:
```XML
  <mappings>
    <mapping filter="/site/mappings/filter.js">
      <pattern>/filter</pattern>
    </mapping>
  </mappings>
</site>
```

`/site/mappings/filter.js`:
```Javascript
exports.filter = function (req, next) {
    log.info(JSON.stringify(req.params, null, 4));
    return next(req);
};
```

A part added to the default site's template:
```Javascript
exports.get = function (req, next) {
    log.info(JSON.stringify(req.params, null, 4));
    return {
        contentType: 'text/html',
        body: body
    };
};
```

This setup will output the following in the XP log for url `http://localhost:8080/admin/site/preview/default/draft/features/filter?q=1`:

```
16:18:35.280 INFO  com.enonic.app.features - (/site/mappings/filter.js) {
    "q": "1"
}
16:18:35.289 INFO  com.enonic.app.features - (/site/parts/my-part/my-part.js) {
    "q": [
        "1",
        "1"
    ]
}
```

After going through the filter, `q` parameter becomes an array with duplicated value.

rymsha:
It appeared after #8210 (although, it is not a regression)

Interestingly enough we are not the first who face such issue https://github.com/eclipse/jetty.project/issues/2074

rymsha:
prams from filter result are appended to params with the same name - that is causing bug in cases where filter uses `next` as a return value.

To do a fix we decided to changes the behavior:
prams with the same name in filter result _replace_ params in the original request.

PR #9109, opened 2021-09-30
Filter causes duplication of query parameters #9100

rymsha:


modules/portal/portal-impl/src/test/java/com/enonic/xp/portal/impl/filter/PortalRequestSerializerTest.java
```diff
@@ -152,6 +155,36 @@ public void serializeMultiParams()
...
 
+    @Test
+    public void serialize_no_params_duplication()
+        throws Exception
+    {
+        final PortalRequest sourceRequest = new PortalRequest();
+        sourceRequest.getParams().put( "a", "oldA" );
+
+        sourceRequest.getParams().put( "b", "oldB" );
+        sourceRequest.getParams().put( "b", "oldB" );
```
anatol-sialitski, line 166: should be `c` instead of `b` and `oldC`

anatol-sialitski (approved):

vbradnitski (approved):

Issue #8987, opened 2021-07-30
Extend lib-auth to include user authentication 


poi33:
Currently its not possible to check if a user has a valid username + password without loging into the system. 

Can we create a javascript function that will do this? 

The use case for it is 2step authenticaton of users, where they first have to send in a username and password, then a second step to login to the system.

rymsha:
We can trivially implement it extending login function with `scope=NONE` 
in this scope login either fails with exception or succeeds (bud does not persist the login fact anyhow)

sigdestad:
That is a quickfix we can live with. Can you make sure this gets into XP 7.8 @rymsha ?

PR #9192, opened 2021-11-03
#8987 Extend lib-auth to include user authentication

anatol-sialitski:


modules/lib/lib-auth/src/test/java/com/enonic/xp/lib/auth/LoginHandlerTest.java
```diff
@@ -75,6 +75,26 @@ public void testLoginSuccess()
...
+        final IdProviders idProviders =
+            IdProviders.from( IdProvider.create().displayName( "system" ).key( IdProviderKey.from( "system" ) ).build() );
+
+        Mockito.when( this.securityService.authenticate( Mockito.any() ) ).thenReturn( authInfo );
+        Mockito.when( this.securityService.getIdProviders() ).thenReturn( idProviders );
+
+
+        runFunction( "/test/login-test.js", "loginWithScopeNONE" );
+
+        assertNull( ContextAccessor.current().getLocalScope().getSession() );
```
rymsha, line 95: should you also check that there is no AuthInfo attribute?

rymsha (approved):

modules/lib/lib-auth/src/test/java/com/enonic/xp/lib/auth/LoginHandlerTest.java
anatol-sialitski: Ah, sure. I thought that will be enough to check session, because that attribute is added to session if last exists

vbradnitski (approved):

Issue #9205, opened 2021-11-12
Improve reliability of CheckNodeExistsCommand


vbradnitski:
#9061 Describes a set of reasons why nodes with the same name can appear. Biggest issue is that there is a "check before create" - very race prone appear, but we can't deal with it right now.

Another reason is that `CheckNodeExistsCommand` checks existence of a node 
- without refreshing the index (and since we use search - 1s refresh delay applies)
- checks only data in local replica (where data may not have been replicated to yet)
- does check in two steps: first searches for node path in `search` index, then gets `NodeBranchEntry` by id from `branch` index.

Improvements for BranchService (used by CheckNodeExistsCommand):
- use only one search query on `branch` index. It is possible to get all necessary info from there. This way we don't depend on refresh of `search` index
- refresh branch index (or just all indices) when checking
- use `_primary` preference

rymsha:
Other fixes:
- since we always use single shard, shard routing is unnecessary and confusing. (BTW, there are a few aggregation queries that assume only-single-shard per index.) In XP routing needed only for parent/child index and even there we just spoof it (we set nodeId.toString value just to bypass elasticsearch check for routing field.  We remove routing wherever possible.
-

PR #9208, opened 2021-11-12
Improve reliability of CheckNodeExistsCommand #9205

vbradnitski:


modules/core/core-api/src/main/java/com/enonic/xp/node/NodeService.java
```diff
@@ -29,14 +29,18 @@
 
     Node getById( NodeId id );
 
+    @Deprecated
```
rymsha, line 32: These are in use in Content API, and should be redone, too.

modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/elasticsearch/SearchRequestBuilderFactory.java
```diff
@@ -57,11 +57,11 @@ public SearchRequestBuilder createSearchRequest()
...
-        searchRequestBuilder.
-            setExplain( query.isExplain() ).
-            setSearchType(
-                query.getSearchOptimizer().equals( SearchOptimizer.ACCURACY ) ? SearchType.DFS_QUERY_THEN_FETCH : SearchType.DEFAULT ).
-            setPreference( SearchPreference.LOCAL.getName() );
+        searchRequestBuilder.setExplain( query.isExplain() )
+            .setSearchType(
+                query.getSearchOptimizer().equals( SearchOptimizer.ACCURACY ) ? SearchType.DFS_QUERY_THEN_FETCH : SearchType.DEFAULT )
+            .setPreference(
+                query.getSearchPreference() != null ? query.getSearchPreference().toString() : SearchPreference.LOCAL.toString() );
```
rymsha, line 64: this way specifying null won't be possible...

modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/elasticsearch/executor/CountExecutor.java
```diff
@@ -24,12 +25,13 @@ public static Builder create( final Client client )
...
-            setQuery( query.getQuery() ).
-            setSearchType( SearchType.COUNT ).
-            setPreference( searchPreference );
+        SearchRequestBuilder searchRequestBuilder =
+            new SearchRequestBuilder( this.client, SearchAction.INSTANCE ).setIndices( query.getIndexNames() )
+                .setTypes( query.getIndexTypes() )
+                .setQuery( query.getQuery() )
+                .setSearchType( SearchType.COUNT )
+                .setPreference(
+                    query.getSearchPreference() != null ? query.getSearchPreference().toString() : SearchPreference.LOCAL.toString() );
```
rymsha, line 34: this way specifying null won't be possible...

modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/node/CheckNodeExistsCommand.java
```diff
@@ -37,20 +39,21 @@ public static Builder create( AbstractNodeCommand source )
 
     public boolean execute()
     {
-        final InternalContext context = InternalContext.from( ContextAccessor.current() );
+        final InternalContext context =
```
rymsha, line 42: Does it mean we always check existence on primary? (I actually like it, but it may hurt performance of Load/Import)

modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/node/CheckNodeExistsCommand.java
```diff
@@ -61,7 +64,7 @@ else if ( !found.equals( skipThrowIfSameId ) )
 
         private boolean throwIfExists;
 
-        private NodeId skipThrowIfSameId;
+        private boolean refresh = false;
```
rymsha, line 67: Maybe should be true by default, and only set to false where we prefer performance over consistency (load/import)

Maybe give it a different name and join with use of primary preference. Need to discuss

modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/node/DuplicateNodeCommand.java
```diff
@@ -225,20 +225,15 @@ private String resolveNewNodeName( final Node existingNode )
         // Process as file name as it is so for images
         String newNodeName = DuplicateValueResolver.fileName( existingNode.name().toString() );
 
-        boolean resolvedUnique = false;
-
-        while ( !resolvedUnique )
+        while ( true )
```
rymsha, line 228: we may likely drop this optimization and instead catch NodeAlreadyExists with retry.

modules/core/core-api/src/main/java/com/enonic/xp/node/NodeService.java
rymsha: Maybe just keep these methods, without deprecation. But don't use NodeId internally - leave it for another round of refactoring

modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/storage/NodeStorageService.java
```diff
@@ -77,8 +77,5 @@
 
     void handleNodePushed( NodeId nodeId, NodePath nodePath, NodePath currentTargetPath, InternalContext nodeContext );
 
-    Node getNode( NodeId nodeId, NodeVersionId nodeVersionId, InternalContext context );
-
-    Node getNode( NodePath nodePath, NodeVersionId nodeVersionId, InternalContext context );
-
+    Node getNode( NodeVersionId nodeVersionId, InternalContext context );
```
rymsha, line 80: signature looks equal to `get` method. Any difference?

modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/storage/NodeStorageServiceImpl.java
```diff
@@ -212,7 +212,7 @@ public Node updateMetadata( final Node node, final InternalContext context )
     public void updateVersion( final Node node, final NodeVersionId nodeVersionId, final InternalContext context )
     {
         //TODO Check
```
rymsha, line 214: Haven't you checked it :-) ?

modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/node/CheckNodeExistsCommand.java
rymsha: MODE Accuracy vs Speed

modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/node/DuplicateNodeCommand.java
```diff
@@ -57,20 +57,28 @@ public Node execute()
...
-            binaryService( binaryService ).
-            build().
-            execute();
+                duplicatedNode =
+                    CreateNodeCommand.create( this ).params( processedParams ).binaryService( binaryService ).build().execute();
+
+            }
+            catch ( NodeAlreadyExistAtPathException e )
+            {
+                // try again with other name
```
rymsha, line 78: Add debug logging
Add test

modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/node/MoveNodeCommand.java
```diff
@@ -274,10 +271,7 @@ private void verifyNoExistingAtNewPath( final NodePath newParentPath, final Node
     {
         final NodePath newNodePath = NodePath.create( newParentPath, newNodeName.toString() ).build();
 
-        CheckNodeExistsCommand.create( this ).
-            nodePath( newNodePath ).
-            throwIfExists().
-            build().
+        CheckNodeExistsCommand.create( this ).nodePath( newNodePath ).throwIfExists().mode( CheckNodeExistsCommand.Mode.SPEED ).build().
```
rymsha, line 274: For "top" nodes we better use ACCURACY, for others we can simplify to SPEED, or don't' do a check at all.

rymsha (changes requested):

rymsha (approved):

rymsha (changes requested):
SearchPreference became API. Let's wait with it.

Issue #9228, opened 2021-11-23
contextLib: setting contentRootPath leads to an error


ashklianko:
If we're setting context with content root 
```
contextLib.run(
        {
            repository: repositoryId,
            branch: 'draft',
            attributes: {
                contentRootPath: '/archive'
            }
        },
        function () {
            return doGetContent(contentId, versionId);
        }
    );
```

then `Context` object will have contentRootPath set as string, and when `ContentNodeHelper.getContentRoot()` will be invoked
```
public static NodePath getContentRoot()
    {
        final NodePath nodePath = (NodePath) ContextAccessor.current().getAttribute( CONTENT_ROOT_PATH_ATTRIBUTE );
        return nodePath != null ? nodePath.asAbsolute() : ContentConstants.CONTENT_ROOT_PATH;
    }
```
it will try to convert contentRootPath `String`  to a `NodePath` and will fail

vbradnitski:
User must use NodePath.class object instead of sending a string. `content.js` will share content-api contstants to provide simpler way to use them:
```

exports.ARCHIVE_ROOT_PATH = com.enonic.xp.archive.ArchiveConstants.ARCHIVE_ROOT_PATH;
exports.CONTENT_ROOT_PATH = com.enonic.xp.content.ContentConstants.CONTENT_ROOT_PATH;
```
 after that context can be created with provided contentRootPath attribute value:

 ```
contextLib.run({
        repoId: 'com.enonic.cms.default',
        branch: 'draft',
        attributes: {
            contentRootPath: contentLib.ArCHIVE_ROOT_PATH
        }
    }, function () {
...

```

PR #9233, opened 2021-11-23
contextLib: setting contentRootPath leads to an error #9228

vbradnitski:


rymsha (approved):

modules/lib/lib-content/src/main/resources/lib/xp/content.js
```diff
@@ -8,6 +8,10 @@
  */
 /* global __*/
 
+exports.CONTENT_ROOT_PATH = com.enonic.xp.content.ContentConstants.CONTENT_ROOT_PATH;
```
anatol-sialitski, line 11: I don't like that approach I would prefer to use helper for constants. But it works so I approve pr.

anatol-sialitski (approved):

rymsha (changes requested):
add test and example

rymsha (approved):

Issue #9239, opened 2021-11-24
Duplicate of a content is inherited in layer without child elements


vbradnitski:
@sgauruseu commented on [Tue Aug 10 2021](https://github.com/enonic/app-contentstudio/issues/3400)

1. Create a layer in Default project
2. Switch to Default context and duplicate `Features` and `Large Tree` folder
3. Switch to the layer's context

BUG - Inherited duplicated content  have no children 
![image](https://user-images.githubusercontent.com/3728712/128887186-91cd66d5-51a5-4c93-a20f-687008fd6bc5.png)

vbradnitski:
The problem is we have no events for children of a duplicated node. So it must be added to the sync process to be able to see them.

PR #9240, opened 2021-11-24
#9239 Duplicate of a content is inherited in layer without child elements

vbradnitski:
…9239

### DIFF
--- a/modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/node/NodeServiceImpl.java
+++ b/modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/node/NodeServiceImpl.java
@@ -592,20 +592,19 @@ public class NodeServiceImpl
     public Node duplicate( final DuplicateNodeParams params )
     {
         verifyContext();
-        final Node duplicatedNode = DuplicateNodeCommand.create().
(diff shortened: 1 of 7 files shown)


modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/node/NodeServiceImpl.java
```diff
@@ -592,20 +592,19 @@ public PushNodesResult push( final NodeIds ids, final Branch target, final PushN
...
+        final DuplicateNodeResult result = DuplicateNodeCommand.create()
+            .params( params )
+            .indexServiceInternal( this.indexServiceInternal )
+            .binaryService( this.binaryService )
+            .storageService( this.nodeStorageService )
+            .searchService( this.nodeSearchService )
+            .build()
+            .execute();
+
+        this.eventPublisher.publish( NodeEvents.duplicated( result.getNode() ) );
```
rymsha, line 604: Isn't in the idea that duplicated event contains all duplicated node ids?

rymsha (changes requested):

rymsha (approved):

anatol-sialitski (approved):

Issue #8160, opened 2020-06-16
lib-portal serviceUrl needs protocol parameter for ws:// and wss://


ComLock:
https://repo.enonic.com/public/com/enonic/xp/docs/6.15.9/docs-6.15.9-libdoc.zip!/module-portal.html#.serviceUrl
https://repo.enonic.com/public/com/enonic/xp/docs/7.3.0/docs-7.3.0-libdoc.zip!/module-portal.html#.serviceUrl
https://developer.enonic.com/docs/xp/stable/api/lib-portal#serviceurl

Perhaps name it scheme, like here:
https://developer.enonic.com/docs/xp/stable/framework/http#http-request

ComLock:
I thought there used to be some parameter for http and https, but I guess that is automatic from context?

rymsha:
https://discuss.enonic.com/t/give-portallib-serviceurl-option-to-create-websocket-urls-too/2372/2

sigdestad:
I guess customising protocol should be an option for all URL functions, so make this task cover all of them.

rymsha:
There are only 4 schemas portal supports `http`, `https`, `ws`, `wss`

And if page is on `https://` , websocket should be `wss://`
I can't imagine a (non academical) case where `https://` page would want to provide a link to `http` or `ws`  especially when we are dealing with the same site.

There is an optional `type` parameter already. we can extend it with `websocket` value - that would generate a `ws(s)` absolute url.

PR #9252, opened 2021-12-06
lib-portal serviceUrl needs protocol parameter for ws:// and wss:// #8160

anatol-sialitski:


modules/lib/lib-portal/src/test/resources/test/url-test.js
```diff
@@ -176,6 +176,21 @@ exports.serviceUrlTest = function () {
...
+        service: 'myservice',
+        type: 'websocket',
+        params: {
+            a: 1,
+            b: [1, 2]
+        }
+    });
+
+    // NOTE: This is not the actual url. Only a mock representation.
+    assert.assertEquals('ServiceUrlParams{type=websocket, params={a=[1], b=[1, 2]}, service=myservice}', result);
```
rymsha, line 190: Why testing mock?

modules/web/web-api/src/main/java/com/enonic/xp/web/servlet/ServletRequestUrlHelper.java
```diff
@@ -126,13 +126,24 @@ public static String getServerUrl()
...
+
+    public static String getServerUrl( final HttpServletRequest req, final boolean websocket )
     {
         final StringBuilder str = new StringBuilder();
 
         //Appends the scheme part
         String scheme = req.getScheme();
-        str.append( scheme );
-
+        if ( websocket )
```
rymsha, line 139: This entire class deserves depreciation. Don't extend it

modules/lib/lib-portal/src/main/resources/lib/xp/portal.js
```diff
@@ -16,7 +16,7 @@
  * @param {object} params Input parameters as JSON.
  * @param {string} params.path Path to the asset.
  * @param {string} [params.application] Other application to reference to. Defaults to current application.
- * @param {string} [params.type=server] URL type. Either `server` (server-relative URL) or `absolute`.
+ * @param {string} [params.type=server] URL type. Either `server` (server-relative URL) or `absolute` or `websocket`.
```
rymsha, line 19: Isn't it strange for asset/attachment/image etc. to even mention we socket?

rymsha (changes requested):

modules/lib/lib-portal/src/test/resources/test/url-test.js
anatol-sialitski: Because all tests in this file written using this approach

modules/lib/lib-portal/src/main/resources/lib/xp/portal.js
anatol-sialitski: All these functions use the same `type`. After adding the `websocket` value that available for the rest of functions.

modules/web/web-api/src/main/java/com/enonic/xp/web/servlet/ServletRequestUrlHelper.java
anatol-sialitski: Any suggestions? This is a public API and that should be done here for now. I am ready to discuss this issue if needed.

rymsha (approved):

vbradnitski (approved):

Issue #8037, opened 2020-04-21
Application node metadata does not get updated when application gets updated


rymsha:
`system-repo / master / root /applications /...` contains applications metadata but it looks like it only gets populated with first app installation and most of the data is kept the same even if application got updated.

It is super confusing and complicates debugging of application installation process, otherwise harmless.

PR #9264, opened 2021-12-10
#8037 Application node metadata does not get updated when application…

anatol-sialitski:
… gets updated

### DIFF
--- a/modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationNodeTransformer.java
+++ b/modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationNodeTransformer.java
@@ -31,7 +31,7 @@ class ApplicationNodeTransformer
         final PropertyTree data = new PropertyTree();
         data.setString( ApplicationPropertyNames.DISPLAY_NAME, app.getDisplayName() );
         data.setString( ApplicationPropertyNames.MAX_SYSTEM_VERSION, app.getMaxSystemVersion() );
-        data.setString( ApplicationPropertyNames.MIN_SYSTEM_VERSION, app.getMaxSystemVersion() );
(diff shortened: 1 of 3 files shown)


modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationServiceImpl.java
```diff
@@ -214,7 +214,7 @@ private Application doInstallLocalApplication( final ByteSource byteSource )
     {
         final ApplicationKey applicationKey = getApplicationKey( byteSource );
 
-        final Application application = installOrUpdateApplication( byteSource, applicationKey, true, false );
+        final Application application = installOrUpdateApplication( byteSource, applicationKey, true, true );
```
rymsha, line 217: local apps must not update repository
```suggestion
        final Application application = installOrUpdateApplication( byteSource, applicationKey, true, false );
```

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationServiceImpl.java
```diff
@@ -319,7 +319,7 @@ private Application doInstallStoredApplication( final NodeId nodeId )
             return null;
         }
 
-        return installOrUpdateApplication( byteSource, applicationKey, false, false );
+        return installOrUpdateApplication( byteSource, applicationKey, false, true );
```
rymsha, line 322: stored app install happens on remote node "because" repository was changed. No need to update it again.

```suggestion
        return installOrUpdateApplication( byteSource, applicationKey, false, false );
```

rymsha (changes requested):

vbradnitski (approved):

rymsha (approved):

Issue #9269, opened 2021-12-15
Trace is failed due to NPE if content in request is not specified


anatol-sialitski:
Trace is failed due to NPE if content in a request is not specified.

PR #9270, opened 2021-12-15
#9269 Trace is failed due to NPE if content in request is not specified

anatol-sialitski:


modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/mapping/MappingFilterHandlerWorker.java
```diff
@@ -41,7 +41,10 @@ public PortalResponse execute()
         final Trace trace = Tracer.current();
         if ( trace != null )
         {
-            trace.put( "contentPath", this.request.getContent().getPath().toString() );
+            if ( this.request.getContent() != null )
+            {
+                trace.put( "contentPath", this.request.getContent().getPath().toString() );
```
rymsha, line 46: request has `contentPath()` method that makes null check unneeded.

rymsha (changes requested):

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/mapping/MappingFilterHandlerWorker.java
anatol-sialitski: Thank you! Did it.

rymsha (approved):

vbradnitski (approved):

PR #9283, opened 2021-12-30
removed image service log statement

poi33:
Log statement removed. 

This should be picked up in code reviews.


rymsha:
will be fixed by #9286

Issue #9325, opened 2022-01-25
Handle 418 response code in page editor


pmi:
@pmi commented on [Wed Jan 19 2022](https://github.com/enonic/app-contentstudio/issues/4138)

PR #9326, opened 2022-01-25
Handle 418 response code in page editor #9325

pmi:


modules/web/web-impl/src/main/java/com/enonic/xp/web/impl/exception/ExceptionMapperImpl.java
```diff
@@ -48,6 +48,6 @@ public void throwIfNeeded( final WebResponse res )
 
     private boolean isError( final HttpStatus status )
     {
-        return ( status != null ) && ( status.is4xxClientError() || status.is5xxServerError() );
+        return status != null && status != HttpStatus.IM_A_TEAPOT && ( status.is4xxClientError() || status.is5xxServerError() );
```
rymsha, line 51: That is too deep to handle 418 specially. 
BasePortalHandler, WebAppHandler and WebDispatcherServlet use this method (indirectly).
`WebAppHandler` would not enjoy a new behavior.

Please do the special handling of 418 only in Handler that needs it.

rymsha (changes requested):

modules/web/web-impl/src/main/java/com/enonic/xp/web/impl/exception/ExceptionMapperImpl.java
pmi: Moved 418 handling to WebDispatcherServlet

Issue #9327, opened 2022-01-27
No error returned on rendering of a page template without descriptor


pmi:
Should return status 500 instead

PR #9330, opened 2022-01-31
Page template content should not be rendered in inline mode without p…

pmi:
…age descriptor #9327

- fixed the out of bounds exception in sliceLines


rymsha (approved):

pmi:
Well, it does contain issue number if you expand it and it looks exactly the way issue looks on github.
Have been doing this way for ages :)

rymsha (approved):

PR #9364, opened 2022-02-11
ResourceService should be able to load data from nodes #9346

vbradnitski:


rymsha (approved):

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/resource/ResourceServiceImpl.java
```diff
@@ -94,4 +122,41 @@ public void invalidate( final ApplicationKey key, final ApplicationInvalidationL
         LOG.debug( "Cleanup Resource cache for {}", key );
         this.cache.invalidate( key );
     }
+
+    private ValueResource findProjectResource( final ResourceKey key )
```
rymsha, line 126: this method goes to `NodeResourceApplicationUrlResolver`

rymsha (changes requested):

sigdestad:
Just a thought, how about shredding these files to disk and load resources from there?

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/resolver/NodeResourceApplicationUrlResolver.java
```diff
@@ -0,0 +1,68 @@
...
+    {
+        this.applicationKey = ApplicationKey.from( bundle );
+        this.nodeService = nodeService;
+    }
+
+    @Override
+    public Set<String> findFiles()
+    {
+        return VirtualAppContext.createContext().callWith( () -> {
+            final FindNodesByQueryResult nodes = this.nodeService.findByQuery( NodeQuery.create()
```
rymsha, line 38: Maybe use DSL?

rymsha (approved):

Issue #9402, opened 2022-03-16
Server-side error when archiving content


vbradnitski:
@alansemenov commented on [Wed Mar 02 2022](https://github.com/enonic/app-contentstudio/issues/4291)

Archiving a content seems to always raise this error in the log. It doesn't seem to matter if it's one or more items. For several items the error seems to be given for each single item.

Originally described [here](https://github.com/enonic/app-contentstudio/issues/3610) (before the archive feature was even released) and closed as fixed.

```
13:27:48.051 ERROR c.e.x.c.i.c.ContentNodeTranslator - Failed to translate node '/archive/log.txt' [c4948850-f2e7-4629-9fb6-a3a9d3b97065] to content
com.enonic.xp.content.ContentNotFoundException: Content with id [c4948850-f2e7-4629-9fb6-a3a9d3b97065] was not found in branch [draft] in content root [/content]
	at com.enonic.xp.core.impl.content.ContentNodeTranslator.doTranslate(ContentNodeTranslator.java:99)
	at com.enonic.xp.core.impl.content.ContentNodeTranslator.doTranslate(ContentNodeTranslator.java:89)
	at com.enonic.xp.core.impl.content.ContentNodeTranslator.doTranslate(ContentNodeTranslator.java:66)
	at com.enonic.xp.core.impl.content.ContentNodeTranslator.fromNodes(ContentNodeTranslator.java:39)
	at com.enonic.xp.core.impl.content.GetContentByIdsCommand.doExecute(GetContentByIdsCommand.java:39)
	at com.enonic.xp.core.impl.content.GetContentByIdsCommand.execute(GetContentByIdsCommand.java:29)
	at com.enonic.xp.core.impl.content.ContentServiceImpl.doGetByIds(ContentServiceImpl.java:668)
	at com.enonic.xp.core.impl.content.ContentServiceImpl.getByIds(ContentServiceImpl.java:651)
	at com.enonic.xp.app.contentstudio.rest.resource.content.ContentResource.getByIds(ContentResource.java:713)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
[...]

PR #9403, opened 2022-03-16
Server-side error when archiving content #9402

vbradnitski:


modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/ContentNodeTranslator.java
```diff
@@ -67,7 +67,7 @@ private Contents doTranslate( final Nodes nodes, final NodesHasChildrenResult no
             }
             catch ( final Exception e )
             {
-                LOG.error( "Failed to translate node '" + node.path() + "' [" + node.id().toString() + "] to content", e );
+                LOG.debug( "Failed to translate node '" + node.path() + "' [" + node.id().toString() + "] to content", e );
```
rymsha, line 70: Maybe add special case for ContentNotFound with debug logging and other Exceptions keep logging as error?

rymsha, line 70: Maybe add special case for ContentNotFound with debug logging and other Exceptions keep logging as error?

rymsha (approved):

anatol-sialitski (approved):

Issue #9316, opened 2022-01-18
Register application actions in the Audit Log


alansemenov:
`start`, `stop`, `install` and `uninstall` actions should be registered in the Audit Log.

Only successful actions should be logged.
We should support a flag which tells the Application service whether an event should be registered in the AuditLog or not.

Data to store:

```json
  "type": ["system.application.start/stop/install/uninstall"],
  "source": ["com.enonic.xp.core-app?"],  // To be checked
  "user": ["user:system:su"],
  "objects": ["com.xp.app.my-app"],
  "data": {
    "params": {
      version ?
      checksum ?
      origin (Url or fileName) ?
      buildTime ?
    }
  }
```

PR #9405, opened 2022-03-21
Register application actions in the Audit Log #9316

anatol-sialitski:


modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationAuditLogSupportImpl.java
```diff
@@ -0,0 +1,130 @@
...
+import com.enonic.xp.data.PropertyTree;
+import com.enonic.xp.security.PrincipalKey;
+import com.enonic.xp.security.RoleKeys;
+import com.enonic.xp.security.auth.AuthenticationInfo;
+
+@Component(configurationPid = "com.enonic.xp.app")
+public class ApplicationAuditLogSupportImpl
+    implements ApplicationAuditLogSupport
+{
+    private static final String SOURCE = "com.enonic.xp.core-app";
```
rymsha, line 27:
```suggestion
    private static final String SOURCE = "com.enonic.xp.app";
```

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationAuditLogSupportImpl.java
```diff
@@ -0,0 +1,130 @@
...
+        if ( appConfig.auditlog_enabled() )
+        {
+            final PropertyTree data = new PropertyTree();
+
+            log( "system.application.install", data, AuditLogUris.from( applicationKey.toString() ) );
+        }
+    }
+
+    @Override
+    public void installStoredApplication( final ApplicationKey applicationKey, final ApplicationInstallationParams params )
```
rymsha, line 87: this should not be logged ever.

rymsha (changes requested):

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationAuditLogSupportImpl.java
anatol-sialitski: done

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationAuditLogSupportImpl.java
anatol-sialitski: done

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationAuditLogSupport.java
```diff
@@ -0,0 +1,18 @@
...
+
+public interface ApplicationAuditLogSupport
+{
+    void startApplication( ApplicationKey applicationKey );
+
+    void stopApplication( ApplicationKey applicationKey );
+
+    void installApplication( ApplicationKey applicationKey, URL url );
+
+    void installApplication( ApplicationKey applicationKey );
```
rymsha, line 15: sure you need a separate method? I don't insist, just wonder.

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationAuditLogSupportImpl.java
```diff
@@ -0,0 +1,115 @@
...
+    private static final String SOURCE = "com.enonic.xp.app";
+
+    private final AuditLogService auditLogService;
+
+    private final AppConfig appConfig;
+
+    @Activate
+    public ApplicationAuditLogSupportImpl( final AppConfig appConfig, @Reference final AuditLogService auditLogService )
+    {
+        this.appConfig = appConfig;
```
rymsha, line 35: you can write use method with `@Modify` annotation. So service is not recreated when config is changed.

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationAuditLogSupportImpl.java
```diff
@@ -0,0 +1,115 @@
...
+        this.auditLogService = auditLogService;
+    }
+
+    @Override
+    public void startApplication( final ApplicationKey applicationKey )
+    {
+        if ( appConfig.auditlog_enabled() )
```
rymsha, line 42: at could be stored in `volatile boolean` for readability.

rymsha (commented):
InstallApplicationParams triggerEvent (that is always false), should always be false. And builder method must be deprecated.

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationAuditLogSupport.java
anatol-sialitski: We can do it using one method. But I don't like to pass the `null` argument to the method. As for me, in this case the explicit argument is more suitable and a developer will not waste a time to investigate what to do this method.

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationAuditLogSupportImpl.java
anatol-sialitski: done

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationAuditLogSupportImpl.java
anatol-sialitski: done

rymsha (approved):

modules/core/core-api/src/main/java/com/enonic/xp/app/ApplicationInstallationParams.java
```diff
@@ -11,7 +11,7 @@
     private ApplicationInstallationParams( final Builder builder )
     {
         start = builder.start;
-        triggerEvent = builder.triggerEvent;
+        triggerEvent = false;
```
vbradnitski, line 14: Do we need a field anymore? Wouldn't be better to return 'false' directly in getter?

vbradnitski (changes requested):

modules/core/core-api/src/main/java/com/enonic/xp/app/ApplicationInstallationParams.java
anatol-sialitski: It still needs for `hashCode` and `equals` methods.

vbradnitski (approved):

Issue #9407, opened 2022-03-22
styles.xml for Virtual Apps


rymsha:
similarly to #9406 we need to support styles.xml in Virtual apps
It is a singleton descriptor, missing by default, can be added/deleted/updated

PR #9412, opened 2022-03-25
styles.xml for Virtual Apps #9407

vbradnitski:


modules/core/core-api/src/main/java/com/enonic/xp/resource/UpdateDynamicStylesParams.java
```diff
@@ -0,0 +1,66 @@
...
+
+import com.enonic.xp.app.ApplicationKey;
+
+public final class UpdateDynamicStylesParams
+{
+    private final ApplicationKey key;
+
+    private final String resource;
+
+    public UpdateDynamicStylesParams( final Builder builder )
```
anatol-sialitski, line 13: `private` instead of `public`

rymsha (approved):

anatol-sialitski (approved):

Issue #9414, opened 2022-03-29
GetContentByIdCommand throws invalid exception


vbradnitski:
ContentNotFoundException must be thrown instead of RuntimeException when content exists in a different content context.

https://enonic.zendesk.com/agent/tickets/5316

PR #9415, opened 2022-03-29
GetContentByIdCommand throws invalid exception #9414

vbradnitski:


rymsha (approved):

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/GetContentByIdCommand.java
```diff
@@ -29,7 +30,7 @@ Content execute()
             final Node node = nodeService.getById( nodeId );
             content = filter( translator.fromNode( node, true ) );
         }
-        catch ( NodeNotFoundException e )
+        catch ( NodeNotFoundException | ContentNotFoundException e )
```
anatol-sialitski, line 33: I think will be better return `null` instead of throw `ContentNotFoundException` in the translator.

anatol-sialitski (approved):

Issue #9426, opened 2022-04-08
BranchNotFoundException on cluster initialization


gbbirkisson:
XP version: 7.9.0-B1

Steps to replicate:
* Install docker :smile: 
* Clone https://github.com/enonic-cloud/docker-compose-enonic-xp-cluster
* Checkout branch [3master-2nodes](https://github.com/enonic-cloud/docker-compose-enonic-xp-cluster/tree/3master-2nodes)
* Run `docker-compose up`
* Since this is a race condition you might need to run it several times before the bug appears. Delete all data between runs with `docker-compose down -v`

Successful replication of the problem (master3 actual master):

```
...

master3 2022-04-08 11:25:57,431 INFO  o.e.cluster.routing.allocation - [master3] Cluster health status changed from [YELLOW] to [GREEN] (reason: [shards started [[search-com.enonic.cms.default][0]] ...]).
master3 2022-04-08 11:25:57,456 INFO  org.elasticsearch.cluster.metadata - [master3] [search-com.enonic.cms.default] create_mapping [master]
master3 2022-04-08 11:25:57,501 INFO  c.e.x.r.i.r.RepositoryServiceImpl - Created root node with id [000-000-000-000] in repository [com.enonic.cms.default]
master3 2022-04-08 11:25:57,529 INFO  org.elasticsearch.cluster.metadata - [master3] [search-system-repo] update_mapping [master]
master2 2022-04-08 11:25:57,541 INFO  com.enonic.xp.init.Initializer - Waiting [1000ms] for com.enonic.cms.default repo to be initialized
master3 2022-04-08 11:25:57,609 INFO  org.elasticsearch.cluster.metadata - [master3] [search-com.enonic.cms.default] create_mapping [draft]
master3 2022-04-08 11:25:57,702 INFO  c.e.x.c.i.p.init.ContentInitializer - Content root-node not found, creating
master3 2022-04-08 11:25:57,743 INFO  org.elasticsearch.cluster.metadata - [master3] [search-com.enonic.cms.default] update_mapping [draft]
master3 2022-04-08 11:25:57,775 INFO  c.e.x.c.i.p.init.ContentInitializer - Created content root-node: /content
master3 2022-04-08 11:25:57,838 INFO  org.elasticsearch.cluster.metadata - [master3] [search-com.enonic.cms.default] update_mapping [master]
master3 2022-04-08 11:25:57,877 INFO  com.enonic.xp.init.Initializer - com.enonic.cms.default repo successfully initialized
[...]

PR #9429, opened 2022-04-11
BranchNotFoundException on cluster initialization #9426

vbradnitski:


modules/core/core-api/src/main/java/com/enonic/xp/init/Initializer.java
```diff
@@ -33,19 +33,42 @@ protected Initializer( final Builder builder )
...
         final String initializationSubject = getInitializationSubject();
+        int failedCount = 0;
 
         for ( int i = 0; i < initializationCheckMaxCount; i++ )
         {
             if ( readyToInitialize() )
             {
-                final boolean initialized = isInitialized();
-                if ( initialized )
+                final boolean isGoingToInitialize = forceInitialization || isMaster();
```
anatol-sialitski, line 42: `isGoingToInitialize` can be declared before `for` statement

anatol-sialitski (approved):

PR #9450, opened 2022-04-26
Migrate Javascript API to Typescript

alansemenov:


modules/lib/lib-event/README.md
```diff
@@ -0,0 +1,50 @@
...
+## Use
+
+__1. Update tsconfig.json__
+
+Add the corresponding types to your `tsconfig.json` file that is used for application's server-side TypeScript code:
+
+```json
+{
+  "compilerOptions": {
+    "types": [
```
tajakobsen, line 20: Have you considered something like this:
```suggestion
    "typeRoots": [
      "@enonic"
    ]
```

modules/lib/lib-event/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "name": "Mikita Taukachou",
+    "email": "edloidas@gmail.com",
+    "url": "https://edloidas.github.io/"
+  },
+  "license": "GPL-3.0",
+  "bugs": {
+    "url": "https://github.com/enonic/xp/issues"
+  },
+  "homepage": "https://github.com/enonic/xp/tree/master#readme",
+  "devDependencies": {
```
tajakobsen, line 37: I would consider adding *eslint* and *prettier* now in the transition to TypeScript

Example how we do it in XP/TypeScript-projects: https://github.com/ItemConsulting/xp-ntb-import

modules/lib/lib-event/src/main/resources/lib/xp/event.ts
```diff
@@ -0,0 +1,136 @@
...
+    distributed?: boolean;
+
+    /**
+     * Additional data for event.
+     */
+    data?: object;
+}
+
+export interface EnonicEvent<EventData extends object = EnonicEventData> {
+    readonly type: EnonicEventTypes | string;
```
tajakobsen, line 49: Make sure that `readonly` is actually what you want for these cases.

modules/lib/lib-event/src/main/resources/lib/xp/event.ts
```diff
@@ -0,0 +1,136 @@
...
+
+export interface EnonicEventDataNode {
+    readonly id: string;
+    readonly path: string;
+    readonly branch: string;
+    readonly repo: string;
+    readonly newPath?: string; // event type <'node.moved' | 'node.renamed'>
+}
+
+type LiteralUnion<T extends U, U = string> = T | (U & Record<never, never>);
```
tajakobsen, line 69: This is used for improving code completion right? You should add a comment to this type to document it. It can be difficult to understand the usage of the type.

modules/lib/lib-event/src/main/resources/lib/xp/event.ts
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * @param {object} params Listener parameters.
+ * @param {string} params.type Event type pattern.
+ * @param {function} params.callback Callback event listener.
+ * @param {boolean} params.localOnly Local events only (default to false).
+ */
+export function listener<EventData extends object = EnonicEventData>(params: ListenerParams<EventData>): void {
+    const helper = __.newBean<any>('com.enonic.xp.lib.event.EventListenerHelper');
+
+    helper.setType(params.type || '');
```
tajakobsen, line 108: Prefer `??` over `||` unless you are actually working with boolean operations.

This applies to the other cases in this file too.

```suggestion
    helper.setType(params.type ?? '');
```

modules/lib/lib-event/src/main/resources/lib/xp/event.ts
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * @example-ref examples/event/listener.js
+ *
+ * @param {object} params Listener parameters.
+ * @param {string} params.type Event type pattern.
+ * @param {function} params.callback Callback event listener.
+ * @param {boolean} params.localOnly Local events only (default to false).
+ */
+export function listener<EventData extends object = EnonicEventData>(params: ListenerParams<EventData>): void {
+    const helper = __.newBean<any>('com.enonic.xp.lib.event.EventListenerHelper');
```
tajakobsen, line 106: I think you should create interfaces to use with `__.newBean()`, because this is where the boundary between TypeScript and Java is. And therefore maybe the most "dangerous" place in the code. 

This might save you later (specially when it comes to nullability).

modules/script/script-impl/global.d.ts
```diff
@@ -0,0 +1,12 @@
...
+map imports via require or __non_webpack_require__.
+
+Warning: all of these interfaces are empty. If you want type definitions for various libraries,
+you need to add types for there libraries, where those interfaces are redefined.
+*/
+interface XpEventLibrary {}
+
+interface XpLibraries {
+    '/lib/xp/event': XpEventLibrary;
+}
```
tajakobsen, line 12: I think it's cleaner to just merge `XpLibraries` in `global` right in the library files, instead of going via `XpEventLibrary` and similar.

modules/script/script-impl/index.d.ts
```diff
@@ -0,0 +1,27 @@
...
+    debug: (...args: unknown[]) => void;
+    info: (...args: unknown[]) => void;
+    warning: (...args: unknown[]) => void;
+    error: (...args: unknown[]) => void;
+};
+
+declare const __: {
+    newBean: <T = unknown>(bean: string) => T;
+    toNativeObject: <T = unknown>(beanResult: T) => T;
+    toScriptValue: <T = Object>(value: T) => T;
```
tajakobsen, line 19:
```suggestion
    toScriptValue: <T = object>(value: T) => T;
```

modules/script/script-impl/index.d.ts
```diff
@@ -0,0 +1,27 @@
...
+    disposer: (callback: (...args: unknown[]) => unknown) => void;
+};
+
+declare const require: <Key extends keyof XpLibraries | string = string>(path: Key) =>
+    Key extends keyof XpLibraries ? XpLibraries[Key] : any;
+
+declare const __non_webpack_require__: <Key extends keyof XpLibraries | string = string>(path: Key) =>
+    Key extends keyof XpLibraries ? XpLibraries[Key] : any;
```
tajakobsen, line 27: Since `require` and `__non_webpack_require__` have the same shape, it could be better to declare it as a type, and reuse it.

```suggestion
type Require = <Key extends keyof XpLibraries | string = string>(path: Key) =>
    Key extends keyof XpLibraries ? XpLibraries[Key] : any;

declare const require: Require;

declare const __non_webpack_require__: Require;
```

modules/lib/lib-event/src/main/resources/lib/xp/event.ts
```diff
@@ -0,0 +1,136 @@
...
+
+/*
+ * Global library definitions for require and __non_webpack_import__
+ */
+
+type LibType = typeof Lib;
+
+declare global {
+    interface XpEventLibrary extends LibType {}
+}
```
tajakobsen, line 11: Have you considered just doing it like this:

```suggestion
declare global {
    // Will be merged
    interface XpLibraries {
      "/lib/xp/event": typeof import('./event')
    }
}
```

modules/script/script-impl/index.d.ts
```diff
@@ -0,0 +1,27 @@
+/// <reference path="global.d.ts" />
+
+declare const app: {
+    name: string;
+    version: string;
+    config: Record<string, string>;
```
tajakobsen, line 6: I have been thinking if `config` should just be `Record<string, string | undefined>`, to force the developer to handle the `undefined` case, because we are making a pretty big assumption that a given config always exists.

tajakobsen (changes requested):
Great work @edloidas ! I'm really excited to see this coming along so nicely!

---

I would like to recommend you start using "eslint" and "prettier" at this point too. Specially prettier will be a time saver for you, as you don't have to bother with manual formatting anymore.

---

Have you thought about how module imports with the `import` keyword will be supported?

```typescript
import { send } from "/lib/xp/event";
```

I'm not a huge fan of the `declare module "*/lib/xp/event"` hack I used in *enonic-types*, but it sort of works at least.

edloidas:
@tajakobsen Thanks for all the input you gave here. You've also made an impressive work adding type definitions for all our libraries.

I just want you to consider all those changes as pre-alpha, since there is still some code, that I'm heavily experimenting right now to make it work in different scenarios, some, that is not yet committed.

We'll definitely add  ESLint and Prettier before the release, when the code became more stable. Right now, when everything is constantly changing it only gets in the way. Same goes for comments, documentation and beans definitions, while I'm still experimenting on `lib-event` only.

The idea behind moving definitions to XP, was, firstly, in adding types to our own code and tests, secondly, in automatic generation and publishing of corresponding types after changes in the libs, thirdly, moving as much of TS under the hood as possible, so the developers will only need to install the dependencies, add basic `tsconfig.json` and change file extensions to start using them. Further configuration should also be as simple as possible.

As for the `import` usage, I'm considering suggestion to use of [externals](https://webpack.js.org/configuration/externals/) option in webpack, and similar options while building with other tools, like it's done with Node modules.

 I'll try to answer on all your comments later.

modules/lib/lib-event/package.json
edloidas: Sure. As I said, ESLint and prettier will definitely be added a bit later. Right now code is to unstable and constantly changes.

modules/script/script-impl/index.d.ts
edloidas: Good catch. That was serious misstype that I missed.

modules/lib/lib-event/src/main/resources/lib/xp/event.ts
edloidas: As you probably saw, function code was not yet modified. I've only added some type. Even the beans are typed with `any`. Will work on this after the lib format if finalized and linter is added. So, don't you worry.

modules/script/script-impl/global.d.ts
edloidas: Yes, definitely. Guess, I overcomplicated it.

modules/lib/lib-event/src/main/resources/lib/xp/event.ts
edloidas: Yup, it's for the completion. Since we expand the type with `string`, that will lead the editors just omit the more narrow types, if just the `|` operator is used.
Will add a comment. I'm just not sure if it'll stay here or be moved to other place, so that such tricks can be used across other libraries.
But you are right, adding at least `// TODO` will be a good thing so I don't forget to properly comment it later.

modules/lib/lib-event/src/main/resources/lib/xp/event.ts
edloidas: Sure, will look. BTW, this part is basically the copy from `enonic-types`, so I haven't dig into it yet.

modules/script/script-impl/index.d.ts
edloidas: Yeah, saw that too. Haven't committed yet.

modules/lib/lib-event/src/main/resources/lib/xp/event.ts
edloidas: Yes, replaced it. Thanks.

modules/lib/lib-event/src/main/resources/lib/xp/event.ts
tajakobsen: This is what I expected. When I started *enonic-types* I used a lot of `readonly`, because my projects use a lot of "defensive copying" (no mutation of objects).

Since this isn't everybodys style, I have removed a lot of `readonly` since then.

But since you guys are starting a bit fresh, it would be worth creating some "rules" for where you wish to use it.

build.gradle
```diff
@@ -30,6 +31,8 @@ configure( rootProject ) {
     apply from: "$rootDir/gradle/tools.gradle"
 }
 
+// TODO: Configure all lib projects here, applying "$rootDir/gradle/typescript.gradle", when moved to TS
```
rymsha, line 34: this an only be merged to master when all type definitions are ready.

gradle/typescript.gradle
```diff
@@ -0,0 +1,40 @@
...
+    workDir = file("${rootProject.projectDir}/.gradle/nodejs")
+}
+
+task lint( type: NpmTask, dependsOn: npmInstall ) {
+    description = 'Lint TS, JS, and JSON'
+    args = [ 'run', 'lint:quiet' ]
+    outputs.upToDateWhen { false }
+}
+
+task typescript( type: NpmTask, dependsOn: [ npmInstall, lint ] ) {
```
rymsha, line 16: typescript does not depend on lint
also don't make more tasks than needed.
typescript, copyDefinitions, definitions seem like good candidates for just one npm task

gradle/typescript.gradle
```diff
@@ -0,0 +1,40 @@
+apply plugin: 'com.github.node-gradle.node'
+
+node {
+    download = true
+    distBaseUrl = null
+    version = '16.15.0'
+    workDir = file("${rootProject.projectDir}/.gradle/nodejs")
+}
+
+task lint( type: NpmTask, dependsOn: npmInstall ) {
```
rymsha, line 10: lint must be part of clean task (clean depends on lint)

modules/lib/lib-event/.eslintrc.js
```diff
@@ -0,0 +1,83 @@
+module.exports = {
```
rymsha, line 1: it should be common for all libs

modules/lib/lib-event/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Global XP types
+
+You may also want to add the `@enonic/script-impl` types, since it provides a definitions for XP global objects, e.g. `log`, `app`, etc. The
+documentation can be found [here](https://github.com/enonic/xp/tree/master/modules/script/script-impl/README.md).
+
+## Example
+
+```typescript
+const eventLib, {EnonicEvent} = __non_webpack_require__('/lib/xp/event');
```
rymsha, line 35: `__non_webpack_require__` is heavily dependent on webpack. We must support not only webpack

modules/lib/lib-event/package.json
```diff
@@ -0,0 +1,53 @@
...
+        "event",
+        "types",
+        "typescript"
+    ],
+    "author": {
+        "name": "Mikita Taukachou",
+        "email": "edloidas@gmail.com",
+        "url": "https://edloidas.github.io/"
+    },
+    "license": "GPL-3.0",
```
rymsha, line 41: library licenses are Apache

modules/lib/lib-event/package.json
```diff
@@ -0,0 +1,53 @@
+{
+    "name": "lib-event",
+    "version": "0.1.0",
```
rymsha, line 3: what is this version? is it goint to be a version of types on npm? than it is hard to know what it works with XP 7.10

modules/lib/lib-event/src/main/resources/lib/xp/event.ts
```diff
@@ -0,0 +1,143 @@
...
+    readonly id: string;
+    readonly path: string;
+    readonly branch: string;
+    readonly repo: string;
+    readonly newPath?: string; // event type <'node.moved' | 'node.renamed'>
+}
+
+type LiteralUnion<T extends U, U = string> = T | (U & Record<never, never>);
+
+export type EnonicEventTypes =
```
rymsha, line 64: This name implies that Enonic XP only issues such events. Which is (soon) won't be true.
It affects all related types, too.
Either drop this handy thing, or rename the type(s). `EnonicNodeEventType`?

modules/lib/lib-event/src/main/resources/lib/xp/event.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+interface EventListenerHelper {
+    setType(type: string): void;
+    setListener<T = unknown>(callback: (event: T) => void): void;
+    setLocalOnly(localOnly: boolean): void;
+    register(): void;
+}
+
+/**
```
rymsha, line 93: where do the jsdock land now?

modules/lib/lib-event/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+    "compilerOptions": {
+        "lib": ["es6"],
```
rymsha, line 3: why is this es6?

modules/script/script-impl/README.md
```diff
@@ -0,0 +1,79 @@
+# Enonic XP global TS types
```
rymsha, line 1: script-impl is an implementation detail module.
You probably meant to put it into script-api

modules/script/script-impl/global.d.ts
```diff
@@ -0,0 +1,18 @@
+/**
```
rymsha, line 1: should probably be part of script-api module

modules/script/script-impl/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "global",
+    "types",
+    "typescript"
+  ],
+  "author": {
+    "name": "Mikita Taukachou",
+    "email": "edloidas@gmail.com",
+    "url": "https://edloidas.github.io/"
+  },
+  "license": "GPL-3.0",
```
rymsha, line 29: Apache

settings.gradle
```diff
@@ -18,6 +19,28 @@ dependencyResolutionManagement {
     }
 }
 
+dependencyResolutionManagement {
```
rymsha, line 22: i guess, instead node plugin should be provided directly in every module

rymsha (changes requested):

build.gradle
edloidas: Should I remove this comment then? Maybe move it to an issue?

gradle/typescript.gradle
edloidas: Okay. Does `typescript` sound like a good name for all of that?

gradle/typescript.gradle
edloidas: I completely don't understand, how `lint` task is a dependency for `clean`. We must `lint` the code before we build/publish/bundle something. But `clean` is not necessary the task, that is run before one of those.

modules/lib/lib-event/.eslintrc.js
edloidas: Of course. I would like to move the definitions to our own `@enonic/eslint-recommended` package and use it later as a plugin, so the only thing that we need to do is to write `module.exports = {'extends': ['plugin:@enonic/eslint-recommended']}`. Actually, I needed to be done long ago, since it's a pain in the back to sync all the ESLint configs across our apps. As for the file itself, I would rather keep it inside the package folder, along with the `package.json` and `tsconfig.json`.

modules/lib/lib-event/README.md
edloidas: You are correct, it's a webpack-related stuff. Will replace it with `require`. 
Since we may want to keep code more generic, in the `base` (`script-impl` types), I'll remove the `__non_webpack_require__` from global definitions, and add it as an example/note in the readme instead. I'll also move the actual code to `starter-webpack`.

modules/lib/lib-event/package.json
edloidas: Got it. `Apache-2.0`

modules/lib/lib-event/package.json
edloidas: `0.1.0` will be first version and meant to be used in other projects and tested. We may need to do additional changes before the actual release. The final version will follow the XP versioning.

modules/lib/lib-event/src/main/resources/lib/xp/event.ts
edloidas: Actually, it implies, that the `lib-event` can have any user-defined `string` event, but we also provide a list of [well-known events](https://developer.enonic.com/docs/xp/stable/api/lib-node#events) for autocompletion. I can see that we can also add `'node.manualOrderUpdated'`, that is not listed there.

modules/lib/lib-event/src/main/resources/lib/xp/event.ts
edloidas: I may have forgot to test it. Can you point me what should be generated and what tasks do we use for that?

modules/lib/lib-event/tsconfig.json
edloidas: Doesn't Nashorn support at least `find` and `Map` features? Just a plain old ES5 with `const`? If not, then it must be `es5`, I guess.

modules/script/script-impl/README.md
edloidas: Not really. This definitions, that will actually be called `base`, are not tightly coupled with some specific module. The reason why I put it there was  actually because of the implementation, for the developer after changing or adding something like [this pre-init code](https://github.com/enonic/xp/blob/5c6b85b194035f2b2e21f036b8909baa8747460e/modules/script/script-impl/src/main/java/com/enonic/xp/script/impl/executor/ScriptExecutorImpl.java#L40) shouldn't forget to add types for it.

modules/script/script-impl/global.d.ts
edloidas: Commented above.

modules/script/script-impl/package.json
edloidas: Done.

settings.gradle
edloidas: Do you mean we should remove node plugin from `typescript.gradle` and apply it in the `build.gradle` in each library instead to get rid of those hacks with `dependencyResolutionManagement`?

modules/lib/lib-event/tsconfig.json
alansemenov: > Doesn't Nashorn support at least find and Map features?

No :(

edloidas: Okay, will change it to "es5" then.

modules/script/script-impl/README.md
alansemenov: Maybe it should be called `common`, not `base`?

edloidas: I don't like the name `common` because we don't provide some common types there, that can be used across apps. We provide definitions for the core properties of the global object: `__`, `log`, etc. I would not name it `core`, because the name implies, that we need it to use with other types, which is not actually true.

edloidas: `global`, `base`, `common`, `core`, `main`, `general` are actually names, that I was thinking about. They are abstract, each and every of them have their downside. But any of them will do anyway.

alansemenov: `global` is good.
@rymsha any objections?

gradle/typescript.gradle
rymsha: ok with me.

gradle/typescript.gradle
rymsha: linting does not do anything with code - thus, not affecting build: you can still build un-linted code.

build.gradle
edloidas: Sure. I'm just want to make sure we won't forget about this after the epic is finished, because it's not probably me, who will add types to other libraries.

modules/script/script-impl/README.md
rymsha: This is a tough one... but "log, require, resolve, __, exports, module" are still part of our script-api. I'll have to thing about it more...

gradle/typescript.gradle
edloidas: Initially, in our app's repos, `build` depends on `lint` to force our developers to fix any code problems before the changes are published. Rather controversial decision. I can see t does not fit the workflow of the XP. Do you want to keep `lint` as an independent task and run it before `test` on CI? No linting, unless it is explicitly run? (I just can't see how `clean` and `lint` tasks can be related).

edloidas: Ah, so you mean `check`, not `clean`! That makes sense. Was a bit confused, figuring out why do we need to run `lint` before `clean`.

settings.gradle
edloidas: Made a bit of adjustments, but the [core problem](https://github.com/node-gradle/gradle-node-plugin/blob/master/docs/faq.md#is-this-plugin-compatible-with-centralized-repositories-declaration) hasn't gone. If the node download option is set:
```gradle
node {
    download = true
    version = '16.15.0'
}
```
We either get problems with adding a repository for the Node to download, or the newly added repository overrides existing one (default mode or explicitly setting `RepositoriesMode.PREFER_PROJECT` mode). I don't see a way to remove the `ivy` repo from `dependencyResolutionManagement` right now. Maybe you can take a look.

gradle/typescript.gradle
rymsha: sorry. `check`, not `clean`.

modules/lib/lib-event/src/main/resources/lib/xp/event.ts
rymsha: I would prefer EnonicNodeEventType | EnonicContentEventType | string
(EnonicContentEventType is not-existent just yet ,but its coming)

edloidas: Will replace with a `string`

modules/script/script-impl/README.md
edloidas: Will move code to `lib/typescript`. The package will be named `global`.

modules/lib/typescript/index.d.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+declare const log: {
+    debug: (...args: unknown[]) => void;
+    info: (...args: unknown[]) => void;
+    warning: (...args: unknown[]) => void;
+    error: (...args: unknown[]) => void;
+};
+
+declare const Java: {
```
rymsha, line 16: This is Nashorn stuff (internal use). Can we avoid having it here?

edloidas: Sure. Moved to root's `global.d.ts`, that will be used in compile-time and is not a part any package that will be published.

modules/lib/typescript/index.d.ts
```diff
@@ -0,0 +1,29 @@
+/// <reference path='global.d.ts' />
+
+declare const app: {
+    name: string;
+    version: string;
+    config: Record<string, string | undefined>;
+};
+
+declare const log: {
```
tajakobsen, line 9: There might be an opportunity to add a little bit of documentation regarding logging as JSDoc here.
Maybe some of the info from here: https://developer.enonic.com/docs/xp/stable/framework/globals#log

tajakobsen (commented):
This is coming along nicely. I'm just posting a few comments.

rymsha (approved):

Issue #9447, opened 2022-04-25
Child contents pop up to the root after syncAll process


vbradnitski:

rymsha:
This happens due to missing node blob `store` call.
When layers were implemented, `move` command became trickier, and could modify node data (to reset inherit flags).
Sadly enough old code (from 2015) assumed that `move` is never changing node data and only updated index skipping blob update. This was done to gain _marginal_ performance increase.

Calling blob store for move operation should solve the problem.
Unfortunately all nodes that were moved (for instance, archived) before the fix have inconsistency between their index value and blob store value. Meaning that syncAll for them will still be problematic. Another problem as for "reindex" also restores values from blob and will have the same effect as "syncAll"

Good to know that this bug only affects moved nodes.

PR #9454, opened 2022-04-27
Child contents pop up to the root after syncAll process #9447

vbradnitski:


jsi (approved):
Same as to master.  I will approve this.

Issue #9468, opened 2022-05-12
Content changes in parent project are not synced to layer after restore from archive


alansemenov:
1. Start with a project that has a child layer.
2. Create a site in the project. 
3. Switch to the layer - the site is there. All is good is far.
4. Archive the site **in the layer**
5. Switch back to the project. Create a new content under the site.
6. Switch back to the layer. Nothing is synced. Still correct.
7. Go to the layer's archive and restore the site.

The expected behaviour is that the new items added to the parent site would be synced to the site in the layer after it has been restored. But instead there's an error in the logs:

```
10:05:38.806 ERROR c.e.x.c.i.c.ProjectContentEventListener - Project node sync failed
java.lang.IllegalArgumentException: targetContent must be set.
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:141)
	at com.enonic.xp.core.impl.content.UpdatedEventSyncCommand$Builder.validate(UpdatedEventSyncCommand.java:193)
	at com.enonic.xp.core.impl.content.UpdatedEventSyncCommand$Builder.build(UpdatedEventSyncCommand.java:199)
	at com.enonic.xp.core.impl.content.ParentContentSynchronizer.lambda$new$5(ParentContentSynchronizer.java:72)
	at com.enonic.xp.core.impl.content.ParentContentSynchronizer.createEventCommand(ParentContentSynchronizer.java:305)
	at com.enonic.xp.core.impl.content.ParentContentSynchronizer.sync(ParentContentSynchronizer.java:188)
	at com.enonic.xp.core.impl.content.ProjectContentEventListener.lambda$doHandleContentEvent$11(ProjectContentEventListener.java:177)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:177)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
[...]

PR #9481, opened 2022-05-27
Content changes in parent project are not synced to layer after restore from archive #9468 

vbradnitski:
…re from archive #9468


modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/DeletedEventSyncCommand.java
```diff
@@ -48,7 +48,8 @@ private boolean isToSyncDelete( final Content targetContent )
 
     private boolean needToDelete( final ContentToSync content )
     {
-        return content.getSourceContext().callWith( () -> !contentService.contentExists( content.getTargetContent().getId() ) ) &&
+        return content.getSourceContent() == null &&
```
anatol-sialitski, line 51: `content.getSourceContent() != null`

anatol-sialitski (approved):

rymsha (approved):

Issue #9512, opened 2022-06-16
`lib-grid` fails with NPE


anatol-sialitski:
The `convertValue` method must check a value for null for Array and Object.

PR #9513, opened 2022-06-16
lib-grid fails with NPE #9512

anatol-sialitski:


modules/lib/lib-grid/src/main/resources/lib/xp/grid.js
```diff
@@ -9,9 +9,11 @@ function requireNotNull(value, parameterName) {
 
 function convertValue(value) {
     if (Array.isArray(value)) {
-        return __.toScriptValue(value).getList();
+        var array = __.toScriptValue(value);
+        return array !== null ? array.getList() : null;
```
rymsha, line 13: Overkill, isn't it?

rymsha (approved):

Issue #7571, opened 2019-10-23
HtmlLinkProcessor: correctly process links with query parameters


ashklianko:
Links like ~`content://123456#insert-tab` or `content://123456?a=1&b=2`~ `content://<id>?query=<query>&fragment=<fragment>` should be correctly resolved.

`<query>` - `encodeuricomponent("key1=value1&key2=value2")` - value will be decoded and added to resulting link as query. If decoded string contains any invalid query symbols the entire query will be ignored.

`<fragment>` - `encodeuricomponent("some-fragment")` - value will be decoded and added to resulting link as fragment. If decoded string contains any invalid fragment symbols the entire fragment will be ignored.

resulting url will be (based on exampples)
`http://bla-bla-bla?key1=value1&key2=value2#some-fragment`

rymsha:
duplicate of #7035 ?

alansemenov:
https://discuss.enonic.com/t/anchor-link-in-the-link-dialog-box/2298

alansemenov:
Client-side validation task: https://github.com/enonic/app-contentstudio/issues/1458

alansemenov:
We also need a task in Content Studio for adding a new field on the "Content" tab for whatever will be appended to the content url (hash, params etc.)

rymsha:
Also occupying query params and anchor in content urls "just" for direct transfer to http urls is probably  squander

alansemenov:
NB! UI will url encode values in the anchor & query string fields before saving them.

rymsha:
NB! Doing it just in UI is not secure enough.

larserikpedersen:
Hi Sergey

Am I supposed to receive these, or is it meant for another Lars Erik?


Med vennlig hilsen

Lars Erik Pedersen
innholdsredaktør
avdeling utvikling og digitale kanaler
divisjon digitalisering og helseregister
mobil: 466 73 423

***@***.***

Postadresse: Pb 220 Skøyen, 0213 Oslo
Besøksadresse: Vitaminveien 4, 0483 Oslo
https://helsedirektoratet.no<https://helsedirektoratet.no/>  https://helsenorge.no/

P Tenk på miljøet; ikke skriv ut denne e-posten hvis det ikke er helt nødvendig



Fra: Sergey Rymsha ***@***.***>
Sendt: torsdag 23. juni 2022 10:25
Til: enonic/xp ***@***.***>
[...]

PR #9521, opened 2022-06-27
#7571 Correctly process links with query parameters

anatol-sialitski:


rymsha (changes requested):
this is not according to spec.

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/HtmlLinkProcessor.java
```diff
@@ -278,4 +282,31 @@ private Map<String, String> extractUrlParams( final String urlQuery )
...
+            return url;
+        }
+        final String query = urlQuery.startsWith( "?" ) ? urlQuery.substring( 1 ) : urlQuery;
+
+        final StringBuilder urlSuffix = new StringBuilder();
+        final Map<String, String> queryParamsAsMap = extractUrlParams( query );
+
+        if ( queryParamsAsMap.containsKey( "query" ) )
+        {
+            urlSuffix.append( "?" ).append( decodeParameterValue( queryParamsAsMap.get( "query" ) ) );
```
rymsha, line 298: you should check that query does not contain illegal characters after `decodeParameterValue( queryParamsAsMap.get( "query" )`

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/HtmlLinkProcessor.java
```diff
@@ -278,4 +282,31 @@ private Map<String, String> extractUrlParams( final String urlQuery )
...
+        }
+        if ( queryParamsAsMap.containsKey( "fragment" ) )
+        {
+            urlSuffix.append( "#" ).append( decodeParameterValue( queryParamsAsMap.get( "fragment" ) ) );
```
rymsha, line 302: you should check that query does not contain illegal characters after `decodeParameterValue( queryParamsAsMap.get( "fragment" )`

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/HtmlLinkProcessor.java
```diff
@@ -122,7 +125,8 @@ public String process( final String text, final String urlType, final PortalRequ
                             type( urlType ).
                             id( id ).
                             portalRequest( portalRequest );
-                        final String pageUrl = portalUrlService.pageUrl( pageUrlParams );
+                        final String pageUrl = addQueryParamsIfPresent( portalUrlService.pageUrl( pageUrlParams ), urlParamsString );
+
```
rymsha, line 129: resulting pageUrl must be html (attribute) encoded

rymsha (changes requested):

vbradnitski (approved):

Issue #9453, opened 2022-04-26
Text component should not be wrapped inside a div in edit mode


alansemenov:
Contents of a text component are wrapped with a `<section>` tag - which is fine - but in edit mode this `<section>` tag is additionally wrapped with a `<div>` with attribute `data-portal-component-type="text"`. This is unnecessary, attribute should be set directly on the `<section>` element.

rymsha:
related to #9395

PR #9522, opened 2022-06-28
Text component should not be wrapped inside a div in edit mode #9453

rymsha:
- removed extra `div` in EDIT mode
In addition empty text component renders the same (empty string) in PREVIEW and INLINE


alansemenov (approved):

ashklianko (approved):

ashklianko:
Frontend is ready

sigdestad:
FYI guys. We (me and PMI) tried to remove this earlier, and it caused havoc in the Next.js project, so this has to be verified carefully before going live. In some cases it may be breaking afaik?

sigdestad:
OFC, textcomponent should ideally not be used for headless approaches.

Issue #9546, opened 2022-07-04
Migrate Javascript API to Typescript for `lib-auth`


anatol-sialitski:
Migrate Javascript API to Typescript for `lib-auth`

PR #9548, opened 2022-07-05
Migrate Javascript API to Typescript for lib-auth #9546

anatol-sialitski:


modules/lib/lib-auth/src/main/resources/lib/xp/auth.ts
```diff
@@ -0,0 +1,832 @@
...
+    modifiedTime: string;
+}
+
+export interface User
+    extends Principal {
+    disabled?: boolean;
+    email?: string;
+    login?: string;
+    idProvider: string;
+    profile?: Record<string, Object>;
```
edloidas, line 46: Just replace it with `object` type, because it is pretty much similar to the current type. `Record` should be used, when we limit available list of keys or values type, e.g. `Record<'foo' | 'bar', number>`

modules/lib/lib-auth/src/main/resources/lib/xp/auth.ts
```diff
@@ -0,0 +1,832 @@
...
+    scope?: string | null;
+    editor: (group: Group) => Group;
+}
+
+interface ModifyProfileHandler {
+    setKey(value: string): void;
+
+    setScope(value?: string | null): void;
+
+    setEditor(value: Object): void;
```
edloidas, line 678: All editors are not `Object`. Actually, it must be a Function, since we use `editor.call(/*...*/)` in Java.
Use `Function` for this purpose.
Or we can go even further and define an editor type on top of the file:
```ts
export type EditorFn = <T>(value: T) => T
```
Later we can do something like this:
```ts
editor: EditorFn<Group>;
// ...
setEditor(value: EditorFn<Group>): void;
```

edloidas (changes requested):
Look through the comments, please. Also, don't forget that the are bunch of similar problems and I haven't marked them all.

modules/lib/lib-auth/src/main/resources/lib/xp/auth.ts
anatol-sialitski: Replaced by `Record<string, unknow>`

edloidas (approved):

Issue #9552, opened 2022-07-06
Migrate Javascript API to Typescript for `lib-scheduler`


anatol-sialitski:
Migrate Javascript API to Typescript for `lib-scheduler`

PR #9553, opened 2022-07-06
Migrate Javascript API to Typescript for lib-scheduler #9552

anatol-sialitski:


modules/lib/lib-scheduler/src/main/resources/lib/xp/scheduler.ts
```diff
@@ -0,0 +1,229 @@
...
+    extends OnTimeSchedule {
+    timeZone: string;
+}
+
+export interface ScheduleJob {
+    name: string;
+    descriptor: string;
+    description: string;
+    enabled: boolean;
+    config?: Record<string, Object> | null;
```
edloidas, line 55: Replace `Record<string, Object>` with `object` here and in other places.

modules/lib/lib-scheduler/src/main/resources/lib/xp/scheduler.ts
```diff
@@ -0,0 +1,229 @@
...
+
+export interface ModifyScheduledJobParams {
+    name: string;
+    editor: (schedule: ScheduleJob) => ScheduleJob;
+}
+
+interface ModifyScheduledJobHandler {
+    setName(value: string): void;
+
+    setEditor(value: Object): void;
```
edloidas, line 128: Look to the previous PR and see how we can type and replace this with `EditorFn<T>`

edloidas (changes requested):

modules/lib/lib-scheduler/src/main/resources/lib/xp/scheduler.ts
anatol-sialitski: Thank you for your suggestion!

edloidas (approved):

Issue #9294, opened 2022-01-05
Sanitize HtmlArea on save. Server side


rymsha:
confirm with CKEEditor

PR #9598, opened 2022-08-01
Sanitize HtmlArea on save. Server side #9294

vbradnitski:


modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/processor/HtmlAreaContentProcessor.java
```diff
@@ -236,9 +255,14 @@ private Set<String> getPaths( final FormItems formItems )
 
     private void processDataTree( final Collection<Property> properties, final ContentIds.Builder processedIds )
     {
-        properties.stream().
-            map( Property::getString ).
-            forEach( value -> processString( value, processedIds ) );
+        properties.forEach( property -> {
+            final String processedValue = processString( property.getString(), processedIds );
+
+            if ( !property.getString().equals( processedValue ) )
```
rymsha, line 261: not needed?

modules/lib/lib-portal/src/main/java/com/enonic/xp/lib/portal/SanitizeHtmlHandler.java
```diff
@@ -17,6 +17,6 @@ public String sanitizeHtml( final String html )
     @Override
     public void initialize( final BeanContext context )
     {
-        this.htmlSanitizer = context.getService( HtmlSanitizer.class ).get();
+        this.htmlSanitizer = context.getService( HtmlSanitizer.class, "(type=function)" ).get();
```
rymsha, line 20: Introduce a new method instead

rymsha (changes requested):

PR #9600, opened 2022-08-01
Sanitize HtmlArea on save. Server side #9294

vbradnitski:


modules/portal/portal-api/src/main/java/com/enonic/xp/portal/owasp/SanitizeType.java
```diff
@@ -0,0 +1,6 @@
+package com.enonic.xp.portal.owasp;
+
+public enum SanitizeType
+{
+    FUNCTION, PROCESSOR
```
rymsha, line 5: `FUNCTION` - probably should be called `STRICT`
`PROCESSOR` - probably should be called `RICH_TEXT`

modules/portal/portal-owasp/src/main/java/com/enonic/xp/portal/owasp/impl/HtmlRichTextSanitizer.java
```diff
@@ -0,0 +1,98 @@
...
+final class HtmlRichTextSanitizer
+{
+    private static final String[] ALLOWED_ELEMENTS =
+        new String[]{"strong", "address", "blockquote", "code", "dd", "dl", "dt", "em", "u", "p", "br", "ol", "li", "ul", "td", "th", "tr",
+            "h1", "h2", "h3", "h4", "h5", "h6", "div", "pre", "strong", "a", "img", "figure", "figcaption", "table", "caption", "tbody",
+            "thead", "tfoot"};
+
+    private final HtmlPolicyBuilder htmlSanitizePolicy;
+
+    @Activate
```
rymsha, line 21: No need to Activate
And can be package-local

rymsha (approved):

Issue #9464, opened 2022-05-04
Regexp is validated only in the first occurrence of an input


alansemenov:
With this content type schema

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<content-type>
  <display-name>Blog</display-name>
  <super-type>base:structured</super-type>
  <form>
    <input name="url" type="TextLine">
      <label>URL</label>
      <occurrences minimum="0" maximum="0"/>
      <config>
        <regexp>^http(s)?:\/\/.?(www\.)?[a-zA-Z0-9][-a-zA-Z0-9@:%._\+~#=]{0,255}\b([-a-zA-Z0-9@:%_\+.~#?&amp;//=]*)</regexp>
      </config>
    </input>
  </form>
</content-type>
```

XP will give an error when trying to save URL `https://www.oslo.kommune.no/barnehage/finn-barnehage-i-oslo/#!c%7Cf_preschool_type_student/c.f_preschool_type_student//m.list` in the first input occurrence, but will let it be saved in the second occurrence.

<img width="954" alt="image" src="https://user-images.githubusercontent.com/11612490/166817785-03f19925-822d-4c5a-a32c-3b17baf204ff.png">

<img width="1217" alt="image" src="https://user-images.githubusercontent.com/11612490/166818046-3a130883-9aae-4a2a-ac37-6e96ef525e52.png">

PR #9615, opened 2022-08-05
Regexp is validated only in the first occurrence of an input #9464

anatol-sialitski:


modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/validate/InputValidationVisitor.java
```diff
@@ -26,7 +26,9 @@ public void visit( final Input input )
         final Property property = propertyTree.getProperty( input.getPath().toString() );
         if ( property != null )
         {
-            checkValidity( input, property );
+            property.getParent().getPropertyArray( input.getName() ).
```
rymsha, line 29: use `PropertySet.getProperties`

rymsha (changes requested):

rymsha (approved):

vbradnitski (approved):

Issue #9661, opened 2022-08-29
Schema Admin role


rymsha:
User with Developer role should be able to create/delete virtual apps and create/modify/delete content

Role should be created on XP startup, if missing

PR #9671, opened 2022-09-02
Developer role #9661

vbradnitski:


rymsha (changes requested):
Need to move role init into initializer of app repository

rymsha (changes requested):
Rename Developer everywhere 

Also where is a test to check that without Schema Admin role no apps or schemas can be managed?

rymsha (approved):

anatol-sialitski (approved):

Issue #9659, opened 2022-08-29
verify attachment checksum on modification of media content 


alansemenov:
When a binary attachment in an existing media content gets modified (f.ex. re-uploaded), XP won't compare checksums of the old and the new attachments which results in the content being modified every time, even when there have been no changes.

PR #9672, opened 2022-09-02
Modification of media content won't verify attachment checksum #9659

rymsha:


modules/core/core-api/src/main/java/com/enonic/xp/content/ResetContentInheritParams.java
```diff
@@ -68,7 +68,7 @@ public Builder projectName( ProjectName projectName )
 
         public Builder inherit( Collection<ContentInheritType> inherit )
         {
-            this.inherit = EnumSet.copyOf( inherit );
+            this.inherit = inherit.isEmpty() ? EnumSet.noneOf( ContentInheritType.class ) : EnumSet.copyOf( inherit );
```
anatol-sialitski, line 71: `inherit == null || inherit.isEmpty()`

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/UpdateNodeParamsFactory.java
```diff
@@ -172,7 +199,11 @@ Builder contentDataSerializer( final ContentDataSerializer contentDataSerializer
 
         void validate()
         {
-            Preconditions.checkNotNull( params );
+            Preconditions.checkNotNull( modifier, "modifier cannot be null" );
+            Preconditions.checkNotNull( editedContent, "contentId cannot be null" );
```
anatol-sialitski, line 203: `content cannot be null` instead of `contentId cannot be null`

anatol-sialitski: or `editedContent cannot be null`

vbradnitski (approved):

anatol-sialitski (approved):

PR #9678, opened 2022-09-06
Fix #9677 portal.getComponent return type object is very loose

### LINKED ISSUE #9677: portal.getComponent return type object is very loose
Screenshot 2022-09-06 at 10 16 28

ComLock:


modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -424,6 +424,25 @@ interface GetCurrentComponentHandler {
...
+
+export type Component<
+  Config extends object = object,
+  Regions extends Record<string,Region> = Record<string,Region>
+> = {
+  config :Config
+  descriptor :string
+  path :string
+  type :'page'|'layout'|'part'
+  regions :Regions
```
alansemenov, line 436: We don't use this kind of annotation in XP. It's always "`prop: propType`", not "`prop :propType`".

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -424,6 +424,25 @@ interface GetCurrentComponentHandler {
...
+}
+
+export type Region<
+  Config extends object = object
+> = {
+  components :Array<Component<Config>>
```
alansemenov, line 442: Same here

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -432,9 +451,12 @@ interface GetCurrentComponentHandler {
  *
  * @returns {object} The current component as JSON.
  */
-export function getComponent(): object {
+export function getComponent<
+  Config extends object = object,
+  Regions extends Record<string,Region> = Record<string,Region>
+>() :Component<Config,Regions> {
```
alansemenov, line 457: And here.

alansemenov (changes requested):

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
ComLock: I do this for multiple reasons. One is color in the editor.
Screenshot 2022-09-07 at 11 52 33

ComLock: The top one just looks better... And is easier to read

ComLock: Also when searching I can differ between searching js objects, and ts types.

alansemenov: Christian, we are not going to have two different types of annotation in one project. It's ok for Explorer since you used this as a standard there. Also you can surely setup highlighting differently in your IDE, so that's not an argument.

edloidas: I agree with @alansemenov . It's okay to have your preferences in one-man projects (that's what we all have). But XP is a large project and we have company code style here. As for the "just looks better" - it's a matter of taste and color scheme. In my color scheme I have a different colors for `:`, `?` and other symbols, because it's easier for ME to find those symbols, e.g.:
image

edloidas (changes requested):

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
edloidas: Same goes for the spaces, semicolons. If you are using the Intellij IDEA, use this settings (though they might be kinds old)
https://github.com/enonic/dev-tools/blob/master/codestyles/idea/Enonic.xml

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
anatol-sialitski: Also, we use `[]` in XP instead of `Array`

edloidas: And, if I'm not mistaken, in XP we prefer `interface` instead of `type`. Again, and matter of project's code style.

edloidas:
@ComLock not related to the PR or issue itself, but since you've mentioned, it would be great, if you could give more details on namespaces: how do you see applying them and how it would improve the type definitions. Feel free to do it an a separate issue or in Slack.

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
ComLock: This one actually matters, interfaces can be monkeyPatched, types cannot.
So strange things can happen upstream or downstream with interfaces...

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
ComLock: Turns out switching syntax theme in Atom was easy (I probably already did before)
Screenshot 2022-09-07 at 13 45 45

alansemenov: Great! Can we now change it in Explorer please? Just kidding 😀

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
edloidas: Are you talking about the declaration merging? Like, when declaring interface with the same name, they will be combined? If it's what you are talking about, then I can assure you, using `interface` is pretty much safe here. Yes, we have globally declared interfaces, where declaration merging is intentional. 
As for the imported stuff, then merging by declaring a local interface/namespace/etc. of the same name won't work. This is actually an error and the compiler will say, that `"Import declaration conflicts with local declaration ..."`. 
Simply, this will NOT work:
```ts
import {EnonicEvent} from '/lib/xp/event';

interface EnonicEvent {
    foo: string;
}
```
But this will:
```ts
import {EnonicEvent} from '/lib/xp/event';

declare module '/lib/xp/event' {
    interface EnonicEvent {
        foo: string;
    }
}
```
In the second version there is a bit of work done, so it's hardly possible to change the interface unintentionally.

ComLock: Maybe this has been "fixed" in newer versions of TypeScript. I don't remember the details about what I read anymore, but I had some problems, so I started using type instead, and all my problems went away.

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
ComLock: Fixed

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
edloidas: Yeah, probably. It's easy to miss changes, unless you specifically read every release notes.
Anyway, I would rather have `interface` here for the sake of code consistency. 
P.S.: There are things to improve in our definitions (like adding `readonly` to some props), but it's just not crucial and we don't have much time for it right now. Though style discussions, suggestions and PRs are pretty much welcome (like that with namespace), because there aren't many people that will use them in Enonic.

ComLock: Readonly typically just causes a lot of headache, and since the code is synchronous, if you modify a property after it is passed to a function, it's not modified before that function has returned anyway? Or am I missing something?

edloidas: Yeah, that's right. Shouldn't probably mentioned this, because it is not related to the current PR. 
But we had a discussion with Tom, who created types for Enonic libs previously in [enonic-types](https://github.com/ItemConsulting/enonic-types), where he had a lot of `readonly` for encouraging people avoid mutable code. We decided not to this right now, but look through all the code later to see if we must really avoid mutations in some places.

alansemenov (approved):

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -432,9 +451,12 @@ interface GetCurrentComponentHandler {
...
  * @returns {object} The current component as JSON.
  */
-export function getComponent(): object {
+export function getComponent<
+  Config extends object = object,
+  Regions extends Record<string,Region> = Record<string,Region>
+>(): Component<Config,Regions> {
     const bean = __.newBean<GetCurrentComponentHandler>('com.enonic.xp.lib.portal.current.GetCurrentComponentHandler');
-    return __.toNativeObject(bean.execute());
+    return __.toNativeObject(bean.execute()) as Component<Config,Regions>;
```
anatol-sialitski, line 459: I suggest  to change output type in `GetCurrentComponentHandler` for the `execute` function to `Component<Config,Regions>`. To avoid type casting here.

edloidas (approved):

rymsha (changes requested):
I'll merge it as soon as it passes tests

rymsha (approved):

Issue #9682, opened 2022-09-07
TypeScript GetUserParams should be optional


ComLock:
![Screenshot 2022-09-07 at 10 22 41](https://user-images.githubusercontent.com/118847/188828527-850b9984-fcb5-4cc3-9631-6ea5f02f1d65.png)

PR #9683, opened 2022-09-07
Fix for TypeScript GetUserParams should be optional #9682

ComLock:


modules/lib/lib-auth/src/main/resources/lib/xp/auth.ts
```diff
@@ -163,7 +163,7 @@ interface GetUserHandler {
  *
  * @returns {User} Information for logged-in user.
  */
-export function getUser(params: GetUserParams): User {
+export function getUser(params ?:GetUserParams): User {
```
alansemenov, line 166: Same problem with annotation as the one mentioned in the other PR. Please don't change it.

ComLock: That's up to ye

alansemenov: Yes, but it's you who are changing the annotation. So, change it back please and make the parameter optional.

rymsha (approved):

Issue #9706, opened 2022-09-19
Fix node query params, making all properties optional


edloidas:
```ts
export interface QueryNodeParams {
    start?: number;
    count?: number;
    query: QueryDsl | string;
    sort?: string | SortDsl | SortDsl[];
    filters?: Filter | Filter[];
    aggregations?: Record<string, Aggregation>;
    suggestions?: Record<string, Suggestion>;
    highlight: Highlight;
    explain: boolean;
}
```
All properties must be optional, because `explain` default value is `false`, `highlight` is constructed in Java, etc.

PR #9708, opened 2022-09-19
Fix node query params, making all properties optional #9706

anatol-sialitski:


rymsha (approved):

edloidas (changes requested):
What about `query`? In TS it is marked as required (and in old [docs](https://developer.enonic.com/docs/xp/stable/api/lib-node#query_2) too). But in the JSDoc it's optional.
```ts
* @param {string|object} [params.query] Query expression.
```
converted to:
```
query    string | object    <optional>    Query expression.
```
It must be fixed either in types or in docs.

edloidas (approved):

Issue #9151, opened 2021-10-14
Add logout button on the 403 page


gbbirkisson:
This applies when using for example the OIDC provider on the /admin endpoint. The first time people log in, their user gets created but it does not have any permissions. They get a 403 page and cannot easily log out. This causes confusion because even if they have been given permission, their old session (that they are stuck with) does not have those permissions.

We can solve this by adding a logout button on the 403 page, that allows people to relog in.

PR #9713, opened 2022-09-19
Add logout button on the 403 page #9151

anatol-sialitski:


rymsha (approved):

rymsha (changes requested):
It looks like any error page gets logout button, not just 403 one.

alansemenov (approved):

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/exception/ErrorPageSimpleBuilder.java
```diff
@@ -45,6 +57,12 @@ public String build()
         {
             htmlBuilder.open( "h4" ).escapedText( tip ).close();
         }
+
+        if ( this.statusCode == HttpStatus.FORBIDDEN.value() )
```
rymsha, line 61: It should check existence of logout url, not the status

rymsha (changes requested):

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/exception/ExceptionInfo.java
```diff
@@ -126,6 +145,22 @@ public PortalResponse toHtmlResponse()
...
+            idProviderFunction( "logout" );
+
+        final AuthenticationInfo authInfo = ContextAccessor.current().getAuthInfo();
+
+        if ( authInfo.isAuthenticated() )
+        {
+            params.idProviderKey( authInfo.getUser().getKey().getIdProviderKey() );
+        }
+
+        return this.portalUrlService.identityUrl( params );
```
rymsha, line 161: There should be no logout url if not authenticated

rymsha (changes requested):

rymsha (approved):

vbradnitski (approved):

Issue #9623, opened 2022-08-10
Specify resource resolver via context 


vbradnitski:
In order to be able to fetch resource (for instance, descriptor) from either node or bundle we should allow to specify resource resolver in `Context`

PR #9715, opened 2022-09-19
Specify resource resolver via context #9623

vbradnitski:


modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationFactory.java
```diff
@@ -78,6 +79,21 @@ ApplicationUrlResolver createUrlResolver( final Bundle bundle )
         }
     }
 
+    ApplicationUrlResolver createUrlResolver( final Bundle bundle, final String name )
+    {
+        switch ( name )
```
rymsha, line 84: we are missing one which is `new MultiApplicationUrlResolver( classLoaderUrlResolver, bundleUrlResolver )`. Let's call it "real" for now

rymsha (changes requested):

rymsha (approved):

anatol-sialitski (approved):

rymsha (approved):

PR #9717, opened 2022-09-20
Add type parameters to Content

tajakobsen:
This commit adds two type parameters to the `Content` interface to make it possible to configure the shape of `Content.data` and `Content.type`.

By having the `Content.data` field specified by the `<Data>` type parameter a JavaScript-/TypeScript-developer can now easily specify the correct shape of `Content` (without using hacks like `as`). The `Content.type` field can be narrowed by setting the `<Type>` type parameter to a string literal (e.g. `"com.mycompany:article"`). This makes a union of `Content` into a discriminated union, where we can use normal `if`-statements to split the union.

The fallback value for `Content.data` is `Record<string, unknown>` and `Content.type` is `string` so the default shape of `Content` is by default the exact same as it was before this commit.

The `x` field has been set to a new global interface `XpXData`. This interface has the same shape as the old `XData` interface, but it allows developers extend the shape, so that it can be configured once for every content type. This will allow e.g. *lib-menu* to have code that automatically adds the correct shape to `XpXData`.

The shape of `Site` was wrong, and has been corrected. And `getSite()` and `getSiteConfig()` can now take the shape of `config` as a type parameter `<Config>`. The default value of `<Config>` is `Record<string, unknown>`. So it will by default have the exact same shape as before this commit.

*portal-lib* has been updated with the same changes to `Content` as *content-lib*.

I moved the `data` field before the `type` field in `Content`. The reason is that type inference in TypeScript might be better if the type parameter order is the same as the order they are found in the code.

### DIFF
--- a/modules/lib/lib-content/src/main/resources/lib/xp/content.ts
+++ b/modules/lib/lib-content/src/main/resources/lib/xp/content.ts
@@ -12,7 +12,6 @@ declare global {
         '/lib/xp/content': typeof import('./content');
     }
 
-    // eslint-disable-next-line @typescript-eslint/no-empty-interface
     interface XpXData {
         [key: string]: Record<string, Record<string, unknown>>
     }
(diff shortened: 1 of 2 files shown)


edloidas (approved):

edloidas:
I approve these changes. 
@rymsha @alansemenov Please, review and merge when ready.

Issue #9722, opened 2022-09-21
Fix global lib README links and wrapping


edloidas:
* The link to the global `README` is broken, since `lib/typescript` folder was removed.
* The lines in docs have unnecessary wrapping sometimes.

PR #9723, opened 2022-09-21
Fix global lib README links and wrapping #9722

edloidas:
* Fixed the link to the global `README`.
* Removed unnecessary wrapping in some libs README.


modules/lib/lib-admin/README.md
```diff
@@ -26,8 +26,7 @@ Add the corresponding types to your `tsconfig.json` file that is used for applic
 
 ### Require and custom imports
 
-To make `require` work out of box, you must install and add the `@enonic-types/global` types. Aside from providing definitions for XP global
-objects, e.g. `log`, `app`, `__`, etc, requiring library by the default path will return typed object.
+To make `require` work out of box, you must install and add the `@enonic-types/global` types. Aside from providing definitions for XP global objects, e.g. `log`, `app`, `__`, etc, requiring library by the default path will return typed object.
```
alansemenov, line 29:
```suggestion
To make `require` work out of the box, you must install and add the `@enonic-types/global` types. Aside from providing definitions for XP global objects, e.g. `log`, `app`, `__`, etc, requiring a library by the default path will return typed object.
```

alansemenov: Should be copied over to the other README-s as well.

alansemenov (changes requested):

modules/lib/lib-admin/README.md
```diff
@@ -48,13 +47,11 @@ objects, e.g. `log`, `app`, `__`, etc, requiring library by the default path wil
 const {getBaseUri, getAssetsUri, getHomeToolUrl, HomeToolUrlType} = require('/lib/xp/admin');
 ```
 
-More detailed explanation on how it works and how to type custom import function can be
-found [here](https://github.com/enonic/xp/tree/master/modules/lib/typescript/README.md).
+More detailed explanation on how it works and how to type custom import function can be found [here](https://github.com/enonic/xp/tree/master/modules/lib/README.md).
```
rymsha, line 50: I guess, it is better to refer to our official docs instead of github

edloidas: Replaced with a link to API. Note, that [TS chapter](https://github.com/enonic/doc-xp/pull/431) is not yet merged and deployed.

alansemenov (approved):

rymsha (approved):

Issue #8698, opened 2021-03-08
Branches with underscore must not be allowed


ComLock:
Try to make a branch called: '_'
Error invalid branch name.
I like it, good behaviour

Try to make a branch called: '1_'
No problemo, branch created

Try to push to it:
2021-03-08 12:46:53,817 INFO  com.enonic.app.explorer - (/webapp/webapp.js) pushParams:{
    "keys": [
        "/"
    ],
    "resolve": false,
    "target": "1_"
}
2021-03-08 12:46:53,819 INFO  com.enonic.app.explorer - (/webapp/webapp.js) pushRes:{
    "success": [],
    "failed": [],
    "deleted": []
}
Nothing happened? No error???

Try to make a node:
2021-03-08 12:46:53,824 ERROR com.enonic.app.explorer - {
    "catchedErrorClassName": "com.enonic.xp.node.NodeNotFoundException"
}
com.enonic.xp.node.NodeNotFoundException: Parent node to node with name '56b4cc8d-e721-4873-af03-2b62f3e54280' with parent path '/' not found

So it's sorta made a node, since it's made a node id, but then again no there is no node!

Try to make a branch called: '1_1'
No problemo, branch created

Try to push to it:
2021-03-08 12:49:53,520 ERROR c.e.x.p.i.e.ExceptionRendererImpl - Node with path [/] found in path-cache but not in storage
com.enonic.xp.web.WebException: Node with path [/] found in path-cache but not in storage

rymsha:
It appears that `_` in branches is a slightly different thing. They cannot be supported due to how branch index ids are constructed.

PR #9724, opened 2022-09-21
Branches with underscore must not be allowed #8698

rymsha:


modules/core/core-api/src/main/java/com/enonic/xp/branch/Branch.java
```diff
@@ -14,31 +15,41 @@
...
+        switch ( name )
+        {
+            case "master":
+                return MASTER;
+            case "draft":
+                return DRAFT;
+            default:
+            {
+                Preconditions.checkArgument( !isNullOrEmpty( name ), "Branch name cannot be null or empty" );
+                Preconditions.checkArgument( name.matches( "^" + VALID_BRANCH_ID_REGEX + "$" ), "Branch name format incorrect: " + name );
```
anatol-sialitski, line 42: What about add ^ and $ to VALID_BRANCH_ID_REGEX?

vbradnitski (approved):

Issue #9732, opened 2022-09-21
Task Library types improvements


edloidas:
_Comments from @tajakobsen:_

* `config` should be configured with a type parameter `<Config>`. [Reference](https://github.com/ItemConsulting/enonic-types/blob/modularize/packages/xp/task/index.d.ts)

PR #9733, opened 2022-09-22
Task Library types improvements #9732

edloidas:


edloidas:
Needs @tajakobsen approve.

modules/lib/lib-task/src/main/resources/lib/xp/task.ts
```diff
@@ -107,9 +107,9 @@ export function executeFunction(params: ExecuteFunctionParams): string {
     return bean.executeFunction();
 }
 
-export interface SubmitNamedTaskParams {
+export interface SubmitNamedTaskParams<Config extends object = never> {
```
tajakobsen, line 110: I think `never` as a default value is very good for TypeScript-developers, but it might trigger an error for JavaScript-developers using XP, since they may not be likely to set the type parameter.

So I suggest a more "flexible" default value:
```suggestion
export interface SubmitNamedTaskParams<Config extends object = Record<string, unknown>> {
```

tajakobsen (changes requested):
I think we need to replace `never` as the default value, with something more flexible (like `Record<string, unknown>`), so that JavaScript-developers – who are less likely to set type parameters – will not get errors flagged in their IDEs.

modules/lib/lib-task/src/main/resources/lib/xp/task.ts
edloidas: Updated

tajakobsen (approved):
Looks good

alansemenov (approved):

Issue #9731, opened 2022-09-21
Scheduler Library types improvements


edloidas:
_Comments from @tajakobsen:_

* `config` should always be a type parameter `<Config>`, so that a developer can lock down the shape. Both on schedule time, but also in the `run` function when the task is being run.
* Also see how `ScheduleTypeCron` and `ScheduleTypeOneTime` is a discrete union because of `type` field. [Reference](https://github.com/ItemConsulting/enonic-types/blob/modularize/packages/xp/scheduler/index.d.ts#L13)

PR #9734, opened 2022-09-22
Scheduler Library types improvements #9731

edloidas:
Made config generic.
Renamed `ScheduleJob` type to `ScheduledJob`.
Marked some properties of `ScheduledJob` non-null. Marked some properties in `create` function params optional. Fixed docs typos.

### DIFF
--- a/modules/lib/lib-scheduler/src/main/resources/lib/xp/scheduler.ts
+++ b/modules/lib/lib-scheduler/src/main/resources/lib/xp/scheduler.ts
@@ -21,54 +21,47 @@ function checkRequired<T extends object>(obj: T, name: keyof T): void {
 
 export type EditorFn<T> = (value: T) => T;
 
-export type ScheduleType = 'CRON' | 'ONE_TIME';
-
-export interface Schedule {
-    value: string;
-    type: ScheduleType;
-    timezone: string;
-}
-
-export interface CreateScheduledJobParams {
+export interface CreateScheduledJobParams<Config extends object = Record<string, unknown>> {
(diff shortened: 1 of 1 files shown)


edloidas:
Needs @tajakobsen approve.

alansemenov (approved):

modules/lib/lib-scheduler/src/main/resources/lib/xp/scheduler.ts
```diff
@@ -21,54 +21,47 @@ function checkRequired<T extends object>(obj: T, name: keyof T): void {
...
+    createdTime: string;
+    modifiedTime: string;
     lastRun?: string | null;
     lastTaskId?: string | null;
-    schedule: OnTimeSchedule | CronSchedule;
+    schedule: OneTimeSchedule | CronSchedule;
 }
 
-interface CreateScheduledJobHandler {
+interface CreateScheduledJobHandler<Config extends object = Record<string, unknown>> {
```
tajakobsen, line 61: I would probably not set av default type parameter for Handlers (like `CreateScheduledJobHandler` or `ModifyScheduledJobParams` or `GetScheduledJobHandler` etc,).

These interface are not exposed externally, and it's better if we just always pass on the `<Config>` shape from the calling function.

```suggestion
interface CreateScheduledJobHandler<Config extends object> {
```

tajakobsen (changes requested):
Just one little suggestion here, and that is to not use a default type parameter for "handlers".

Issue #9730, opened 2022-09-21
Grid Library types improvements


edloidas:
_Comments from @tajakobsen:_

See my code for good types that optionally lets a developer limit the shape of the shared object (across all of their code).
Just copy every type in this [file](https://github.com/ItemConsulting/enonic-types/blob/modularize/packages/xp/grid/index.d.ts#L5).

PR #9736, opened 2022-09-22
Grid Library types improvements #9730

edloidas:
Added generics for better and typesafe mappings.
Renamed `ISharedMap` to `JavaSharedMap`.
Renamed `SharedMap` to `SharedMapImpl`.
Added `SharedMap` interface and made it exportable instead of `SharedMapImpl` class. Removed default value in internal generics.

### DIFF
--- a/modules/lib/lib-grid/src/main/resources/lib/xp/grid.ts
+++ b/modules/lib/lib-grid/src/main/resources/lib/xp/grid.ts
@@ -15,47 +15,75 @@ declare global {
 
 export type SharedMapValueType = string | number | boolean | object | null;
 
-export type SharedMapModifierFn = (oldValue: SharedMapValueType) => SharedMapValueType;
+export type GridMap = Record<string, SharedMapValueType>;
 
-interface ISharedMap {
-    get(key: string): SharedMapValueType;
+export type SharedMapModifierFn<Map extends GridMap, Key extends keyof Map> = (value: Map[Key] | null) => Map[Key] | null;
 
-    delete(key: string): SharedMapValueType;
+interface JavaSharedMap<Map extends GridMap> {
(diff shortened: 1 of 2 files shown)


modules/lib/lib-grid/src/main/resources/lib/xp/grid.ts
```diff
@@ -125,29 +153,25 @@ export class SharedMap {
...
-function requireNotNull<T extends string | SharedMapModifierFn>(value: T, parameterName: string): T {
+function requireNotNull<T>(value: T, parameterName: string): T {
     if (value == null) {
         throw `Parameter "${parameterName}" is required`;
     }
     return value;
 }
 
-function convertValue(value: SharedMapValueType): SharedMapValueType {
+function convertValue<Map extends GridMap, Key extends keyof Map>(value: Map[Key] | null): Map[Key] | null {
```
tajakobsen, line 168: When I see the `as` keyword, I always try to figure out a way to solve a problem with out it.

In this case I think we might be fooling ourselves a little bit with the return type of `convertValue()`. In reality it can return `Record<string, unknown>` from `getMap()` or `object[]` from `getList()` too.

The job of this function is to create values that can be consumed by `JavaSharedMap`. So I think we should update the return type here, and change `JavaSharedMap` to accept the actual types returned.

```suggestion
function convertValue(value: SharedMapValueType | null): SharedMapValueType | Record<string, unknown> | object[] | null {
```

---

Not now, but in the long run having "branded types" to represent the Java-version of values might help prevent bugs. Because developers will not be able to use JavaScript-values, when Java-values are needed.

tajakobsen (changes requested):
I think you have done a really good piece of work here!

I have one comment that relates to the types from `convertValue()`.

tajakobsen (approved):
Good job

rymsha (approved):

Issue #9726, opened 2022-09-21
Libraries shared types


edloidas:
Some of the code (like types and functions) between the libraries is duplicated. Look at how it can be shared across libraries.

---
_Comments from @tajakobsen:_

You should consider creating a shared types library (@enonic-types/shared-types), so that you can maintain some of the types there.
Some duplicated types that can give you problems later are
 - `Content` (from Content Lib)
 - `PrincipalKey`, `PrincipalKeyUser`, `PrincipalKeyGroup`, `PrincipalKeyRole` (from Auth lib)
 - `Component` (from Portal Lib)
There are also some very useful types in my global library that is needed in every project, and that I don't think you have included yet.
These center around `XP.Request` and `XP.Response`. Note that I have namespaced them, since there already exists a global Request and Response (from the web platform I think).

Issue #9729, opened 2022-09-21
Core Libraries shared types imports


edloidas:
Use shared types for:

* [x] `RunContext.principals` can be of type `PrincipalKey[]`. [Reference](https://github.com/ItemConsulting/enonic-types/blob/modularize/packages/xp/context/index.d.ts#L5)
* [x] `PrincipalKeys` in Scheduler lib.  [Reference](https://github.com/ItemConsulting/enonic-types/blob/modularize/packages/xp/scheduler/index.d.ts#L29-L31)
* [x] Move `PrincipalKeys` from Auth lib.
* [x] Content lib
* [x] Node lib
* [x] IO (`Resource`)
* [x] Portal lib: 
  * [x] `Site` is a `Content`
  * [x] `getContent()` should return `Content.page` with type parameters for `<Config>` and `<Regions>`. [Reference](https://github.com/ItemConsulting/enonic-types/blob/modularize/packages/xp/portal/index.d.ts#L70)

PR #9739, opened 2022-09-26
Libraries shared types #9726

edloidas:
Added shared type, that will be later imported in some core libs to prevent duplicated code. Removed redundant code in event lib.


modules/lib/package.template.json
```diff
@@ -26,6 +26,7 @@
         "url": "https://github.com/enonic/xp/issues"
     },
     "homepage": "https://github.com/enonic/xp/tree/master#readme",
+    "dependencies": {%DEPENDECNIES%},
```
alansemenov, line 29: Typo

modules/lib/shared/README.md
```diff
@@ -0,0 +1,58 @@
...
+## Modules
+
+* `auth` — principal-related types.
+* `content` — `Content`, `Workflow`, and different states.
+* `portal` — `Component` and `Region` types.
+* `utils` — contains a bunch of utility types that simplify custom types creation.
+
+### Content
+
+`Content` contains the `x` property, that have a very special `XpXData` type.
```
alansemenov, line 48: "have" -> "has"

alansemenov (changes requested):

rymsha (changes requested):
It looks like overengineering to me.
Lets have a meeting.

modules/lib/shared/utils.d.ts
```diff
@@ -0,0 +1,19 @@
+export type EmptyObject = Record<string, never>;
+
+export type AnyObject = Record<string | number | symbol, unknown>;
+
+export type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
+
+export type XOR<T, U> = T | U extends object ? (Without<T, U> & U) | (Without<U, T> & T) : T | U;
```
tajakobsen, line 7: I don't know how I feel about using `XOR` anymore. It makes the code a little bit difficult to read. I think I would prefer code more like this style:

```TypeScript
export interface ByPath {
  id: never;
  path: string;
}

export interface ById {
  id: string;
  path: never;
}

export type ImageUrlParams = (ById | ByPath) & {
  scale: ImageScale;
  quality?: number;
  background?: string;
  format?: string;
  filter?: string;
  type?: "server" | "absolute";
  params?: Params;
};

```

modules/lib/shared/content.d.ts
```diff
@@ -0,0 +1,63 @@
...
+    from?: string;
+    to?: string;
+    first?: string;
+}
+
+export interface Content<
+    Data = Record<string, unknown>,
+    Type extends string = string,
+    Config extends object = object,
+    Regions extends Record<string, Region> = Record<string, Region>,
```
tajakobsen, line 37: I am a little bit unsure if `Config` and `Regions` should always be type parameters but lets talk about that.

tajakobsen (changes requested):
Did you consider letting each library have it's own *package.json*, and using "npm workspaces" to operate on all of them together?

edloidas:
* Move `utils` into a separate repo
* Rename `shared` -> `core`
* Move all types into `index.d.ts`
* Add `Resource` type from IO library

modules/lib/core/index.d.ts
```diff
@@ -0,0 +1,79 @@
...
+    from?: string;
+    to?: string;
+    first?: string;
+}
+
+export interface Component<Config extends object = object, Regions extends Record<string, Region> = Record<string, Region>> {
+    config: Config;
+    descriptor: string;
+    path: string;
+    type: 'page' | 'layout' | 'part';
```
alansemenov, line 40: Shouldn't this be a union type as well? Similar to `WorkflowCheckState` etc.

edloidas: Not necessary. I would rather think of moving some of other union types from separate declaration to property type.
Component's type is not something that will be frequently imported or used apart from `Component` itself. It can also be retrieved via `type ComponentType = Component['type']` anytime.

modules/lib/build.gradle
```diff
@@ -8,11 +8,18 @@ node {
     version = '16.15.0'
 }
 
+def coreDir = "$projectDir/core/"
```
rymsha, line 11: It would be nice to NOT introduce an extra directory.
Can we have a core.ts file instead and not interfare with the build.
Or make core a gradle module

modules/lib/build.gradle
```diff
@@ -139,6 +175,8 @@ subprojects {
     task prepareToPublish( type: Copy, dependsOn: processResources ) {
         def fullName = project.name
         def shortName = project.name.substring( 4 )
+        // TODO: Replace with the actual shared dependencies in the future, like ['@enonic-types/core']
```
rymsha, line 178: there should be any todo left

modules/lib/core/index.d.ts
```diff
@@ -0,0 +1,79 @@
...
+export type GroupKey = `group:${string}:${string}`;
+export type RoleKey = `role:${string}`;
+
+export type PrincipalKey = UserKey | GroupKey | RoleKey;
+
+export type WorkflowState = 'IN_PROGRESS' | 'PENDING_APPROVAL' | 'REJECTED' | 'READY';
+
+export type WorkflowCheckState = 'PENDING' | 'REJECTED' | 'APPROVED';
+
+export type ContentInheritType = 'CONTENT' | 'PARENT' | 'NAME' | 'SORT';
```
tajakobsen, line 16: I would consider inlining these straight in the interfaces that uses them because: 
 1. I don't think anybody will use these types in application code.
 2. They will be located where they are used without abstraction (you don't have to jump around in the code)
 3. They can always be derived in application if somebody absolutely need them:
     ```typescript
     type WorkflowState = Workflow['state'];
     type ContentInheritType = NonNullable<Content['inherit']>
     ```

modules/lib/core/index.d.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export interface Workflow {
```
tajakobsen, line 18: I would inline `Workflow` in `Content` too, for the same reasons above

tajakobsen (changes requested):
Good job! I mostly have minor comments. The one larger issue is for `Content`. I would consider joining the last two type parameters to a single one `<Page>`. Simplifying `Content` as much as possible will save headache in the future, believe me :)

modules/lib/build.gradle
edloidas: It's just a reminder for [this](https://github.com/enonic/xp/issues/9729) task, so there be no miss understanding, why the following array is actually empty. The actual list of the libs that will use shared types is not final yet.

modules/lib/build.gradle
edloidas: We can, but we have a unique README file for this module. Where should it be placed? Just pile all in the `/lib`?

modules/lib/core/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+```json
+{
+  "compilerOptions": {
+    "types": [
+      "@enonic-types/core"
+    ]
+  }
+}
+```
```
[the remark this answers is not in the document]
edloidas: My concern was that you must import that library to update/merge `XpXData`, unlike `XpLibraries` interface from the libs. But I guess, since it's empty and won't be used without Content, we can remove this.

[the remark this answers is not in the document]
tajakobsen: Yes, I think `XpXData` til be created globally as soon as *core* is imported (to be able to use `Content`)

modules/lib/build.gradle
edloidas: As discussed, will be updated later in different task.

tajakobsen (approved):
It's looking very good. :+1: I'm approving this now, but I have one comment about README.md that you should look at before merging.

tajakobsen (approved):
Looks good :+1:

rymsha (approved):

alansemenov (approved):

rymsha (approved):

Issue #9741, opened 2022-09-27
Refactor IO library Resource class


edloidas:
Lib must not export class. Export interface instead. Keep class for internal usage.

PR #9743, opened 2022-09-27
Refactor IO library Resource class #9741

edloidas:
Prevented class from being exported.
Updated docs.

### DIFF
--- a/modules/lib/lib-io/src/main/resources/lib/xp/io.ts
+++ b/modules/lib/lib-io/src/main/resources/lib/xp/io.ts
@@ -13,9 +13,7 @@ declare global {
     }
 }
 
-const bean = __.newBean<IOHandlerBean>('com.enonic.xp.lib.io.IOHandlerBean');
-
-interface IResource {
+interface JavaResource {
     getSize(): number;
 
     getBytes(): object;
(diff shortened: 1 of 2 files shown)


alansemenov (approved):

modules/lib/lib-node/src/main/resources/lib/xp/node.ts
```diff
@@ -820,10 +820,6 @@ export interface Node {
     _permissions?: AccessControlEntry[];
 }
 
-/**
```
rymsha, line 823: why is it removed?

edloidas: The interface do not need to be documented, classes do, because we fetch the docs from `js` files where all types will be removed.

rymsha (changes requested):

modules/lib/lib-io/src/main/resources/lib/xp/io.ts
```diff
@@ -13,24 +13,50 @@ declare global {
     }
 }
 
-const bean = __.newBean<IOHandlerBean>('com.enonic.xp.lib.io.IOHandlerBean');
-
-interface IResource {
+interface JavaResource {
```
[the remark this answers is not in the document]
rymsha: I want JS Resource  interface to be in core.

rymsha (approved):

rymsha (approved):

Issue #9668, opened 2022-08-31
NodeMultiRepoQuery does not return suggestions in result


anatol-sialitski:
NodeMultiRepoQuery does not return suggestions in result

PR #9744, opened 2022-09-27
NodeMultiRepoQuery does not return suggestions in result #9668

anatol-sialitski:


modules/lib/lib-node/src/main/resources/lib/xp/node.ts
```diff
@@ -416,15 +416,9 @@ export interface Explanation {
     details: Explanation[];
 }
 
-export interface NodeQueryResult {
+export interface NodeQueryResultBase {
```
edloidas, line 419: Since `NodeQueryResult` and  `NodeMultiRepoQueryResult` are almost the same, I would suggest to use the following:
```ts
interface NodeQueryResult {
    total: number;
    count: number;
    aggregations?: AggregationsResult;
    suggestions?: {
        [suggestionName: string]: {
            text: string;
            length: number;
            offset: number;
            options: {
                text: string;
                score: number;
                freq?: number; // only for term
            }[];
        }[];
    }[];
    hits: {
        id: string;
        score: number;
        explanation?: Explanation;
        highlight?: HighlightResult;
    }[];
}

export type NodeMultiRepoQueryResult = NodeQueryResult & {
    hits: {
        repoId: string;
        branch: string;
    }[];
};
```

edloidas (changes requested):

rymsha (approved):

vbradnitski (approved):

Issue #9742, opened 2022-09-27
Add type definitions for Request and Response


edloidas:
Add `Request` and `Response` types under the `lib-portal` types.

Page requests doesn't have contextPath

ComLock:
It would be nice if Request headers keys are stricter than string so code will be future proof for HTTP/2.

[8.1.2.6](https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2.6).  Malformed Requests and Responses

   A malformed request or response is one that is an otherwise valid
   sequence of HTTP/2 frames but is invalid due to the presence of
   extraneous frames, prohibited header fields, the absence of mandatory
   header fields, or the inclusion of uppercase header field names.

ComLock:
https://github.com/enonic/lib-react4xp/blob/master/src/main/resources/types/Request.d.ts
https://github.com/enonic/lib-react4xp/blob/master/src/main/resources/types/Response.d.ts

ComLock:
I wonder how many times I've done this :)
https://github.com/enonic/app-explorer/blob/master/src/main/resources/types/Request.d.ts

ComLock:
Many :)
https://github.com/enonic/lib-explorer/blob/master/src/main/resources/lib/explorer/types/Request.d.ts
https://github.com/enonic/lib-explorer/blob/master/src/main/resources/lib/explorer/types/Response.d.ts

ComLock:
Nope
<img width="423" alt="Screenshot 2023-04-14 at 14 42 25" src="https://user-images.githubusercontent.com/118847/232046521-bf39a5a7-ebac-4a9d-b6ff-795b2770d995.png">

ComLock:
I keep forgetting there is no such thing as lowercasestring in typings yet.

ComLock:
`Lowercase<StringType>`
https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html#lowercasestringtype

ComLock:
Here we go again :)
https://github.com/enonic/lib-static/blob/master/src/main/resources/lib/enonic/static/types/Request.d.ts
https://github.com/enonic/lib-static/blob/master/src/main/resources/lib/enonic/static/types/Response.d.ts

ComLock:
I think what needs to be done here is find some real life examples to test the Request and Response types against.

PR #9745, opened 2022-09-27
Add Request and Response to lib-portal #9742

edloidas:
After discussion with @rymsha, it was decided, that `Request` and `Response` types will go to `lib-portal`, because it's where they actually belong in Java as part of the [`portal-api`](https://developer.enonic.com/docs/xp/stable/framework/http).

### DIFF
--- a/modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
+++ b/modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
@@ -83,6 +83,86 @@ export interface SiteConfig<Config> {
     config: Config;
 }
 
+export type DefaultHeaders = {
+    Accept: string;
+    'Accept-Charset': string;
+    'Accept-Encoding': string;
+    Authorization: string;
+    'Content-Length': string,
+    'Content-Type': string,
+    Cookies: string;
+    Expect: string;
+    Host: string;
+    'If-None-Match': string;
+    Language: string;
+    'User-Agent': string;
+};
+
+export interface Cookie {
+    value: string;
+    path?: string;
+    domain?: string;
+    comment?: string;
+    maxAge?: number; // -1
+    secure?: boolean; // false
+    httpOnly?: boolean; // false
+    sameSite?: 'lax' | 'strict' | 'none' | ''; // ''
+}
+
+interface Request<Params extends Record<string, string | undefined> = Record<string, string | undefined>,
+    Headers extends Record<string, string | undefined> = DefaultHeaders,
+    Cookies extends Record<string, string | Cookie | undefined> = Record<string, string | Cookie | undefined>,
+    > {
+    method: 'GET' | 'PUT' | 'POST' | 'DELETE' | 'HEAD' | 'PATCH' | 'OPTIONS' | 'TRACE' | 'CONNECT';
+    scheme: 'http' | 'https';
+    host: string;
+    port: number;
+    path: string;
+    rawPath: string;
+    url: string;
+    remoteAddress: string;
+    mode: 'inline' | 'edit' | 'preview' | 'live';
+    webSocket?: boolean;
+    repositoryId: string;
+    branch: 'draft' | 'master';
+    contextPath: string;
+    body: string;
+    params: Params;
+    headers: Headers;
+    cookies: Cookies;
+    contentType: string;
+}
+
+type ResponseType = string | object | unknown[] | ReadonlyArray<unknown> | null;
+
+type LiteralUnion<T extends U, U = string> = T | (U & Record<never, never>);
+
+export interface PageContributions {
+    headBegin?: string | string[];
+    headEnd?: string | string[];
+    bodyBegin?: string | string[];
+    bodyEnd?: string | string[];
+}
+
+export interface Response<ResponseBody extends ResponseType = ResponseType,
+    Headers extends Record<string, string | undefined> = DefaultHeaders,
+    Cookies extends Record<string, string | Cookie | undefined> = Record<string, string | Cookie | undefined>,
+    > {
+    status: number;
+    body: ResponseBody;
+    contentType: LiteralUnion<| 'text/html'
+        | 'application/json'
+        | 'application/problem+json'
+        | 'text/xml'
+        | 'application/xml'>;
+    headers: Headers;
+    cookies: Cookies;
+    redirect: string;
+    postProcess: boolean;
+    pageContributions: PageContributions;
+    applyFilters: boolean;
+}
+
 export interface AssetUrlParams {
     path: string;
     application?: string;
--- a/modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
+++ b/modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
@@ -83,37 +83,25 @@ export interface SiteConfig<Config> {
     config: Config;
 }
 
-export type DefaultHeaders = {
-    Accept: string;
-    'Accept-Charset': string;
-    'Accept-Encoding': string;
-    Authorization: string;
-    'Content-Length': string,
-    'Content-Type': string,
-    Cookies: string;
-    Expect: string;
-    Host: string;
-    'If-None-Match': string;
-    Language: string;
-    'User-Agent': string;
-};
-
 export interface Cookie {
     value: string;
     path?: string;
     domain?: string;
-    comment?: string;
-    maxAge?: number; // -1
-    secure?: boolean; // false
-    httpOnly?: boolean; // false
-    sameSite?: 'lax' | 'strict' | 'none' | ''; // ''
+    maxAge?: number;
+    secure?: boolean;
+    httpOnly?: boolean;
+    sameSite?: 'lax' | 'strict' | 'none' | '';
 }
 
-interface Request<Params extends Record<string, string | undefined> = Record<string, string | undefined>,
-    Headers extends Record<string, string | undefined> = DefaultHeaders,
+type LiteralUnion<T extends U, U = string> = T | (U & Record<never, never>);
+
+export type RequestParams = Record<string, string | string[] | undefined>;
+
+export interface Request<Params extends RequestParams = RequestParams,
+    Headers extends Record<string, string | undefined> = Record<string, string | undefined>,
     Cookies extends Record<string, string | Cookie | undefined> = Record<string, string | Cookie | undefined>,
     > {
-    method: 'GET' | 'PUT' | 'POST' | 'DELETE' | 'HEAD' | 'PATCH' | 'OPTIONS' | 'TRACE' | 'CONNECT';
+    method: LiteralUnion<'GET' | 'PUT' | 'POST' | 'DELETE' | 'HEAD' | 'PATCH' | 'OPTIONS' | 'TRACE' | 'CONNECT'>;
     scheme: 'http' | 'https';
     host: string;
     port: number;
@@ -121,57 +109,44 @@ interface Request<Params extends Record<string, string | undefined> = Record<str
     rawPath: string;
     url: string;
     remoteAddress: string;
-    mode: 'inline' | 'edit' | 'preview' | 'live';
-    webSocket?: boolean;
+    mode: 'inline' | 'edit' | 'preview' | 'live' | 'admin';
+    webSocket: boolean;
     repositoryId: string;
-    branch: 'draft' | 'master';
+    branch: LiteralUnion<'draft' | 'master'>;
     contextPath: string;
-    body: string;
+    body?: string;
     params: Params;
     headers: Headers;
     cookies: Cookies;
-    contentType: string;
+    contentType?: string;
 }
 
-type ResponseType = string | object | unknown[] | ReadonlyArray<unknown> | null;
-
-type LiteralUnion<T extends U, U = string> = T | (U & Record<never, never>);
-
-export interface PageContributions {
-    headBegin?: string | string[];
-    headEnd?: string | string[];
-    bodyBegin?: string | string[];
-    bodyEnd?: string | string[];
-}
-
-export interface Response<ResponseBody extends ResponseType = ResponseType,
-    Headers extends Record<string, string | undefined> = DefaultHeaders,
-    Cookies extends Record<string, string | Cookie | undefined> = Record<string, string | Cookie | undefined>,
-    > {
+export interface Response<ResponseBody = unknown> {
     status: number;
     body: ResponseBody;
-    contentType: LiteralUnion<| 'text/html'
-        | 'application/json'
-        | 'application/problem+json'
-        | 'text/xml'
-        | 'application/xml'>;
-    headers: Headers;
-    cookies: Cookies;
+    contentType: string;
+    headers: Record<string, string | undefined>;
+    cookies: Record<string, string | Cookie | undefined>;
     redirect: string;
     postProcess: boolean;
-    pageContributions: PageContributions;
+    pageContributions: {
+        headBegin?: string | string[];
+        headEnd?: string | string[];
+        bodyBegin?: string | string[];
+        bodyEnd?: string | string[];
+    };
     applyFilters: boolean;
 }
 
-export interface AssetUrlParams {
+export interface AssetUrlParams<Params extends RequestParams = RequestParams> {
     path: string;
     application?: string;
     type?: string;
-    params?: object;
+    params?: Params;
 }
 
 interface AssetUrlHandler {
-    createUrl(value: object): string;
+    createUrl(value: ScriptValue): string;
 }
 
 /**
@@ -187,12 +162,12 @@ interface AssetUrlHandler {
  *
  * @returns {string} The generated URL.
  */
-export function assetUrl(params: AssetUrlParams): string {
+export function assetUrl<Params extends RequestParams = RequestParams>(params: AssetUrlParams<Params>): string {
     const bean = __.newBean<AssetUrlHandler>('com.enonic.xp.lib.portal.url.AssetUrlHandler');
     return bean.createUrl(__.toScriptValue(params));
 }
 
-export interface ImageUrlParams {
+export interface ImageUrlParams<Params extends RequestParams = RequestParams> {
     id: string;
     path: string;
     scale: string;
@@ -201,11 +176,11 @@ export interface ImageUrlParams {
     format?: string;
     filter?: string;
     server?: string;
-    params?: object;
+    params?: Params;
 }
 
 interface ImageUrlHandler {
-    createUrl(value: object): string;
+    createUrl(value: ScriptValue): string;
 }
 
 /**
@@ -226,21 +201,21 @@ interface ImageUrlHandler {
  *
  * @returns {string} The generated URL.
  */
-export function imageUrl(params: ImageUrlParams): string {
+export function imageUrl<Params extends RequestParams = RequestParams>(params: ImageUrlParams<Params>): string {
     const bean = __.newBean<ImageUrlHandler>('com.enonic.xp.lib.portal.url.ImageUrlHandler');
     return bean.createUrl(__.toScriptValue(params));
 }
 
-export interface ComponentUrlParams {
+export interface ComponentUrlParams<Params extends RequestParams = RequestParams> {
     id?: string;
     path?: string;
     component?: string;
     type?: string;
-    params?: object;
+    params?: Params;
 }
 
 interface ComponentUrlHandler {
-    createUrl(value: object): string;
+    createUrl(value: ScriptValue): string;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,6 +83,86 @@ export interface SiteConfig<Config> {
     config: Config;
 }
 
+export type DefaultHeaders = {
```
rymsha, line 86: There are a few problems with these headers definitions 
- the list is not complete (and cannot be)
- it enforces headers characters case
- it is called default,  but it is not 

I suggest to remove it

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,6 +83,86 @@ export interface SiteConfig<Config> {
...
+    'If-None-Match': string;
+    Language: string;
+    'User-Agent': string;
+};
+
+export interface Cookie {
+    value: string;
+    path?: string;
+    domain?: string;
+    comment?: string;
```
rymsha, line 105: Remove comment. It must not be used externally.

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,6 +83,86 @@ export interface SiteConfig<Config> {
...
+    host: string;
+    port: number;
+    path: string;
+    rawPath: string;
+    url: string;
+    remoteAddress: string;
+    mode: 'inline' | 'edit' | 'preview' | 'live';
+    webSocket?: boolean;
+    repositoryId: string;
+    branch: 'draft' | 'master';
```
rymsha, line 127: We support more than that

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,6 +83,86 @@ export interface SiteConfig<Config> {
...
+    contextPath: string;
+    body: string;
```
rymsha, line 129: It can be undefined

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,6 +83,86 @@ export interface SiteConfig<Config> {
...
+    method: 'GET' | 'PUT' | 'POST' | 'DELETE' | 'HEAD' | 'PATCH' | 'OPTIONS' | 'TRACE' | 'CONNECT';
+    scheme: 'http' | 'https';
+    host: string;
+    port: number;
+    path: string;
+    rawPath: string;
+    url: string;
+    remoteAddress: string;
+    mode: 'inline' | 'edit' | 'preview' | 'live';
+    webSocket?: boolean;
```
rymsha, line 125: Is it correct? websocket is optional boolean?

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,6 +83,86 @@ export interface SiteConfig<Config> {
...
+    secure?: boolean; // false
+    httpOnly?: boolean; // false
+    sameSite?: 'lax' | 'strict' | 'none' | ''; // ''
+}
+
+interface Request<Params extends Record<string, string | undefined> = Record<string, string | undefined>,
+    Headers extends Record<string, string | undefined> = DefaultHeaders,
+    Cookies extends Record<string, string | Cookie | undefined> = Record<string, string | Cookie | undefined>,
+    > {
+    method: 'GET' | 'PUT' | 'POST' | 'DELETE' | 'HEAD' | 'PATCH' | 'OPTIONS' | 'TRACE' | 'CONNECT';
```
rymsha, line 116: This list seem too short

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,6 +83,86 @@ export interface SiteConfig<Config> {
...
+    rawPath: string;
+    url: string;
+    remoteAddress: string;
+    mode: 'inline' | 'edit' | 'preview' | 'live';
+    webSocket?: boolean;
+    repositoryId: string;
+    branch: 'draft' | 'master';
+    contextPath: string;
+    body: string;
+    params: Params;
```
rymsha, line 130: Params is way more complicated than Record

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,6 +83,86 @@ export interface SiteConfig<Config> {
...
+    headers: Headers;
+    cookies: Cookies;
+    contentType: string;
```
rymsha, line 133: I think it may be undefined in some cases

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,6 +83,86 @@ export interface SiteConfig<Config> {
...
+    bodyEnd?: string | string[];
+}
+
+export interface Response<ResponseBody extends ResponseType = ResponseType,
+    Headers extends Record<string, string | undefined> = DefaultHeaders,
+    Cookies extends Record<string, string | Cookie | undefined> = Record<string, string | Cookie | undefined>,
+    > {
+    status: number;
+    body: ResponseBody;
+    contentType: LiteralUnion<| 'text/html'
```
rymsha, line 153: I am surprised with a set of predefined ones. I suggest to leave just string

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,6 +83,86 @@ export interface SiteConfig<Config> {
...
+    branch: 'draft' | 'master';
+    contextPath: string;
+    body: string;
+    params: Params;
+    headers: Headers;
+    cookies: Cookies;
+    contentType: string;
+}
+
+type ResponseType = string | object | unknown[] | ReadonlyArray<unknown> | null;
```
rymsha, line 136: We basically support almost anything. Why having a type?

rymsha (changes requested):

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
tajakobsen: Maybe make it a `LiteralUnion` to let it accept `string` too. But give code completion on these most used verbs?

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,6 +83,86 @@ export interface SiteConfig<Config> {
...
+    path?: string;
+    domain?: string;
+    comment?: string;
+    maxAge?: number; // -1
+    secure?: boolean; // false
+    httpOnly?: boolean; // false
+    sameSite?: 'lax' | 'strict' | 'none' | ''; // ''
+}
+
+interface Request<Params extends Record<string, string | undefined> = Record<string, string | undefined>,
```
tajakobsen, line 112: If you have a url with the same query parameter listed more then once, the value of the param in XP is an `Array<string>`

```suggestion
interface Request<Params extends Record<string, string | Array<string> | undefined> = Record<string, string | Array<string> | undefined>,
```

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
tajakobsen: I am ambivalent about this one. :)

**Pro:** Since it's a `LiteralUnion` it can be any `string`. But I get that nice code completion for the most common cases, so that I don't have to google the syntax every time (and trust me, I will google it every time to be sure).

**Con:** It can be anything, so why limit it? And this doesn't really have anything to do with XP.

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
edloidas: We don't support numbers and boolean, I guess

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
edloidas: It's just a suggested values, `contentType` can be any string, but those one are the most used and therefore listed here.

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
edloidas: Is it? Do we support something non-standard?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
edloidas: I though it may be just any map/object. Do we also support arrays and primitive types?

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
tajakobsen: I think it's better to not pass in these type parameters, because your `get` or `post` function has to be able to handle any input, both missing and string and Arrays, so there is little value in locking in a shape.

The value of specifying these shapes in type parameters, would be if you could use the same shape "in both ends". That means that you specify an `interface` that can be used both by `Request` and `serviceUrl()` or `pageUrl()`.

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
edloidas: Okay

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
tajakobsen: I think @rymsha is correct. This type just kind of grew over time to encompass everything.

@edloidas Maybe just use `unknown` as the default type for `<ResponseBody>`?

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
edloidas: Do we? I though it's the only possible values, as described in [docs](https://developer.enonic.com/docs/xp/stable/framework/http#http-request). Okay, I can add those values as suggested and make property accept string.

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,6 +83,86 @@ export interface SiteConfig<Config> {
...
+export interface PageContributions {
+    headBegin?: string | string[];
+    headEnd?: string | string[];
+    bodyBegin?: string | string[];
+    bodyEnd?: string | string[];
+}
+
+export interface Response<ResponseBody extends ResponseType = ResponseType,
+    Headers extends Record<string, string | undefined> = DefaultHeaders,
+    Cookies extends Record<string, string | Cookie | undefined> = Record<string, string | Cookie | undefined>,
```
tajakobsen, line 149: As with `Request` I'm a little bit unsure if having type parameters for `Headers` and `Cookies` brings any value.

The value would be if that same interface could be used in both ends of the communication. But the other end here is the browser, or another application.

I think it's not worth having, because of the rise in complexity of the type for the application developers.

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,6 +83,86 @@ export interface SiteConfig<Config> {
...
+    headers: Headers;
+    cookies: Cookies;
+    contentType: string;
+}
+
+type ResponseType = string | object | unknown[] | ReadonlyArray<unknown> | null;
+
+type LiteralUnion<T extends U, U = string> = T | (U & Record<never, never>);
+
+export interface PageContributions {
```
tajakobsen, line 140: I would just inline this type, since it's only used once. It also cleans up the exports, and it is co-located with where it's used.

If I need to use it in my application I would do:

```typescript
type PageContributions = Response['pageContributions'];
```

tajakobsen (changes requested):
I think it might not be a good idea to add type parameters for `Params`, `Cookies` and `Headers`. The reason is that we usually can't use that same interface in both ends of the "communication", and we don't really know what shape will come in on these parameters, because the initiator of the call can really pass in anything.

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
rymsha: it is better explained https://github.com/enonic/xp/pull/9745/files#r982034003

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
rymsha: Content-type is a tricky one and error prone.
The suggested "most used" are not without problems either.
For instance : `text/html` without charset may end up not decoding correctly in browsers.

edloidas: I see. Then let us use the `string` here

rymsha (changes requested):

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,15 +83,70 @@ export interface SiteConfig<Config> {
...
+    contentType?: string;
+}
+
+export interface Response<ResponseBody = unknown> {
+    status: number;
+    body: ResponseBody;
+    contentType: string;
+    headers: Record<string, string | undefined>;
+    cookies: Record<string, string | Cookie | undefined>;
+    redirect: string;
```
rymsha, line 130:
```suggestion
    redirect?: string;
```

redirect is a "write only" property, initially it is undefined

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,15 +83,70 @@ export interface SiteConfig<Config> {
...
+    postProcess: boolean;
+    pageContributions: {
+        headBegin?: string | string[];
+        headEnd?: string | string[];
+        bodyBegin?: string | string[];
+        bodyEnd?: string | string[];
+    };
+    applyFilters: boolean;
```
rymsha, line 138: optional `webSocket` field is missing
https://developer.enonic.com/docs/xp/stable/framework/websocket#http-websockets

rymsha (changes requested):

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,15 +83,70 @@ export interface SiteConfig<Config> {
...
+    sameSite?: 'lax' | 'strict' | 'none' | '';
+}
+
+type LiteralUnion<T extends U, U = string> = T | (U & Record<never, never>);
+
+export type RequestParams = Record<string, string | string[] | undefined>;
+
+export interface Request<Params extends RequestParams = RequestParams,
+    Headers extends Record<string, string | undefined> = Record<string, string | undefined>,
+    Cookies extends Record<string, string | Cookie | undefined> = Record<string, string | Cookie | undefined>,
```
tajakobsen, line 102: Is it correct that a cookie can be `string | Cookie | undefined`? Since this comes from XP, it seems like a risky thing to get a cookie be both `string` and `Cookie`?

Could it be that cookies in `Response` has this shape, but not in `Request`?

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,15 +83,70 @@ export interface SiteConfig<Config> {
...
+    branch: LiteralUnion<'draft' | 'master'>;
+    contextPath: string;
+    body?: string;
+    params: Params;
+    headers: Headers;
+    cookies: Cookies;
+    contentType?: string;
+}
+
+export interface Response<ResponseBody = unknown> {
```
tajakobsen, line 124: Every field in `Response` should be optional, right?

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -241,20 +296,20 @@ interface PageUrlHandler {
  *
  * @returns {string} The generated URL.
  */
-export function pageUrl(params: PageUrlParams): string {
+export function pageUrl<Params extends RequestParams = RequestParams>(params: PageUrlParams<Params>): string {
```
tajakobsen, line 299: For all of these functions (`pageUrl`, `serviceUrl`) both `number` and `boolean` is valid types of the value (and `object` are just url-encoded, but that might be more coincidental).

Should maybe `Params` not extend `RequestParams` but another shape instead that allows for other types of the values.

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,15 +83,70 @@ export interface SiteConfig<Config> {
...
+    secure?: boolean;
+    httpOnly?: boolean;
+    sameSite?: 'lax' | 'strict' | 'none' | '';
+}
+
+type LiteralUnion<T extends U, U = string> = T | (U & Record<never, never>);
+
+export type RequestParams = Record<string, string | string[] | undefined>;
+
+export interface Request<Params extends RequestParams = RequestParams,
```
tajakobsen, line 100: Did you consider removing (all or some of) the type parameters like I suggested earlier?

tajakobsen (changes requested):
I have a few comments here. :) Don't know if I'm coming with this feedback prematurely.

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,15 +83,70 @@ export interface SiteConfig<Config> {
...
     path: string;
     application?: string;
     type?: string;
-    params?: object;
+    params?: Params;
 }
 
 interface AssetUrlHandler {
-    createUrl(value: object): string;
+    createUrl(value: ScriptValue): string;
```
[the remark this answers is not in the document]
rymsha: Aren't global always exist? 
`__` is always there, so does `__.toScriptValue`. same as `require`

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
rymsha: Request cookies are always `string`. (except for the cases when there is no such cookie)

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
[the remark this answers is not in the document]
edloidas: No, not during runtime, unless global is included as dependency.
Right, I'd better revert it back to object. Totally forgot about this problem.

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
rymsha: We have a dilemma here. Response that controllers receive has many non-optional fields. But the one that controllers could return may have all fields optional (because it is sort-of patch to existing one)

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
edloidas: I see, let's make it string/undefined then. But the Response cookie can still be a string/Cookie/undefined, right?

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
tajakobsen: It's probably necessary to have two `Response` types. Maybe something like this:

```typescript
export interface InputResponse { // sorry about the bad naming here
  status: number;
  ...
}

export type Response = Partial<InputResponse>;
```

Be careful about naming it `Response` or `Request`, I think there will normally be a naming collision against the global fields in the web platform. (That was why I namespaced it in *enonic-types*)

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -83,15 +83,70 @@ export interface SiteConfig<Config> {
...
+
+type LiteralUnion<T extends U, U = string> = T | (U & Record<never, never>);
+
+export type RequestParams = Record<string, string | string[] | undefined>;
+
+export interface Request<Params extends RequestParams = RequestParams,
+    Headers extends Record<string, string | undefined> = Record<string, string | undefined>,
+    Cookies extends Record<string, string | Cookie | undefined> = Record<string, string | Cookie | undefined>,
+    > {
+    method: LiteralUnion<'GET' | 'PUT' | 'POST' | 'DELETE' | 'HEAD' | 'PATCH' | 'OPTIONS' | 'TRACE' | 'CONNECT'>;
```
tajakobsen, line 104: This is actually a list of HTTP verbs you can use in the Java-part of the system:
https://github.com/enonic/xp/blob/52ba85a552c09823bcd499487d824a7a9abe0b98/modules/web/web-api/src/main/java/com/enonic/xp/web/HttpMethod.java#L7

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
ComLock: > naming collision against the global fields in the web platform
Do you mean tsconfig lib: ['DOM']?

I think using request is fine, because we are talking about Enonic XP server-side files.

It's a bit tricky to "remove" the DOM from default environment.

I found that adding this:
node_modules/typescript/lib/lib.es5.d.ts

To the includes files does the trick.

It should probably be used in combination with nolib: true

PR #9751, opened 2022-09-30
Core Libraries shared types imports #9729

edloidas:
Replaced local types with core types.
Updated lib root dependencies.
Removed unused lib root dependencies.
Updated build.

### DIFF
--- a/modules/lib/lib-content/src/main/resources/lib/xp/content.ts
+++ b/modules/lib/lib-content/src/main/resources/lib/xp/content.ts
@@ -17,11 +17,11 @@ import type {Attachment, Component, Content, PublishInfo, Region} from '@enonic-
 
 export type {Attachment, PublishInfo, Content, Component, Region} from '@enonic-types/core';
 
-export type Attachments = Content['attachments'];
+type Attachments = Content['attachments'];
 
-export type ContentInheritType = Content['inherit'];
+type ContentInheritType = Content['inherit'];
 
-export type Workflow = Content['workflow'];
+type Workflow = Content['workflow'];
 
 export type Schedule = Omit<PublishInfo, 'first'>;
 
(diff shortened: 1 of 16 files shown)


modules/lib/lib-content/src/main/resources/lib/xp/content.ts
```diff
@@ -11,12 +11,20 @@ declare global {
...
 
+import type {Attachment, Component, Content, PublishInfo, Region} from '@enonic-types/core';
+
+export type {Attachment, PublishInfo, Content, Component, Region} from '@enonic-types/core';
+
+export type Attachments = Content['attachments'];
+
+export type ContentInheritType = Content['inherit'];
+
+export type Workflow = Content['workflow'];
```
tajakobsen, line 24: I would not export these types from the library. I would expect the app-developers to be able to create these types as they need them.

Maybe I would add an example in the documentation on how to do this.

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -487,21 +431,14 @@ interface GetCurrentContentHandler {
  *
  * @returns {object|null} The current content as JSON.
  */
-export function getContent<Data = Record<string, unknown>, Type extends string = string>(): Content<Data, Type> | null {
+export function getContent<
+    Data = Record<string, unknown>,
+    Type extends string = string,
+    Config extends object = object,
+    Regions extends Record<string, Region> = Record<string, Region>
```
tajakobsen, line 438: Yes! This is the best way to implement this for simple usage by the app developer. It's because when creating a controller for a page (not a part etc), the common thing to do is to grab the config from `content.page` that we got from `portalLib.getContent()`.

tajakobsen (approved):
I think this is looking very good! Good job! :+1:
I'm approving this, but the only thing I would do different is not to export the derived types in `contentLib`. Otherwise I'm very happy with this.

alansemenov (approved):

rymsha (approved):

Issue #9757, opened 2022-10-03
application getApplicationMode method instead of hasVirtua/hasReal


rymsha:
`getApplicationMode(appKey)` : `bundled` | `virtual` | `augmented`

PR #9764, opened 2022-10-04
application getApplicationMode method instead of hasVirtua/hasReal #9757

vbradnitski:


modules/lib/lib-app/src/main/resources/lib/xp/examples/app/getApplicationMode.js
```diff
@@ -5,14 +5,14 @@ var assert = require('/lib/xp/testing');
 
 // BEGIN
 // Create virtual app.
```
rymsha, line 7:
```suggestion
// Get Application mode.
```

rymsha (changes requested):

rymsha (approved):

modules/lib/lib-app/src/main/resources/lib/xp/app.ts
```diff
@@ -188,28 +188,28 @@ export function hasVirtual(params: HasVirtualApplicationParams): boolean {
...
+export interface GetApplicationModeParams {
     key: string;
 }
 
-interface HasRealApplicationHandler {
+interface GetApplicationModeHandler {
     setKey(value: string): void;
 
-    execute(): boolean;
+    execute(): string;
```
anatol-sialitski, line 198: `string | null` instead `string`

Issue #9040, opened 2021-08-25
The start-up script reports the wrong date on the commit


jsi:
When XP starts, it logs a lot of info, including the last commit in the code:
`# Built on 2021-08-23T07:50:30Z (hash = f36a2151a1b53e4de97c5bc050fbfe22169e995e, branch = master)`

The date is wrong.  In this case, a code change was made in a branch on Monday, August 23rd.  The commit for this was `e553389` - Two days later, the commit was approved and merged to master in a commit with the hash above (`f36a215`).

This is a minor issue, since releases always are based on a change in the build script, setting a new version, and this is directly committed to the build branch, so then the last commit to master and its date is matched.
However, in situations like this, where the commit is used for testing, the wrong date can cause confusion about what code is actually running, so if the startup script can be changed, so that the date matches the merge to master, that would be helpful.

PR #9765, opened 2022-10-04
The start-up script reports the wrong date on the commit #9040

anatol-sialitski:


rymsha (changes requested):
lets try Commiter time instead

rymsha (approved):

vbradnitski (approved):

Issue #9766, opened 2022-10-04
Content library type definitions contain invalid core reference


edloidas:
When the content.ts compiled into the `content.d.ts`, a reference to core library is included:
```ts
/// <reference types="core" />
```
This leads to errors while using this package.
The reason for this if the usage of `XpXData` interface in the imported type `CreateContentParams`. To prevent the reference injection, a local interface must be added to the global scope:
```ts
declare global {
    interface XpXData {
        [key: string]: Record<string, Record<string, unknown>>
    }
}
```

PR #9767, opened 2022-10-04
Content library type definitions contain invalid core reference #9766

edloidas:
Added global `XpXData` interface to prevent reference injection into the DTS file on compile.
Removed unused imports.


tajakobsen (approved):
Looks good

rymsha (approved):

Issue #9780, opened 2022-10-11
Fix node lib queries 


edloidas:
* [x] Boolean query DSL must allow `boost` parameter.
* [x] Boolean query DSL must accept any query type as values for it's properties.
* [x] `refresh` method arguments must be optional.

PR #9781, opened 2022-10-11
Fix node lib queries #9780

edloidas:


ComLock (approved):

modules/lib/lib-node/src/main/resources/lib/xp/node.ts
```diff
@@ -838,7 +827,7 @@ export interface RepoConnection {
 
     findChildren(params: FindChildrenParams): FindNodesByParentResult;
 
-    refresh(mode: RefreshMode): void;
+    refresh(mode?: RefreshMode): void;
```
anatol-sialitski, line 830: @edloidas why `mode` is optional here? If, yes, then it should be marked as optional in ` NodeHandler` too

ComLock: Because it has always been optional, since it has a default.

anatol-sialitski: I agree for JS API, but this is projection on Java, where mode can not be optional.

edloidas: It is optional in interface. In `RepoConnectionImpl` class it has a default value `'ALL'` (we cant have a parameter initializer in interfaces). Keeping `mode` non-optional in the `NodeHandler` makes sure we always pass a value there

anatol-sialitski: `RepoConnection.refresh(mode?: RefreshMode)`

edloidas: Yes, `RepoConnection` is the interface that is basically goes to types definitions after compilation. The `mode` here is optional because in actual implementation `RepoConnectionImpl` it has a default value.
`NodeHandler` relates to java code. There we have no default value for the mode, so it must be required here.

tajakobsen (approved):
Looks good.

alansemenov (approved):

rymsha (approved):

Issue #9805, opened 2022-10-17
NodeMultiRepoQueryResult type is wrong


ComLock:
(TypeScript doesn’t merge nested properties)

PR #9806, opened 2022-10-17
NodeMultiRepoQueryResult type is wrong #9805

ComLock:
…ed properties) #9805

### DIFF
--- a/modules/lib/lib-node/src/main/resources/lib/xp/node.ts
+++ b/modules/lib/lib-node/src/main/resources/lib/xp/node.ts
@@ -427,7 +427,7 @@ export interface NodeQueryResult {
     suggestions?: NodeQueryResultSuggestion[];
 }
 
-export type NodeMultiRepoQueryResult = {
+export interface NodeMultiRepoQueryResult {
     total: number;
     count: number;
     hits: (NodeQueryResultHit & {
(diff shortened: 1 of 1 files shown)


tajakobsen (approved):
It looks good. I don't have any feedback here.

modules/lib/lib-node/src/main/resources/lib/xp/node.ts
```diff
@@ -399,35 +399,43 @@ export interface Explanation {
...
-                freq?: number; // only for term
-            }[];
-        }[];
-    }[];
+    suggestions?: NodeQueryResultSuggestion[];
 }
 
-export type NodeMultiRepoQueryResult = NodeQueryResult & {
-    hits: {
+export type NodeMultiRepoQueryResult = {
```
anatol-sialitski, line 430: Maybe will be better to declare NodeMultiRepoQueryResult as interface instead of type here?

ComLock: That's up to ye, I didn't change it :)

edloidas: Yes, let's change it to the `interface` here, since it is no longer a union type.

edloidas (approved):

tajakobsen (approved):

modules/lib/lib-node/src/main/resources/lib/xp/node.ts
ComLock: Fixed, I assume @rymsha will do the merging?

tajakobsen (approved):

anatol-sialitski (approved):

Issue #9808, opened 2022-10-18
Some functions return incorrect Content type


alansemenov:
Many of the functions that return Content should return Content<A, 'a'> | Content<B, 'b'> instead of Content<A | B, 'a' | 'b'> .
The difference is that the first way allows us to use if -statements on content.type to split the unions in application code.

PR #9819, opened 2022-10-20
TypeScript type for Union of Content #9808

tajakobsen:
The return type of `contentLib.query()` (and similar functions) should have the shape `Content<A, 'a'> | Content<B, 'b'>` instead of the current implementation which is `Content<A | B, 'a' | 'b'>`.

This change allows TypeScript to infer the correct type of content by using a simple if-statement against `Content.type`, which is the natural way for an application programmer to do this.

### DIFF
--- a/modules/lib/lib-content/src/main/resources/lib/xp/content.ts
+++ b/modules/lib/lib-content/src/main/resources/lib/xp/content.ts
@@ -55,7 +55,7 @@ interface GetContentHandler {
 
     setVersionId(value?: string | null): void;
 
-    execute<Data, Type extends string>(): Content<Data, Type> | null;
+    execute<Hit extends Content<unknown>>(): Hit | null;
 }
 
 interface GetAttachmentsHandler {
@@ -542,7 +542,7 @@ export type SortDsl = FieldSortDsl | GeoDistanceSortDsl;
  *
  * @returns {object} The content (as JSON) fetched from the repository.
  */
-export function get<Data = Record<string, unknown>, Type extends string = string>(params: GetContentParams): Content<Data, Type> | null {
+export function get<Hit extends Content<unknown> = Content>(params: GetContentParams): Hit | null {
     checkRequired(params, 'key');
 
     const bean = __.newBean<GetContentHandler>('com.enonic.xp.lib.content.GetContentHandler');
@@ -550,7 +550,7 @@ export function get<Data = Record<string, unknown>, Type extends string = string
     bean.setKey(params.key);
     bean.setVersionId(__.nullOrValue(params.versionId));
 
-    return __.toNativeObject(bean.execute<Data, Type>());
+    return __.toNativeObject(bean.execute<Hit>());
 }
 
 /**
@@ -795,10 +795,10 @@ export {
     _delete as delete,
 };
 
-export interface ContentsResult<Data, Type extends string> {
+export interface ContentsResult<Hit extends Content<unknown>> {
     total: number;
     count: number;
-    hits: Content<Data, Type>[];
+    hits: Hit[];
     aggregations?: Record<string, AggregationsResult>;
     highlight?: Record<string, HighlightResult>;
 }
@@ -819,7 +819,7 @@ interface GetChildContentHandler {
 
     setSort(value?: string | null): void;
 
-    execute<Data, Type extends string>(): ContentsResult<Data, Type>;
+    execute<Hit extends Content<unknown>>(): ContentsResult<Hit>;
 }
 
 /**
@@ -835,7 +835,7 @@ interface GetChildContentHandler {
  *
  * @returns {Object} Result (of content) fetched from the repository.
  */
-export function getChildren<Data = Record<string, unknown>, Type extends string = string>(params: GetChildContentParams): ContentsResult<Data, Type> {
+export function getChildren<Hit extends Content<unknown> = Content>(params: GetChildContentParams): ContentsResult<Hit> {
     checkRequired(params, 'key');
 
     const {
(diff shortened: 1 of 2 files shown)


tajakobsen:
I have not changed the shape of the functions that `create` or `modify` or `move` content. They basically operate on a single content that we know the type of, and doesn't have to deal with `ContentUnion`.

The consequence is that some functions take a **tuple** with the shape of the content, and the name of the type – and some functions just take in the shape and the name as two parameters.

I'm happy with this. But I would like to hear what you are thinking about it.

anatol-sialitski (approved):

ComLock:
I tried to use it here:
https://github.com/enonic/site-developer/blob/issue-891/typeScript/src/main/resources/site/parts/cards/cards.ts#L66-L70

But it says:

```
src/main/resources/site/parts/cards/cards.ts:67:3 - error TS2344: Type 'ContentTypeArticle | ContentTypeDoc | ContentTypeGuide' does not satisfy the constraint 'readonly [data: unknown, type: string]'.
  Type 'ContentTypeArticle' is not assignable to type 'readonly [data: unknown, type: string]'.
```
  
  Haven't studied it closely, running to lunch.

edloidas:
@ComLock That was actually the main point to return the correct Content type. BTW you are passing the Content, we actually accept Data and Type.

**Previously** we had this:
```ts
// definition
export function query<Data = Record<string, unknown>, Type extends string = string>(params: QueryContentParams): ContentsResult<Data, Type> { ... }
// usage
query<Article | Employee, "article" | "employee">({query: '...'}); 
// will return result with Content<Article | Employee, "article" | "employee">
// This is not correct!
```

**Now** we have this:
```ts
// definition
export function query<DataAndType extends DataAndTypeTuple = readonly [data: Record<string, unknown>, type: string]>(params: QueryContentParams): ContentsResult<DataAndType> { ... }
// usage
query<[Article, "article"] | [Employee, "employee"]>({query: '...'});
// will return result with Content<Article, "article"> | Content<Employee, "employee">
````

**Comment by edloidas** [comment]
@tajakobsen I'm okay with functions that operates on a single element (`create`, `modify`, `move`) not using `ContentUnion`.

File: modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts [hunk 1]
```diff
@@ -439,13 +439,12 @@ interface GetCurrentContentHandler {
  * @returns {object|null} The current content as JSON.
  */
 export function getContent<
-    Data = Record<string, unknown>,
-    Type extends string = string,
+    DataAndType extends DataAndTypeTuple = readonly [data: Record<string, unknown>, type: string],
```
edloidas, line 442: Same question here.

modules/lib/lib-content/src/main/resources/lib/xp/content.ts
```diff
@@ -1008,7 +1008,7 @@ interface QueryContentHandler {
  *
  * @returns {object} Result of query.
  */
-export function query<Data = Record<string, unknown>, Type extends string = string>(params: QueryContentParams): ContentsResult<Data, Type> {
+export function query<DataAndType extends DataAndTypeTuple = readonly [data: Record<string, unknown>, type: string]>(params: QueryContentParams): ContentsResult<DataAndType> {
```
edloidas, line 1011: Are you sure we want a Tuple here? It always return a single Content, so it might be consistent to keep <Data, Type>, like with functions that operate on with single value.

tajakobsen: This is `query()` which takes in `contentTypes` as a parameter, so the results can contain different content types.

See `getContent()` for the more interesting answer :)

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
tajakobsen: So, while `getContent()` will only return one answer, we don't necessarily know the content type in advance.

Imagine a part that can be used on different content types. We don't know which exact type `getContent` will return, but we can use if-statements to handle each possibility (or return e.g. a 404 if it was none of them).

edloidas (approved):

ComLock:
Why do one have to say the same thing twice? Both the type and the type of the type? Doesn't the type contain it's type? Just use a type guard on the result?

ComLock:
Each time you light your lighter your lighter gets lighter until your lighter gets so light that it won't light

tajakobsen:
> Why do one have to say the same thing twice?

 1. The first field in the tuple gives the shape of `content.data`, so that we can program against it. 
 2. The second field of the tuple contains the name of the content type which sets `content.type` to the *string literal* (e.g `myapp:Article`), which allows the application developer to use a simple if-statement to separate the union of possible types the result can be.

The combination of `content.data` and `content.type` lets the application developer – in a typesafe way – write TypeScript-code that is very close to the natural JavaScript-code, but with full type safety:

Imagine a **part** that can be used on two content types, Article and Employee:

```typescript
interface Article {
  title: string;
}

interface Employee {
  fullName: string;
}

// the type parameter does the heavy lifting in describing the possible shapes of content
const content = getContent<[Article, 'myapp:Article'] | [Employee, 'myapp:Employee']>();

const title = (content.type === "myapp:Article")
  ? content.data.title  // Because of the tuple TypeScript knows the shape the `data` that belongs to "myapp:Article" 
  : content.data.fullName; // And it knows if it isn't "myapp:Article", it has have the shape of "myapp:Employee"

// no other fields on `content.data` would be allowed to be used
```

edloidas (approved):

ComLock:
Nice. I've only tested the query.

Hovering hit it looks perfect:
Screenshot 2022-10-24 at 14 32 20

data.repository also looks perfect:
Screenshot 2022-10-24 at 14 34 12

Weird that data looks perfect here:
Screenshot 2022-10-24 at 14 36 04

But rather complex here:
Screenshot 2022-10-24 at 14 36 40

Maybe just my editor?

ComLock (approved):

Issue #9828, opened 2022-10-26
Invalid portal lib types 


edloidas:
* `ImageUrlParams` must not allow `id` and `path` properties at the same time.
* `type` and `contextPath` properties are too wide. narrow them to the specific values.

tajakobsen:
> `ImageUrlParams` must not allow `id` and `path` properties at the same time.

The problem is not just allowing both at the same time, but *requiring* both a t the same time.

PR #9830, opened 2022-10-26
Invalid portal lib types #9828

edloidas:


modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -27,10 +27,15 @@ export interface SiteConfig<Config> {
     config: Config;
 }
 
+export type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
+export type XOR<T, U> = T | U extends object ? (Without<T, U> & U) | (Without<U, T> & T) : T | U;
+
+export type IdXorPath = XOR<{id: string}, {path: string}>;
```
tajakobsen, line 33: Correct me if I'm wrong, but maybe this could be expressed in a slightly simpler way, without the `XOR` abstraction:

```suggestion
export type IdXorPath = 
  | {
    id: string;
    path?: never;
  }
  | {
    id?: never;
    path: string;
  };
```

tajakobsen (changes requested):
Looks good. But I have a slight suggestion on a possible solution that lets us drop the `XOR` type.

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
edloidas: That's what I actually did. Unfortunately, IDE/Editor shows the `path` as `undefined` in the suggestions box, so I though that in the end this more complex notation will be better for the developers using types.

tajakobsen: Ok, I will leave this choice up to you. So I will approve this PR.

tajakobsen (approved):

Issue #9838, opened 2022-10-28
Improve getProfile return type based on key provided


edloidas:
We have a clear distinction between principal key, so we can determine what principal type will be returned based on the key provided.

PR #9839, opened 2022-10-28
Improve getProfile return type based on key provided #9838

edloidas:


tajakobsen (approved):
Very nice! I like this!

anatol-sialitski (approved):

Issue #9843, opened 2022-10-31
Portal library: index param in lib functions must be optional


edloidas:
`index` parameter in `getMultipart*` is marked optional in the documentation, but it is required in the functions.

PR #9846, opened 2022-10-31
Portal library: index param in lib functions must be optional #9843

edloidas:


modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -526,7 +526,7 @@ export function getMultipartForm(): MultipartForm {
  *
  * @returns {object|null} The named multipart form item.
  */
-export function getMultipartItem(name: string, index: number): MultipartItem | null {
+export function getMultipartItem(name: string, index?: number): MultipartItem | null {
     const bean = __.newBean<MultipartHandler>('com.enonic.xp.lib.portal.multipart.MultipartHandler');
     return __.toNativeObject(bean.getItem(name, index ?? 0));
 }
```
tajakobsen, line 532: I prefer this syntax, as it ties the default value to the variable to the `index` variable.

```suggestion
export function getMultipartItem(name: string, index: number = 0): MultipartItem | null {
    const bean = __.newBean<MultipartHandler>('com.enonic.xp.lib.portal.multipart.MultipartHandler');
    return __.toNativeObject(bean.getItem(name, index));
}
```

tajakobsen (changes requested):
I just have a suggestion of an alternative syntax. Nothing blocking.

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
edloidas: Yeah, the default value will look better.

tajakobsen (approved):
Looks good.

Issue #9850, opened 2022-11-02
Node and Content libs: wrong query results


edloidas:
`StatsAggregation` interface is duplicated. 
The query results has wrong type for aggregations and suggestions.

PR #9851, opened 2022-11-02
Node and Content libs: wrong query results #9850

edloidas:
Types for aggregations and suggestions in result and in params were incorrect. I tried to make them match the Java implementation and work with JS examples, but I'm not 100% everything is correct, so it's better to test them on some real data before actually releasing.

What were changed:
* `StatsAggregation` interface was declared twice and used both in result and in params types. The stats aggregation in result is now called `StatsAggregationResult`.
* `suggestions` in result type was invalid, according to the [examples](https://github.com/enonic/xp/blob/master/modules/lib/lib-node/src/main/resources/lib/xp/examples/node/query.js#L175).
* Same for the `aggregations`. See [examples](https://github.com/enonic/xp/blob/master/modules/lib/lib-node/src/main/resources/lib/xp/examples/node/query.js#L129).

### DIFF
--- a/modules/lib/lib-content/src/main/resources/lib/xp/content.ts
+++ b/modules/lib/lib-content/src/main/resources/lib/xp/content.ts
@@ -63,7 +63,7 @@ interface GetAttachmentsHandler {
 }
 
 export interface Bucket {
-    [subAggregationName: string]: BucketsAggregation | StatsAggregation | SingleValueMetricAggregation | string | number | undefined;
+    [subAggregationName: string]: AggregationsResult | string | number | undefined;
 
     key: string;
     docCount: number;
@@ -81,11 +81,11 @@ export interface DateBucket
     to?: string;
 }
 
-export interface BucketsAggregation {
+export interface BucketsAggregationResult {
     buckets: (DateBucket | NumericBucket)[];
 }
 
-export interface StatsAggregation {
+export interface StatsAggregationResult {
     count: number;
     min: number;
     max: number;
@@ -93,11 +93,11 @@ export interface StatsAggregation {
     sum: number;
 }
 
-export interface SingleValueMetricAggregation {
+export interface SingleValueMetricAggregationResult {
     value: number;
 }
 
-export type AggregationsResult = BucketsAggregation | StatsAggregation | SingleValueMetricAggregation;
+export type AggregationsResult = BucketsAggregationResult | StatsAggregationResult | SingleValueMetricAggregationResult;
 
 export type Aggregation =
     | TermsAggregation
@@ -118,9 +118,7 @@ export interface TermsAggregation {
         size?: number;
         minDocCount?: number;
     };
-    aggregations?: {
-        [subAggregations: string]: Aggregation;
-    };
+    aggregations?: Record<string, Aggregation>;
 }
 
 export interface HistogramAggregation {
@@ -132,9 +130,7 @@ export interface HistogramAggregation {
         extendedBoundMax?: number;
         minDocCount?: number;
     };
-    aggregations?: {
-        [subAggregations: string]: Aggregation;
-    };
+    aggregations?: Record<string, Aggregation>;
 }
 
 export interface DateHistogramAggregation {
@@ -144,9 +140,7 @@ export interface DateHistogramAggregation {
         minDocCount?: number;
         format: string;
     };
-    aggregations?: {
-        [subAggregations: string]: Aggregation;
-    };
+    aggregations?: Record<string, Aggregation>;
 }
 
 export interface NumericRange {
@@ -160,9 +154,7 @@ export interface NumericRangeAggregation {
         field: string;
         ranges?: NumericRange[];
     };
-    aggregations?: {
-        [subAggregations: string]: Aggregation;
-    };
+    aggregations?: Record<string, Aggregation>;
 }
 
 export interface NumericRangeAggregation {
@@ -170,9 +162,7 @@ export interface NumericRangeAggregation {
         field: string;
         ranges?: NumericRange[];
     };
-    aggregations?: {
-        [subAggregations: string]: Aggregation;
-    };
+    aggregations?: Record<string, Aggregation>;
 }
 
 export interface DateRange {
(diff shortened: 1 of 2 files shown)


modules/lib/lib-node/src/main/resources/lib/xp/node.ts
```diff
@@ -641,7 +637,7 @@ export interface QueryNodeParams {
     sort?: string | SortDsl | SortDsl[];
     filters?: Filter | Filter[];
     aggregations?: Record<string, Aggregation>;
-    suggestions?: Record<string, Suggestion>;
+    suggestions?: Record<string, TermSuggestion>;
```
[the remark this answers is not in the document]
edloidas: The reason why I removed `Suggestion` type and placed `TermSuggestion` here was that it was basically an alias for that. Having two same types with different names exported doesn't look very wise. Since we haven't released any stable types yet, it'll be better to export as less as possible if we are not sure how this types will look in future releases. It's always easier to add types, than remove them, because in the second case it can break stuff.

ComLock:
durationStats doesn't appear in the expected, so it's untested?
https://github.com/enonic/xp/blob/master/modules/lib/lib-node/src/main/resources/lib/xp/examples/node/query.js#L66
https://github.com/enonic/xp/blob/master/modules/lib/lib-node/src/main/resources/lib/xp/examples/node/query.js#L110

ComLock:
ASFAIK Stats Aggregations can only be used as a subAggregation, so that might be an issue.

ComLock (approved):

modules/lib/lib-content/src/main/resources/lib/xp/content.ts
```diff
@@ -81,23 +81,23 @@ export interface DateBucket
     to?: string;
 }
 
-export interface BucketsAggregation {
+export interface BucketsAggregationResult {
     buckets: (DateBucket | NumericBucket)[];
 }
```
tajakobsen, line 86: I mentioned earlier that it is a bit difficult to split `AggregationsResult`, since it's not a discriminate union (yet). 

And here we have another union (`DateBucket | NumericBucket`) inside it again, which is also a bit difficult to split. 

I would suggest changing this code, so that we only need to do one splitting of unions, and that is on the `AggregationsResult` level.

```suggestion
export interface DateBucketsAggregationResult {
    buckets: DateBucket[];
}

export interface NumberBucketsAggregationResult {
    buckets: NumericBucket[];
}
```

and then later:

```typescript
export type AggregationsResult = 
  | DateBucketsAggregationResult 
  | NumberBucketsAggregationResult 
  | StatsAggregationResult 
  | SingleValueMetricAggregationResult;
```

tajakobsen (changes requested):
I don't feel I have so much to say regarding correctness of this code. But I have one suggestion related to developer ergonomics that can be considered.

edloidas:
We have a working [test1](https://github.com/enonic/xp/blob/master/modules/lib/lib-content/src/test/resources/test/QueryContentHandlerTest.js#L220) and [test2](https://github.com/enonic/xp/blob/master/modules/lib/lib-content/src/test/resources/test/QueryContentHandlerTest_dsl_query.js#L229), where the Stats are used for top level aggregations, so I'm not sure it is for sub-aggregations only:
Maybe @rymsha or @anatol-sialitski can tell more about this.

ComLock:
"The stats-aggregations calculates the following statistics for the parent-aggregation buckets:"
https://developer.enonic.com/docs/xp/stable/storage/aggregations#stats

edloidas:
@ComLock In the files you've mention, StatsAggregation is not present in the results, so maybe those examples must be updated? Yeah, I saw that in docs too, but they must be misleading. Let's wait for more feedback from core devs.

ComLock:
According to ES docs, stats can be a top-level aggregation, so that could be an error in our documentation.

rymsha:
> "The stats-aggregations calculates the following statistics for the parent-aggregation buckets:" https://developer.enonic.com/docs/xp/stable/storage/aggregations#stats

ES can aggregate 
https://www.elastic.co/guide/en/elasticsearch/reference/2.4/search-aggregations-metrics-stats-aggregation.html

So, XP docs are not correct.

ComLock:
Looking at the types, I can see that StatsAggregation, GeoDistanceAggregation, MinAggregation, MaxAggregation and ValueCountAggregation cannot have SubAggregations. We might want to reflect that in the output types aswell?

edloidas:
@ComLock Well, they don't have sub-aggregations if you look at the code. Or am I missing something?

modules/lib/lib-content/src/main/resources/lib/xp/content.ts
edloidas: Not sure we can split those buckets. 
@rymsha can we return the mixed buckets, list of both `DateRangeBucket` and `NumericRangeBucket`? It's hard to say from the [mapper](https://github.com/enonic/xp/blob/master/modules/lib/lib-node/src/main/java/com/enonic/xp/lib/node/mapper/AggregationMapper.java#L70)? We can return, at least.

ComLock:
What code, the Java code? No I have not looked at it. But I agree that those probably don't have subAggregations.
Which is reflected in the aggregation input types, but not in the aggregation output types.

Unless one says that the undefined here:
https://github.com/enonic/xp/blob/71cb937e1e88093d8e00324279d0dc03ab874994/modules/lib/lib-content/src/main/resources/lib/xp/content.ts#L66

Is good enough.

TypeScript is not Java so aiming for perfection might not be "cost effective", I think there's such a thing as "good enough" for now and "improve if needed later".

edloidas:
I meant all those types in `node.ts` and `content.ts`. The "input" types you've mentioned, they don't have sub aggregations. But the "output" type `Bucket` indeed have. It'll be risky trying to match input with output, at least for this release, so the "output" have Java-like, more loose type, that is not affected by the input.

tajakobsen (approved):

alansemenov (approved):

rymsha (approved):

Issue #9871, opened 2022-11-11
Project Timezone API


rymsha:
Project new field timeZone. 
- Type of `ZoneId`. 
- Indexed as string.
- optional. No-value has no default.

PR #9872, opened 2022-11-11
Project Timezone API #9871

anatol-sialitski:


modules/core/core-api/src/main/java/com/enonic/xp/project/CreateProjectParams.java
```diff
@@ -75,6 +80,11 @@ public boolean isForceInitialization()
         return forceInitialization;
     }
 
+    public String getTimezone()
```
rymsha, line 83:
```suggestion
    public TimeZone getTimeZone()
```

rymsha (changes requested):

modules/core/core-api/src/main/java/com/enonic/xp/project/CreateProjectParams.java
anatol-sialitski: After discuss we decided to use `ZoneId`

modules/core/core-project/src/main/java/com/enonic/xp/core/impl/project/ProjectServiceImpl.java
```diff
@@ -466,6 +467,7 @@ private PropertyTree createProjectData( final CreateProjectParams params )
 
         projectData.setString( ProjectConstants.PROJECT_DESCRIPTION_PROPERTY, params.getDescription() );
         projectData.setString( ProjectConstants.PROJECT_DISPLAY_NAME_PROPERTY, params.getDisplayName() );
+        projectData.setString( ProjectConstants.PROJECT_TIMEZONE_PROPERTY, Objects.toString( params.getTimeZone(), null ) );
```
rymsha, line 470: don't set if null

modules/lib/lib-project/src/main/java/com/enonic/xp/lib/project/CreateProjectHandler.java
```diff
@@ -135,4 +139,9 @@ public void setSiteConfig( final ScriptValue value )
         final List<SiteConfig> configs = buildSiteConfigs( value );
         this.siteConfigs = configs != null ? SiteConfigs.from( configs ) : null;
     }
+
+    public void setTimeZone( final String timeZone )
+    {
+        this.timeZone = timeZone != null ? ZoneId.of( timeZone, ZoneId.SHORT_IDS ) : null;
```
rymsha, line 145: don't use SHORT_IDS

rymsha (changes requested):

rymsha (approved):

Issue #9877, opened 2022-11-16
TypeScript createMedia data is string, should be object


ComLock:
https://github.com/enonic/xp/blob/master/modules/lib/lib-content/src/main/resources/lib/xp/content.ts#L1198
https://developer.enonic.com/docs/xp/stable/api/lib-content#createmedia

PR #9878, opened 2022-11-16
TypeScript createMedia data is string, should be object #9877

ComLock:


modules/lib/core/index.d.ts
```diff
@@ -99,12 +99,19 @@ export interface Content<
     inherit?: ('CONTENT' | 'PARENT' | 'NAME' | 'SORT')[];
 }
 
+// com.google.common.io.ByteSource
+export interface ByteSource {
```
rymsha, line 103: Both exposed methods are dangerously inefficient and must not be exposed in JS API.

rymsha (changes requested):

modules/lib/core/index.d.ts
tajakobsen: The point of this interface (when I first wrote it) – was to create something that was "different" then `object`, just so that conflicts would arise if something other then `ByteSource` was passed into a function.

The methods were never really intended to be used. But I wanted to pick some that really existed on `com.google.common.io.ByteSource`.

If you don't want to expose any of the methods. You can create this like a [branded/tagged](https://medium.com/@KevinBGreene/surviving-the-typescript-ecosystem-branding-and-type-tagging-6cf6e516523d) type.

```typescript
export interface ByteSource {
  type: 'ByteSource'
}

// or

export interface ByteSource {
  __javaClass: 'com.google.common.io.ByteSource'
}

// or

type Brand<K, T> = K & { __brand: T }
type ByteSource = Brand<object, "com.google.common.io.ByteSource">

// or

export interface ByteSource {
  _brand: 'ByteSource'
}
```

You just have to make sure that this interface will always conflict with everything that doesn't match.

tajakobsen: I'm doing some branded versions of Java-classes in graphQL-lib here: https://github.com/ItemConsulting/enonic-types/blob/main/packages/graphql/index.d.ts#L47-L100

ComLock: > @rymsha  "must not be exposed in JS API"
Isn't that the case already though?

So you rather meant "must not be revealed to be exposed by our TypeScript types"?

rymsha: In XP 8. This is will no longer be com.google.common.io.ByteSource . Nether there will be the methods isEmpty/size 
So in this particular case yes, I meant that Typescript types should not expose these methods

tajakobsen (approved):

edloidas (approved):

tajakobsen (approved):
Looks good

ComLock:
So I wrote some type tests, just to show the difference between using object and ByteSource
https://github.com/enonic/enonic-js-utils/blob/main/test/types/ByteSource.test-d.ts#L60-L66

@rymsha I could potentially write tests for all the affected API functions. How much you want?

rymsha (approved):

PR #9884, opened 2022-11-21
TypeScript: Roles and Users #9883

ComLock:


modules/lib/core/index.d.ts
```diff
@@ -4,11 +4,65 @@ declare global {
...
+	Login extends string = string
+> = `user:${IdProvider}:${Login}`;
 
-export type PrincipalKey = UserKey | GroupKey | RoleKey;
+export type GroupKey<
+	IdProvider extends string = string,
+	Name extends string = string
+> = `group:${IdProvider}:${Name}`;
+
+export type RoleKey<T extends string = string> = `role:${T}`;
```
tajakobsen, line 17: I'm not a fan of these changes to `UserKey`, `GroupKey` and `RoleKey`. I think they become more complex  to read and reason about.

modules/lib/core/index.d.ts
```diff
@@ -4,11 +4,65 @@ declare global {
...
+
+export namespace Roles {
```
tajakobsen, line 19: Are you sure that you want to start to introduce `namespaces`? You have not used them elsewhere in the code, so you are introducing a new way of doing things here.

This pattern could also be used:

```typescript
export type RoleKeySystemAdmin = "role:system.admin";
```

modules/lib/core/index.d.ts
```diff
@@ -4,11 +4,65 @@ declare global {
...
+    // | Roles.Cms.Cm.App
+    // | Roles.Cms.Expert
+    | Roles.System.Admin
+    | Roles.System.Admin.Login
+    | Roles.System.Auditlog
+    | Roles.System.Authenticated
+    | Roles.System.Everyone
+    | Roles.System.User.Admin
+    | Roles.System.User.App
+    | Users.System.Su
```
tajakobsen, line 64: Are these providing code completion at the moment? I thought they would need to use that "literal" util-type.

tajakobsen (commented):
Good job! I only have a few discussion points to look at.

modules/lib/core/index.d.ts
ComLock: Sure, no. In @enonic/js-utils I did the same without namespaces.

The nice thing is that you can import all the roles with a single

`import type {Roles} from '/lib/xp/auth';`

And then it will autocomplete through the namespaces.

But the drawback is you can't import a single type

This doesn't work

`import type {Roles.System.Admin} from '/lib/xp/auth';`

ComLock: Would be cool if import supported deconstruction.

```
import type {
  Roles: {
    System: {
      Admin
    }
  }
} from '/lib/xp/auth';
```

modules/lib/core/index.d.ts
ComLock: In @enonic/js-utils I made seperate Generics.

`export type RoleKeyGeneric<T extends string> = `role:${T}``

modules/lib/core/index.d.ts
ComLock: I suspect they don't.
What do you mean with `that "literal" util-type` ?

modules/lib/core/index.d.ts
ComLock: It was rather an experiment, since I wanted the discussion.

modules/lib/core/index.d.ts
ComLock: Perhaps these are even overkill, that literal strings are enough.
Such "extensions" could simply continue to live in @enonic/js-utils

modules/lib/core/index.d.ts
edloidas: I'm not against `namespace`. It keeps everything in one places and gets rid of lot of exports if we were to import everything separately.

BUT, I'm not sure we should provide predefined Principal keys at all. It's always XP specific. We don't provide  the types for the Node events, which in some way is similar.

My thought is that we must only provide a string literal types for code/framework specific data, and not just for something, that is later can be added/updated/removed by XP and/or the apps: `type DslQueryType = 'dateTime' | 'time';` is a go, Node events are not.

I guess @rymsha can add his thoughts on this too.

ComLock: Interesting. I've always been thinking of "Enonic XP" as the "core" product, but you're saying with regards to libs, Enonic XP just happens to be one of the integration points (data source/destination). So there could be another Enonic XP data shapes type library. And using it would be optional. In that case I think this whole PR should simply be closed :)

Issue #9888, opened 2022-11-23
TypeScript: Library types


ComLock:
Useful when making mock libraries

rymsha:
According to #9892 should be closed

PR #9892, opened 2022-11-23
TypeScript: Library types #9888

ComLock:
These are useful when making mockups of our libs.

I'm working in a *.d.ts file and I tried

```
import {
	create,
	createBranch,
	delete as deleteRepo,
	deleteBranch,
	get,
	list,
	refresh,
} from '/lib/xp/repo'

const repoLibrary = {
	create,
	createBranch,
	delete: deleteRepo,
	deleteBranch,
	get,
	list,
	refresh,
}

type RepoLibrary = typeof repoLibrary;
```

But that only results in

```
'const' initializer in an ambient context must be a string or numeric literal
```

So I'd much rather simply do:

```
import type {RepoLibrary} from '/lib/xp/repo';
```

### DIFF
--- a/modules/lib/lib-admin/src/main/resources/lib/xp/admin.ts
+++ b/modules/lib/lib-admin/src/main/resources/lib/xp/admin.ts
@@ -160,3 +160,19 @@ export function getInstallation(): string {
 export function getVersion(): string {
     return helper.getVersion();
 }
+
+const adminLibrary = {
+    getAssetsUri,
+    getBaseUri,
+    getHomeToolUrl,
+    getInstallation,
+    getLauncherPath,
+    getLauncherUrl,
+    getLocale,
+    getLocales,
+    getPhrases,
+    getToolUrl,
+    getVersion,
+};
+
+export type AdminLibrary = typeof adminLibrary;
--- a/modules/lib/lib-app/src/main/resources/lib/xp/app.ts
+++ b/modules/lib/lib-app/src/main/resources/lib/xp/app.ts
@@ -217,3 +217,15 @@ export function getApplicationMode(params: GetApplicationModeParams): string | n
     bean.setKey(params.key);
     return __.toNativeObject(bean.execute());
 }
+
+
+const appLibrary = {
+    createVirtualApplication,
+    deleteVirtualApplication,
+    get,
+    getApplicationMode,
+    getDescriptor,
+    list,
+};
+
+export type AppLibrary = typeof appLibrary;
--- a/modules/lib/lib-auditlog/src/main/resources/lib/xp/auditlog.ts
+++ b/modules/lib/lib-auditlog/src/main/resources/lib/xp/auditlog.ts
@@ -203,3 +203,11 @@ export function find(params: FindAuditLogParams): AuditLogs {
     bean.setObjectUris(__.toScriptValue(objects));
     return __.toNativeObject(bean.execute());
 }
+
+const auditLogLibrary = {
+    find,
+    get,
+    log,
+};
+
+export type AuditLogLibrary = typeof auditLogLibrary;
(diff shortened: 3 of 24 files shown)


ComLock:
This works, but hovering RepoLibrary doesn't show it's functions.

```
import * as repoLibrary from '/lib/xp/repo';

export type RepoLibrary = typeof repoLibrary;
```

modules/lib/lib-admin/src/main/resources/lib/xp/admin.ts
```diff
@@ -160,3 +160,19 @@ export function getInstallation(): string {
...
+    getLauncherPath,
+    getLauncherUrl,
+    getLocale,
+    getLocales,
+    getPhrases,
+    getToolUrl,
+    getVersion,
+};
+
+export type AdminLibrary = typeof adminLibrary;
```
tajakobsen, line 178: Is `AdminLibrary` not just the same as `XpLibraries["/lib/xp/admin"]`;

We cloud do this: 
```suggestion
export type AdminLibrary = XpLibraries["/lib/xp/admin"];
```

Or this
```suggestion
export type AdminLibrary = typeof import('./admin');
```

Or just use `XpLibraries["/lib/xp/admin"]` directly where we want to use that type.

tajakobsen (changes requested):
I have an alternative suggestion on how to solve this.

modules/lib/lib-admin/src/main/resources/lib/xp/admin.ts
ComLock: Screenshot 2022-11-23 at 14 34 09

Screenshot 2022-11-23 at 14 34 41

ComLock: Screenshot 2022-11-23 at 14 36 44

ComLock: I think the problem might be that import() with parentheses, is dynamic, while types should be static?

ComLock: Going the opposite way makes hovering even a import file path look nice.

Screenshot 2022-11-23 at 14 40 50

ComLock: Dynamic imports of types makes no sense. Dynamic imports are runtime, types are NOT runtime.

edloidas: I still can't figure out the original problem. What exactly are those changes suppose to fix? Do you mean that the type details in the editor are incorrect? Or am I missing something? Can you, please, provide an example of how you use this?

ComLock: So I would do this in all the libs:
Screenshot 2022-11-23 at 15 03 40

ComLock: This was originally about adding a feature not fixing anything.

But along the way I've come across the dynamic import issue.

And just now I found another interesting thing.
When switching from named exports to a default export object, I discovered we are exporting two extra undocumented functions.

Namely:
* modify
* getBinary

If we change from dynamic import to using the default export, then then I don't need any extras. 

Screenshot 2022-11-23 at 15 29 21

ComLock: This Pull Request should not hinder the release.

And the default export was just another experiment, I don't like it.

I need to be able to explain the original "problem", before a potential solution can be implemented.

However I still think the dynamic import is wrong, but could be fixed in another PR?

rymsha:
Shall we close this one?

tajakobsen:
> Shall we close this one?

I think this can be closed. I think this adds more manual maintenance without enough benefit.

ComLock:
Sure, let's close it for now. I'll probably come back to this later.

And as I said:
I still think the dynamic import is wrong, but could be fixed in another PR?

Issue #9855, opened 2022-11-07
Content library: wrong Permissions type


edloidas:
[Permissions](https://github.com/enonic/xp/blob/master/modules/lib/lib-content/src/main/resources/lib/xp/content.ts#L1368-L1371) type looks like this:
```ts
export interface Permissions {
    inheritsPermissions: boolean;
    permissions?: Permission[];
}
```
Must be like this (see [example](https://github.com/enonic/xp/blob/master/modules/lib/lib-content/src/main/resources/lib/xp/examples/content/getPermissions.js#L19-L30) or [docs](https://developer.enonic.com/docs/xp/stable/api/lib-content#getpermissions)):
```ts
export interface Permissions {
    inheritsPermissions: boolean;
    permissions?: AccessControlEntry[];
}
```

rymsha:
Am also not sure if permissions is optional

PR #9893, opened 2022-11-23
Content library: wrong Permissions type #9855

edloidas:


ComLock (approved):

modules/lib/lib-content/src/main/resources/lib/xp/content.ts
```diff
@@ -1378,7 +1374,7 @@ export interface SetPermissionsParams {
 
 export interface Permissions {
     inheritsPermissions: boolean;
-    permissions?: Permission[];
+    permissions: AccessControlEntry[];
```
rymsha, line 1377:
```suggestion
    permissions?: AccessControlEntry[];
```

Issue #9895, opened 2022-11-24
TypeScript: Node<> Generic: Too strict


ComLock:
If the Node<> Generic is supposed to describe the shape of a node:

It's fails because a node:
* doesn't have _parentPath
* only sometimes have _manualOrderValue

We have a CreateNodeParams Generic to represent the parameters to the connection.create() function.

I think we need one more Generic to represent what the the connection.modify() editor function is supposed to return.
The editor function takes a node in, but it should return a different shape.
There are differences in _indexConfig, Arrays, References and probably more.

PR #9897, opened 2022-11-25
TypeScript: Node<> Generic: Too strict #9895

ComLock:
I also wrote some tests https://github.com/enonic/enonic-js-utils/blob/main/test/types/Node.test-d.ts

### DIFF
--- a/modules/lib/lib-node/src/main/resources/lib/xp/node.ts
+++ b/modules/lib/lib-node/src/main/resources/lib/xp/node.ts
@@ -550,7 +550,7 @@ export type CreateNodeParams<NodeData = unknown> = NodePropertiesOnCreate & Node
 
 export interface ModifyNodeParams<NodeData = unknown> {
     key: string;
-    editor: (node: Node<NodeData>) => ModifyNode<NodeData>;
+    editor: (node: Node<NodeData>) => ModifiedNode<NodeData>;
 }
 
 export interface GetNodeParams {
@@ -832,7 +832,7 @@ export type NodePropertiesOnRead = CommonNodeProperties & {
     _parentPath?: never;
 };
 
-export type ModifyNode<Data = Record<string, unknown>> = NodePropertiesOnModify & Data;
+export type ModifiedNode<Data = Record<string, unknown>> = NodePropertiesOnModify & Data;
 
 export type Node<Data = Record<string, unknown>> = NodePropertiesOnRead & Data;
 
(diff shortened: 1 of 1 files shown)


tajakobsen (commented):
Very good! I think this is the kind of stuff that I could never have figured out. :)

---

There is a lot of new types here. And I was thinking that maybe this could be simplified by having the basic `Node` be the core of everything. And just change the shape of it using the `Omit` where necessary.

```typescript
export type Node<Data = Record<string, unknown>> = {
  _id: string;
  _indexConfig: NodeIndexConfig;
  _childOrder: string;
  _inheritsPermissions: boolean;
  _manualOrderValue?: number; 
  _name: string;
  _nodeType: string;
  _path: string;
  _permissions: AccessControlEntry[];
  _state: string;
  _ts: string;
  _versionKey: string;
} & Data;

// Does this `ModifiedNode` to be exported?
type ModifiedNode<NodeData> = Omit<Node<NodeData>, '_indexConfig'> & { 
  _indexConfig: NodeIndexConfigParams // are we sure that this will accept NodeIndexConfig too?
};

export interface ModifyNodeParams<NodeData = unknown> {
    key: string;
    editor: (node: Node<NodeData>) => ModifiedNode<NodeData>
}
```

The advantage is that we don't have to jump up and down in the code to figure out what belongs together, and the shape of types.

We just look at the `Omit` and understand the diff from regular `Node`.

ComLock:
Using Omit would have been easier, I was just trying to avoid using Omit, since we have avoided it in the past.
Sidenote: type-fest has a stricter Omit called Except.

In terms of jumping up and down, I dislike that myself, which is why I would have reordered the code.
It nags me that types are used before they are defined.
I typically try keeping things alphabetical and/or defined order.

The order seems rather random in these files?

ComLock:
I don't like my naming of ModifiedNode, but the reason it was exported, is that I used it in some code.
That code is however in a transitional face, so perhaps when I use @enonic-types for "everything" then I don't need that type anymore.

I love type, I hate transitioning types. High complexity.

edloidas (changes requested):
Awesome work.
The only thing I would ask is to rename `ModifyNode` to `ModifiedNode` (as you even named it in the comments).

---

I would not want to use `Omit` here, because it will be logical to export the `Node`, but the `Node` in this case won't be an actual type with a corresponding object in the lib.

As for the types order, it would probably be better to place non-common ones next to the functions that are using them, like `*Params`. Will change it later with some other improvements that we discussed in other PRs, like import/export sorting.

edloidas (approved):

Issue #8233, opened 2020-07-15
Layers: Multi project inheritance


rymsha:
It must not be possible to create cyclic sync dependencies between projects.

PR #9903, opened 2022-11-29
Layers: Multi project inheritance #8233

vbradnitski:


modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/ParentContentSynchronizer.java
```diff
@@ -267,6 +278,21 @@ else if ( actualTargetContext != null )
 
         } ).filter( Objects::nonNull ).collect( Collectors.toList() );
 
+        if ( result.isEmpty() )
+        {
```
rymsha, line 282: Do not use format in logging. Use slf4j `{}`

rymsha (approved):

Issue #9915, opened 2022-12-05
Content references are not extracted from all input occurrences


alansemenov:
Using the schema below, if HTMLArea in the first occurrence of the option-set has links to another content, they will be extracted as references upon save. However, this won't happen to links in HTML Area in the second and further occurrences.

```XML
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<content-type>
  <display-name>Html Area</display-name>
  <super-type>base:structured</super-type>
  <form>
    <option-set name="radioOptionSet">
      <label>Single selection</label>
      <help-text>Single selection radio option set</help-text>
      <occurrences minimum="1" maximum="0"/>
      <options minimum="1" maximum="1">
        <option name="option_1">
          <label>Option 1</label>
          <help-text>Help text for Option 1</help-text>
          <items>
            <input name="htmlarea_text2" type="HtmlArea">
              <label>HTML Area</label>
              <occurrences minimum="1" maximum="1"/>
            </input>
          </items>
        </option>
        <option name="option_2">
          <label>Option 2</label>
          <help-text>Help text for Option 2</help-text>
          <items>
            <input name="name" type="TextLine">
              <label>Name2</label>
              <occurrences minimum="1" maximum="1"/>
            </input>
          </items>
        </option>
      </options>
    </option-set>
  </form>
</content-type>
```
![image](https://user-images.githubusercontent.com/11612490/205618471-d70f7ad0-3800-4c1b-b626-1cbbb9393f73.png)

PR #9919, opened 2022-12-06
Content references are not extracted from all input occurrences #9915

vbradnitski:


rymsha (changes requested):

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/InputVisitorHelper.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import com.enonic.xp.data.Property;
+import com.enonic.xp.data.PropertySet;
+import com.enonic.xp.data.PropertyTree;
+import com.enonic.xp.form.Input;
+
+public class InputVisitorHelper
+{
```
rymsha, line 14: Add private constructor

rymsha (approved):

Issue #9777, opened 2022-10-11
repo.deleteBranch(master) fails, but has still deleted nodes


ComLock:
```
repo.deleteBranch({
  branchId: 'master',
  repoId: 'whatever'
});
```

Fails with the sorta weird error message: `java.lang.IllegalArgumentException: branches must contain master branch`

Which is fine, however upon inspection the root node is deleted.
I have not tested whether all nodes are deleted.

According to @rymsha, nothing should be deleted...

PR #9925, opened 2022-12-08
repo.deleteBranch(master) fails, but has still deleted nodes #9777

rymsha:
check is only done on lib layer.
and it does not consider project repositories as CMS repositories


rymsha:
- Move check to Java Service level
- Delete nodes more efficiently

anatol-sialitski (approved):

vbradnitski (approved):

Issue #9922, opened 2022-12-07
Content deletion isn't propagated to a grandchild layer


vbradnitski:
to be able to distinguish contents without inbound dependencies and children all removed contents should be analyzed in one batch.

PR #9929, opened 2022-12-09
Content deletion isn't propagated to a grandchild layer #9922

vbradnitski:


modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/DeletedEventSyncCommand.java
```diff
@@ -21,37 +33,92 @@ public static Builder create()
...
+
+    private boolean hasNoChildren( final ContentToSync contentToSync, final Set<ContentId> idsToRemove )
+    {
+        return !contentToSync.getTargetContent().hasChildren() ||
+            getAllChildren( contentToSync.getTargetContent().getId() ).stream().allMatch( idsToRemove::contains );
+    }
+
+    private boolean hasNoInboundDependencies( final ContentToSync contentToSync, final Set<ContentId> idsToRemove )
+    {
+        return getInboundDependencies( contentToSync.getTargetContent().getId() ).stream().allMatch( idsToRemove::contains );
```
rymsha, line 98: Optimize - move to query

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/ParentContentSynchronizer.java
```diff
@@ -246,14 +258,38 @@ else if ( actualTargetContext != null )
...
             }
+            else
+            {
+                return null;
+            }
+
+            if ( targetContent != null && ( targetContent.getOriginProject() == null ||
+                !actualSourceContext.getRepositoryId().equals( targetContent.getOriginProject().getRepoId() ) ) )
+            {
+                return null; // do not sync contents without origin parent or from other source
```
rymsha, line 269: This is a tricky part and must be covered with test

modules/runtime/src/bin/server.sh
```diff
@@ -37,7 +37,7 @@ locateJava() {
 }
 
 setupDefaults() {
-    DEFAULT_JAVA_OPTS="-XX:-OmitStackTraceInFastThrow -XX:+AlwaysPreTouch"
+    DEFAULT_JAVA_OPTS="-XX:-OmitStackTraceInFastThrow -XX:+AlwaysPreTouch -XX:-MaxFDLimit"
```
rymsha, line 40: Let's have a separate issue for this

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/DeletedEventSyncCommand.java
vbradnitski: 'idsToRemove' conversation could be really big

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/ParentContentSynchronizer.java
vbradnitski: removed it, it's unreal behavior outside of multi-inheritance POC for now

rymsha (approved):

anatol-sialitski (approved):

Issue #9779, opened 2022-10-11
TS Move common types to core package


edloidas:
Move the following types to core package and reexport them from the libs:
* [x] DSL definitions (`QueryDsl`, `Aggregation`) are shared between `node` and `content` libs
* [x] Content types (`InputType`, `FormItemSet`, etc) are shared between `content` and `schema` libs.

PR #9940, opened 2022-12-20
TS Move common types to core package #9779

edloidas:
Moved DSL queries, Aggregations, Filters, and Form items to core package.

All of the types are been re-exported from the old libs.

All `FormItem*` have a certain type right now, e.g. `FormItemInput` has `formItemType: 'Input'`. This is how it was implemented in `schema` lib. In `content` lib we had a more loose definition `formItemType: string | FormItemType;`. I'm not sure it was correct (more to that, it should be using LiteralUnion in that case).

### DIFF
--- a/modules/lib/core/index.d.ts
+++ b/modules/lib/core/index.d.ts
@@ -266,3 +266,205 @@ export interface GeoDistanceSortDsl
 }
 
 export type SortDsl = FieldSortDsl | GeoDistanceSortDsl;
+
+//
+// START AGGREGATIONS, FILTERS
+//
+export interface Bucket {
+    [subAggregationName: string]: AggregationsResult | string | number | undefined;
+
+    key: string;
+    docCount: number;
+}
+
+export interface NumericBucket
+    extends Bucket {
+    from?: number;
+    to?: number;
+}
+
+export interface DateBucket
+    extends Bucket {
+    from?: string;
+    to?: string;
+}
+
+export interface BucketsAggregationResult {
+    buckets: (DateBucket | NumericBucket)[];
+}
+
+export interface StatsAggregationResult {
+    count: number;
+    min: number;
+    max: number;
+    avg: number;
+    sum: number;
+}
+
+export interface SingleValueMetricAggregationResult {
+    value: number;
+}
+
+export type AggregationsResult = BucketsAggregationResult | StatsAggregationResult | SingleValueMetricAggregationResult;
+
+export type Aggregation =
+    | TermsAggregation
+    | HistogramAggregation
+    | DateHistogramAggregation
+    | NumericRangeAggregation
+    | DateRangeAggregation
+    | StatsAggregation
+    | GeoDistanceAggregation
+    | MinAggregation
+    | MaxAggregation
+    | ValueCountAggregation;
+
+export interface TermsAggregation {
+    terms: {
+        field: string;
+        order?: string;
+        size?: number;
+        minDocCount?: number;
+    };
+    aggregations?: Record<string, Aggregation>;
+}
+
+export interface HistogramAggregation {
+    histogram: {
+        field: string;
+        order?: string;
+        interval?: number;
+        extendedBoundMin?: number;
+        extendedBoundMax?: number;
+        minDocCount?: number;
+    };
+    aggregations?: Record<string, Aggregation>;
+}
+
+export interface DateHistogramAggregation {
+    dateHistogram: {
+        field: string;
+        interval?: string;
+        minDocCount?: number;
+        format: string;
+    };
+    aggregations?: Record<string, Aggregation>;
+}
+
+export interface NumericRange {
+    from?: number;
+    to?: number;
+    key?: string;
+}
+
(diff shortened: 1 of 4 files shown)


anatol-sialitski (approved):

rymsha (approved):

tajakobsen (approved):
Looks good. Just one thing. Did you consider creating multiple files in core to modularize the types a bit. _index.d.ts_ is getting a bit large. 

You can always re-export types from _index.d.ts_ too, but they live in a file named for their domain.

edloidas:
@tajakobsen As I remember, all types were in their files (under `shared`) [initially](https://github.com/enonic/xp/pull/9739/commits/00029cf7fa60ecbf7fcd413df0827eda08ba85fc#diff-c97b8adf4134c9ad823819efe70948cb4e8ff281b1289ca03fe6869fbf6cd8aa). But then they were merged into one and renamed to `core` after discussion. I can not recall the details of that meeting now, so we probably need to discuss this question again.

modules/lib/core/index.d.ts
```diff
@@ -118,3 +118,472 @@ export interface Resource {
...
+    | NumericRangeAggregation
+    | DateRangeAggregation
+    | StatsAggregation
+    | GeoDistanceAggregation
+    | MinAggregation
+    | MaxAggregation
+    | ValueCountAggregation;
+
+export interface TermsAggregation {
+    terms: {
```
ComLock, line 323: It would be nice if this was "split" up even more

```
export interface TermsAggregationParams {
        field: string;
        order?: string;
        size?: number;
        minDocCount?: number;
}
    
export interface TermsAggregation {
    terms: TermsAggregationParams;
    aggregations?: Record<string, Aggregation>;
}
```

Because that's certainly something I've used. The workaround for me is using TermsAggregation['terms'], but that's not very pretty

modules/lib/lib-content/src/main/resources/lib/xp/content.ts
```diff
@@ -48,6 +60,50 @@ export type {
     Content,
     Component,
     Region,
+    QueryDsl,
```
ComLock, line 63: This is a generic comment for all Enonic code. Why aren't we sorting alphabetically? It becomes hard to see what's there or not when it's "random" order.

modules/lib/lib-schema/src/main/resources/lib/xp/schema.ts
```diff
@@ -13,17 +13,20 @@ declare global {
...
-} from '@enonic-types/core';
+import type {ByteSource, FormItem, UserKey} from '@enonic-types/core';
 
 export type {
     ByteSource,
     GroupKey,
     PrincipalKey,
     RoleKey,
     UserKey,
+    FormItemSet,
```
ComLock, line 24: Just another example of non alphabetical order

ComLock (changes requested):

modules/lib/core/index.d.ts
tajakobsen: I actually find `TermsAggregation['terms']` to be preferable to extracting it out. 

There are so many interfaces already, and this helps keeping the "sub type" in the context it is used in.

@ComLock In the end, I think it might be more about what you eyes are used to. Personally I'm happy when I see types like `TermsAggregation['terms']`, because it also tells me about where this value is intended to end up in a larger context.

modules/lib/lib-content/src/main/resources/lib/xp/content.ts
tajakobsen: I recommend looking into having the tooling (like eslint) doing the job if this is something you are going for. 

_(Disclamer: I have not checked if eslint actually can do this.)_

modules/lib/core/index.d.ts
edloidas: Yeah, on one side it's cleaner to use `TermsAggregationParams`, on the other side the context is less clear and the imports/exports become dirtier. 
What I usually do in my projects in similar cases is `type TermsAggregationParams = TermsAggregation['terms']`. But it is only okay if you need to do this in 1-2 files in a whole project. 
Are developers (and you guys) are using `TermsAggregation['terms']` often? If it's a pretty rare case, then I suggest we keep it as is and just "rename" it locally in your project.

modules/lib/lib-content/src/main/resources/lib/xp/content.ts
edloidas: The formatting is really I hard thing to discuss. The only working solution that I can see right now is Prettier.

But let me tell you about the whole picture first. It so happened that the core team uses the IntelliJ IDEA and the shared config where various code styles are defined. In the past we tried different approaches, even git precommit hooks + lint-staged + prettier. These practices have not caught on though. It's kind off hard to keep the current styles, make the formatting process simple, not demanding and non-blocking.

I think that we'll discuss the styling improvements this month and, probably expand our eslint config with prettier in the future, but definitely not right now and not in this PR.

Issue #9686, opened 2022-09-12
Allow to customize base URL for URL rewriting in portal.url functions


rymsha:
Currently base url is read from the HTTP request so the resulting domain and path can only match the one from the original request.
It is inconvenient in cases where the result of rewriting should be different from the request (like building URLs during on the request from GraphQL)

It should be possible to specify (via context) which base URL the rewrites should happen.

rymsha:
#10193 solves it for media

PR #9946, opened 2023-01-02
Allow to customize base URL for URL rewriting in portal.url functions #9686

anatol-sialitski:


modules/web/web-api/src/main/java/com/enonic/xp/web/servlet/ServletRequestUrlHelper.java
```diff
@@ -134,7 +138,7 @@ public static String getServerUrl( final HttpServletRequest req )
         str.append( scheme );
 
         //Appends the host
-        str.append( "://" ).append( req.getServerName() );
+        str.append( "://" ).append( Objects.requireNonNullElse( getCustomHostFromContext(), req.getServerName() ) );
```
rymsha, line 141: Configuring host is not sufficient. It should be baseUrl.
Example: `https://example.com/some/path`

modules/web/web-api/src/main/java/com/enonic/xp/web/servlet/ServletRequestUrlHelper.java
```diff
@@ -298,4 +302,9 @@ private static void appendRfc8187Encoded( final StringBuilder builder, final Str
             }
         }
     }
+
+    private static String getCustomHostFromContext()
```
rymsha, line 306: This attribute does not need to have a helper method. Remove

rymsha (changes requested):

vbradnitski (approved):

rymsha (changes requested):
It doesn't solve the problem. Back to whiteboard.

Issue #9793, opened 2022-10-13
Improve TypeScript libs code


edloidas:
Functions similar to `isString` can be improved [using type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates):
```ts
// from
function isString(value: unknown): boolean {
    return typeof value === 'string' || value instanceof String;
}
// to
const isString(value: unknown): value is string => typeof value === 'string' || value instanceof String;
```
When used in `if`, it will be unnecessary to use explicit type casting, e.g. `value as string`.

Same goes for ` isObject` and other functions.

_These changes will lead to more robust code._

PR #9966, opened 2023-01-19
Improve TypeScript libs code #9793

edloidas:
Part of the functions were already improved, so `isObject` was the only one that left.


anatol-sialitski (approved):

tajakobsen (approved):
Looks good.

rymsha (approved):

Issue #9761, opened 2022-10-04
TS Add return type of resolve function 


edloidas:
* [x] Add `ResourceKey` to `core`, since it can be shared between libs
  ```ts
  export interface ResourceKey {
      applicationKey: string;
      path: string;
      uri: string;
      root: boolean;
      name: string;
      extension: string;
  }
  ```
* [x] Make `resolve` function return `ResourceKey` in `global`
  ```ts
  declare const resolve: (path: string) => ResourceKey;
  ```
* [x] Make `global` to depend on `core`

PR #9969, opened 2023-01-19
TS Add return type of resolve function #9761

edloidas:
Moved the `ResourceKey` to core package.
Made `resolve` function definition in `global.d.ts` to return the same object as `ResourceKey`. I don't want to make `global.d.ts` depend on `core` and be importable, neither define `ResourceKey` in global scope. I guess `ResourceKey` is not something that will change and we can keep compromise here using the identical "types" in different contexts.

### DIFF
--- a/modules/lib/core/index.d.ts
+++ b/modules/lib/core/index.d.ts
@@ -109,6 +109,9 @@ type Brand<
 
 export type ByteSource = Brand<object, 'ByteSource'>;
 
+//
+// RESOURCES
+//
 export interface Resource {
     getSize(): number;
 
@@ -117,6 +120,15 @@ export interface Resource {
     getStream(): ByteSource;
 
     exists(): boolean;
+}
+
+export interface ResourceKey {
+    getApplicationKey(): string;
+    getPath(): string;
+    getUri(): string;
+    isRoot(): boolean;
+    getName(): string;
+    getExtension(): string;
 }
 
 //
--- a/modules/lib/global.d.ts
+++ b/modules/lib/global.d.ts
@@ -193,11 +193,18 @@ declare const require: XpRequire;
  * var path = resolve('./other.html');
  *
  * @example
- * // Reslove absolute
+ * // Resolve absolute
  * var path = resolve('/path/to/other.html');
  *
  * @param {string} path Path to resolve.
  * @returns {*} Reference to an object.
  * @global
  */
-declare const resolve: (path: string) => object;
+declare const resolve: (path: string) => {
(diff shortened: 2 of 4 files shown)


rymsha (changes requested):
lib-io `getResource` accepts ResourceKey | string
It should be updated in this pool

modules/lib/core/index.d.ts
```diff
@@ -587,3 +587,15 @@ export interface FormItemInlineMixin {
...
+//
+// RESOURCES
+//
+export interface ResourceKey {
+    applicationKey: string;
+    path: string;
+    uri: string;
+    root: boolean;
+    name: string;
+    extension: string;
```
rymsha, line 600: This is a Java object. It does not have fields, it has getters.

rymsha (changes requested):

modules/lib/global.d.ts
```diff
@@ -193,11 +193,18 @@ declare const require: XpRequire;
...
  */
-declare const resolve: (path: string) => object;
+declare const resolve: (path: string) => {
+    getApplicationKey(): string;
+    getPath(): string;
+    getUri(): string;
+    isRoot(): boolean;
+    getName(): string;
+    getExtension(): string;
+};
```
tajakobsen, line 210: Is there a reason why `ResourceKey` isn't used here? Like this:

```suggestion
declare const resolve: (path: string) => ResourceKey;
```

ComLock (approved):
Seems good.

Just a thought:
If we allowed a single name "Enonic" to be allocated in the global scope.
Then we could define ResourceKey and more there.

```
declare namespace Enonic {
    interface ResourceKey {
        applicationKey: string;
        path: string;
        uri: string;
        root: boolean;
        name: string;
        extension: string;
    }
}
```

Anywhere one wanted to references the type one would not need to import the type just simple state Enonic.ResourceKey

tajakobsen (changes requested):
Looks good. I just have a small question here.

modules/lib/global.d.ts
ComLock: Because ResourceKey isn't defined there. And as @edloidas said in his initial comment, he doesn't want to define it, nor import it from core.

tajakobsen: Sorry. I missed that. :) Maybe add a comment both in *core* and *global* to remind to update the other place, if the shape is changed.

tajakobsen (approved):

rymsha (approved):

Issue #9970, opened 2023-01-19
Dsl Term query doesn't work with uppercase


vbradnitski:
```
{
  "term": {
    "field": "displayName",
    "value": "Header"
  }
}
``` 

does not find anything

rymsha:
This happens because XP calls `.tolowerCase()` on every value set in search index. 
`term` DSL uses `term` ES query - it does not analyze the string and can never find any string with uppercase letters.
- we may reconsider how we index values in XP8, but for now we have to lower-case values before searching them

Additionally localDate and localTime need special formatting before search for almost the same reason - XP applies special format for them when storing in search index.

PR #9971, opened 2023-01-19
Add match dsl query builder #9970

vbradnitski:


rymsha (changes requested):
Probably we can avoid it by implementing term the same way it was done in old query filters

Issue #9556, opened 2022-07-06
TypeScript Make objects with object properties generic


edloidas:
In global definitions we have:
```ts
// typescript/index.ts
declare const __: {
    toNativeObject: <T = unknown>(beanResult: T) => T;
};
```

__Before:__
```ts
// audit.ts
export interface AuditLogParams {
    data?: Record<string, unknown>;
}

export interface AuditLog {
    data: Record<string, unknown>;
}

interface CreateAuditLogHandler {
    execute(): AuditLog;
}

export function log(params: AuditLogParams): AuditLog {
    return __.toNativeObject(bean.execute());
}
```
Test
```ts
const result = log({data: {foo: 'bar'}});
result.data.foo; // OK
result.data.x;   // OK, but must fail
```

__After:__
```ts
export interface AuditLogParams<DATA extends object = object> {
    data?: DATA;
}

export interface AuditLog<DATA extends object = object> {
    data: DATA;
}

interface CreateAuditLogHandler<DATA extends object = object> {
    execute(): AuditLog<DATA>;
}

export function log<DATA extends object>(params: AuditLogParams<DATA>): AuditLog<DATA> {
    const bean = __.newBean<CreateAuditLogHandler<DATA>>('com.enonic.xp.lib.audit.CreateAuditLogHandler');
    // ...
    return __.toNativeObject(bean.execute());
}
```
Test
```ts
const result = log({data: {foo: 'bar'}});
result.data.foo; // OK
result.data.x;   // Error, since property `x` does not exist in `data`
```

PR #9972, opened 2023-01-19
TypeScript Make objects with object properties generic #9556

edloidas:


modules/lib/lib-auditlog/src/main/resources/lib/xp/auditlog.ts
```diff
@@ -23,26 +23,26 @@ function checkRequired<T extends object>(obj: T, name: keyof T): void {
     }
 }
 
-export interface AuditLogParams {
+export interface AuditLogParams<DATA extends Record<string, unknown> = Record<string, unknown>> {
```
tajakobsen, line 26: The convention used other places in the code is that type parameters are capitalized (not upper case).

```suggestion
export interface AuditLogParams<Data extends Record<string, unknown> = Record<string, unknown>> {
```

tajakobsen: Where is it smart to have default type parameters (` = Record<string, unknown>`)? I'm thinking maybe on the functions, and not in the interfaces. Specially when the objects are only created by this library.

(This goes for the other cases below too.)

What do you think?

modules/lib/lib-project/src/main/resources/lib/xp/project.ts
```diff
@@ -31,31 +31,31 @@ export interface ProjectReadAccess {
     public: boolean;
 }
 
-export interface CreateProjectParams {
+export interface CreateProjectParams<CONFIG extends Record<string, unknown> = Record<string, unknown>> {
```
tajakobsen, line 34: The same comments as above. Don't uppercase type parameters, and think through where we want default type parameters (I think we maybe only want them on functions).

tajakobsen (changes requested):
Good job. Think a bit about where it's smart to have *default type parameters*. I think it's maybe only on functions if the users are not creating instances of this object themselves.

ComLock (approved):
One problem with the current positional generics, is keeping backwards compatibility when one wants to add more generics later on.

Hopefully named generics will be added soon (give it a thumbs up?):
https://github.com/microsoft/TypeScript/issues/38913

edloidas:
@tajakobsen Agree on all points. Had an uppercased template names in other project I worked on before this commit, so unintentionally used this style here too.
As for the default type params, we have an inconsistency across the project, so I looked through all the files and updated them: Params won't have defaults, but functions will. The result types, like AuditLog, Content, etc will also have a default type params in most of the cases.

tajakobsen (approved):
Looks good.

Issue #9984, opened 2023-01-23
GeoDistanceAggregation has multiple problems


ComLock:
https://github.com/enonic/xp/blob/master/modules/lib/core/index.d.ts#L402-L412

1. unit should be an enum
https://developer.enonic.com/docs/xp/stable/storage/aggregations#geodistance

The meassurement unit to use for the ranges. Legal values are either the full name or the abbreviation of the following: km (kilometers), m (meters), cm (centimeters), mm (millimeters), mi (miles), yd (yards), ft (feet) or nmi (nauticalmiles).

2. unit should be optional?
It works without it, the doc doesn't say what the default fallback is

3. origin should be required (elastic will fail if it's missing)

4. ranges should be required (elastic will fail if it's missing)

PR #9991, opened 2023-01-24
GeoDistanceAggregation has multiple problems #9984

ComLock:


rymsha (changes requested):
there is no need to explain what happens if required parameter is not specified

rymsha (approved):

Issue #9573, opened 2022-07-18
XmlObjectParser skips validation for sources without namespaces


vbradnitski:
Parser validates only xml with set namespaces and skips it without any notification otherwise.

rymsha:
#11182 obsoletes this

PR #10010, opened 2023-01-30
XmlObjectParser skips validation for sources without namespaces #9573

vbradnitski:


modules/core/core-api/src/main/java/com/enonic/xp/xml/parser/XmlObjectParser.java
```diff
@@ -29,10 +33,17 @@
 {
     private static final SchemaValidator VALIDATOR = new SchemaValidator();
 
+    private final String namespace;
+
     private String systemId;
 
     private CharSource source;
 
+    public XmlObjectParser( final String namespace )
```
rymsha, line 42: add default ctor that, if used, does not force any namespace. `namespace = null`

modules/core/core-api/src/main/java/com/enonic/xp/xml/parser/XmlObjectParser.java
```diff
@@ -114,13 +125,61 @@ private P doParse( final Document source )
...
+        final Document processedDoc = processDocument( doc );
+
+        final DOMSource source = new DOMSource( processedDoc );
         source.setSystemId( this.systemId );
 
         final DOMResult result = VALIDATOR.validate( source );
         return (Document) result.getNode();
     }
 
+    private Document processDocument( final Document doc )
```
rymsha, line 137: maybe better call it `forceNamespace`

rymsha (changes requested):

rymsha (changes requested):
I think we need to cancel this change
- apps may suddenly stop working due to extra validation
- we may want to focus on json schema for descriptors

Issue #9952, opened 2023-01-04
Compare versions modal dialog - not added data is present in the dialog


sgauruseu:

alansemenov:
This looks like an XP core issue to me. Basically, when a new content item is created it's missing x-data fields, then Content Studio adds them the first time the item is modified and saved.

After creation:
![image](https://user-images.githubusercontent.com/11612490/211298721-a874f94c-a170-48e4-8245-cdefe0a21ab4.png)

After first update:
![image](https://user-images.githubusercontent.com/11612490/211298837-87f10681-253e-4226-be03-308e301a55bf.png)

PR #10031, opened 2023-02-07
X-data is not added to content when it's created #9952

anatol-sialitski:


rymsha (changes requested):
We need a separate method in lib-schema to get default x-data
This default x-data should be provided into content.create if needed

vbradnitski (approved):

Issue #9980, opened 2023-01-23
Race condition in Branch cache


rymsha:
Race condition in Branch cache leads to stale node id returned when getByPath is used.

rymsha:
Even obvious fix won't work, due to bug in Guava
https://softwaremill.com/race-condition-cache-guava-caffeine/

PR #10046, opened 2023-02-23
Race condition in Branch cache #9980

rymsha:


modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/storage/NodeStorageServiceImpl.java
```diff
@@ -300,7 +285,7 @@ public NodeBranchEntry getBranchNodeVersion( final NodeId nodeId, final Internal
     @Override
     public NodeBranchEntries getBranchNodeVersions( final NodeIds nodeIds, final InternalContext context )
     {
```
anatol-sialitski, line 287: Check empty `nodeIds` or not

vbradnitski (approved):

Issue #10067, opened 2023-03-06
StatusServlet fails to render an error


vbradnitski:
It tries to open a stream that is already open.

PR #10068, opened 2023-03-06
StatusServlet fails to render an error #10067

vbradnitski:


modules/server/server-status/src/main/java/com/enonic/xp/server/impl/status/StatusServlet.java
```diff
@@ -107,18 +101,17 @@ private void serialize( final HttpServletRequest req, final HttpServletResponse
...
     {
+        res.reset();
         res.setStatus( status );
         res.setContentType( MediaType.JSON_UTF_8.toString() );
 
-        final PrintWriter out = res.getWriter();
-        out.println( json.toString() );
-        out.close();
+        final OutputStream out = res.getOutputStream();
+        out.write( json.toString().getBytes( StandardCharsets.UTF_8 ) );
```
rymsha, line 114: It is possible to save one copy if you open a response Writer and write a resulting string directly. 
Note that Jetty uses encoding ContentType - so we will be fine.
To get really mad about performance we can use JsonWriter but it would complicate the code

rymsha (approved):

anatol-sialitski (approved):

Issue #10086, opened 2023-03-15
Aggregations type cast error


alansemenov:
<img width="1215" alt="image" src="https://user-images.githubusercontent.com/11612490/225280247-73adf244-f5e2-4461-8309-30bb46339425.png">

PR #10087, opened 2023-03-15
Aggregations type cast error #10086

edloidas:
Moved `GeoDistanceAggregation` type to separate check from `BucketsAggregationsUnion`, since it's the only bucket aggregation, that does not contain sub-aggregations, wich resulted in type errors in DTS files (not sure why TS does not show errors here).


tajakobsen (approved):
I think this will work. Good job!

Issue #10084, opened 2023-03-15
Incorrect type of DuplicateParams.dataProcessor in lib-node types


alansemenov:
https://github.com/enonic/xp/blob/master/modules/lib/lib-node/src/main/resources/lib/xp/node.ts#L446

`ScriptValue` (from `global.d.ts`) is used as a type for `dataProcessor` inside exported interface `DuplicateParams`. This would break the build unless `global.d.ts` is always included together with `node.d.ts`.

PR #10088, opened 2023-03-15
Incorrect type of DuplicateParams.dataProcessor in lib-node types #10084

anatol-sialitski:


modules/lib/lib-node/src/main/resources/lib/xp/node.ts
```diff
@@ -438,12 +438,12 @@ interface FindChildrenHandlerParams {
     setRecursive(recursive: boolean): void;
 }
 
-export interface DuplicateParams {
+export interface DuplicateParams<NodeData = Record<string, unknown>> {
```
tajakobsen, line 441: If this should be consistent with the way it's solved in `CreateNodeParams` and `ModifyNodeParams` the default value should just be `unknown`.

```suggestion
export interface DuplicateParams<NodeData = unknown> {
```

tajakobsen (changes requested):
Have a look at the default type parameter for `DuplicateParams`, it's not consistent with what is chosen in similar interfaces.

tajakobsen:
Btw, I'm not getting any errors from these types anymore.

modules/lib/lib-node/src/main/resources/lib/xp/node.ts
ComLock: Since we don't actually support anything but objects that uses string as the key, would this be better?
```suggestion
export interface DuplicateParams<NodeData extends Record<string, unknown> = Record<string, unknown>> {
```

ComLock: And to the same for `CreateNodeParams` and `ModifyNodeParams`?

tajakobsen: I don't think using `extends Record<string, unknown>` will work. I tried it out, and it rejects my (legal) `interface` with the  shape that I want.

tajakobsen: Regarding my initial suggestion, I don't feel strongly about this. It's was more a comment to try to maintain consistency of code across functions.

Either way is good. 

Developers will most likely not interact with this shape independently, but only trough using the `duplicate()`-function which is the one that actually takes the type parameter, and has a default type parameter.

ComLock (approved):
I approve. My comments are not important :)

tajakobsen (approved):
I fine with either way regarding my suggestion. So I approve too.

Issue #10104, opened 2023-03-27
MultipartItem does not report known size


rymsha:
MultipartItem extends ByteSource but returns no known size - that, for instance, does not allow to set correct content-length when if multipart-item is reposted via http-client.

PR #10105, opened 2023-03-27
MultipartItem does not report known size #10104

rymsha:


modules/lib/lib-app/src/main/java/com/enonic/xp/lib/app/mapper/IconByteSource.java
```diff
@@ -41,7 +42,13 @@ public boolean isEmpty()
...
     public long size()
     {
-        return isEmpty() ? 0 : icon.getSize();
+        return icon == null ? 0L : icon.getSize();
+    }
+
+    @Override
+    public Optional<Long> sizeIfKnown()
+    {
+        return Optional.of( icon == null ? 0L : icon.getSize() );
```
anatol-sialitski, line 51:
```suggestion
        return Optional.of( size() );
```

anatol-sialitski (approved):

vbradnitski (approved):

Issue #7667, opened 2019-11-22
portal.getComponent returns null for page controllers


sigdestad:
As of xp7, a page is a component, conceptually like any other.
Make sure this function returns the page component definition, juse like it does for parts and layouts.

GlennRicaud:
Page is only a component for the serialization. Everywhere inside XP and at application level, the structure is exactly the same.

I am closing this issue as this is a much bigger issue

sigdestad:
Glenn, I don't know what this looks like on Java level, but why does returning the basic component info require refactoring?

GlennRicaud:
Because Page is not a component.
It is serialized as one but it has never been a component I think

sigdestad:
But we are takling about returning a serialized object here

ComLock:
React4xp uses the page component in order to make some smart choices for rendering.
Documentation and examples uses getComponent in the page controller.

There are also documentation and examples from other apps, that uses getComponent in page.

https://github.com/search?q=org%3Aenonic+getComponent%28%29+path%3A%2Fsrc%2Fmain%2Fresources%2Fsite%2Fpages&type=Code

https://github.com/search?q=org%3Aenonic-customer-projects+getComponent%28%29+path%3A%2Fsrc%2Fmain%2Fresources%2Fsite%2Fpages

ComLock:
@sigdestad want's this fixed in Enonic XP 8

ComLock:
The Component type should include page, text, fragment (and maybe macro?)
https://github.com/enonic/xp/blob/master/modules/lib/core/index.d.ts#L59

rymsha:
The result of a method call from page controller should look like
```
var expected = {
  "path": "/",
  "type": "page",
  "descriptor": "myapplication:myPage",
  "config": {
    "a": "1"
  },
  "regions": {
    "bottom": {
      "components": [
        {
          "path": "/main/0",
          "type": "part",
          "descriptor": "myapplication:mypart",
          "config": {
            "a": "1"
          }
        }
      ],
      "name": "bottom"
    }
  }
};
```

ComLock:
There is a full type declaration for PageComponent here: https://github.com/enonic/xp/blob/master/modules/lib/core/index.d.ts

rymsha:
Some clarifications:

1. Documentation states:    [getComponent](https://developer.enonic.com/docs/xp/stable/api/lib-portal#getcomponent)    
[...]

PR #10117, opened 2023-04-14
Fix 10115 Component type fragment is missing from the types

ComLock:


modules/lib/core/index.d.ts
```diff
@@ -52,14 +52,54 @@ export interface PublishInfo {
...
+}
+
+export type Component<Config extends object = object, Regions extends Record<string, Region> = Record<string, Region>> =
+    | FragmentComponent
+    | LayoutComponent<Config, Regions>
+    | PartComponent<Config>
+    | PageComponent
+    | TextComponent;
+
 export interface Region<Config extends object = object> {
```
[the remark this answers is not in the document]
ComLock: If definetly like the idea of passing Component rather than Config into Region, after all Fragment and Text components doesn't even have Config... It's a "breaking" change though.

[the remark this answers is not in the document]
ComLock: I think it's always better to pass in the full shapes of properties on the "root" level of an interface, or one will run into problems later on, like now, when it's a breaking change.

modules/lib/core/index.d.ts
```diff
@@ -52,14 +52,54 @@ export interface PublishInfo {
...
 }
 
-export interface Component<Config extends object = object, Regions extends Record<string, Region> = Record<string, Region>> {
-    config: Config;
-    descriptor: string;
+export interface FragmentComponent {
+    fragment: string;
     path: string;
-    type: 'page' | 'layout' | 'part';
+    type: 'fragment'
```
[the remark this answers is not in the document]
ComLock: What do you mean "on top"? Ordered first in the list of properties?

modules/lib/core/index.d.ts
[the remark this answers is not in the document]
ComLock: For example

```
export interface Region<Components extends Component[] = object[]> {
    name: string;
    components: Components;
}
```

Or just

```
export interface Region<Components extends object[] = object[]> {
    name: string;
    components: Components;
}
```

Then maybe one could pass in even a tuple type. Say the fist component should always be text and the second a part. But that not really possible to restrict in Content Studio (yet?)

modules/lib/core/index.d.ts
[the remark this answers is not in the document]
tajakobsen: Yes, on top

modules/lib/core/index.d.ts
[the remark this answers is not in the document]
tajakobsen: Yea, it's not good that it's a breaking change. But this will probably not be an issue that happens more then a handful of times in a project.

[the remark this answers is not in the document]
tajakobsen: ...unless you change the signaure of  `getComponent` to:

```typescript
export declare function getComponent<
  T extends PartComponent | LayoutComponent = PartComponent | LayoutComponent
>(): T | null;
```

Which would create breakage in many parts, but would be a very nice Interface to program against.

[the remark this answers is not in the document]
ComLock: Sidenote: I believe getComponent used to work in page controller aswell (pre xp7) https://github.com/enonic/xp/issues/7667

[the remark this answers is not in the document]
ComLock: I would change getComponent too :)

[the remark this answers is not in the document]
ComLock: Which also requires GetCurrentComponentHandler to be changed
```
interface GetCurrentComponentHandler<
    T extends PartComponent | LayoutComponent = PartComponent | LayoutComponent
> {
    execute(): T | null;
}
```

tajakobsen:
@alansemenov I think we need to resolve the issues from my previous review before I can review again. 

More specifically, the types for traversing a component tree is not very good. I think `Component.descriptor` should work the same way as `Content.type`. Then we can iterate over the components and pick out the one with the correct `config` based on `descriptor`.

modules/lib/core/index.d.ts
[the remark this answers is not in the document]
ComLock: Or like this
Screenshot 2023-05-02 at 14 13 16

modules/lib/core/index.d.ts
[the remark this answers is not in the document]
ComLock: Conversation continues in other conversation

modules/lib/core/index.d.ts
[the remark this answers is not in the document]
ComLock: Alternative not using Union:
Screenshot 2023-05-02 at 15 47 46

tajakobsen (changes requested):
I think you have done a very good job making it possible to model the shape of the data, and making it possible to traverse the tree.

But, I have a feeling that the datastructures can become pretty complex. So I have a suggestion here, where we do like with `XpXData`, and instead register Pages, Parts and Layouts in global interfaces. This way we don't need to use type parameters to build up data structures at all, since the components only need to be registered globally once, we can find them everywhere in the tree they might show up.

ComLock:
How is using globals any less complex to use?

modules/lib/core/index.d.ts
```diff
@@ -52,19 +52,105 @@ export interface PublishInfo {
...
+> {
+    type: 'layout'
+    descriptor: Descriptor
+    config: Config
+    path?: string // Missing in fragmentPreview https://github.com/enonic/xp/issues/10116
+    regions: Regions;
+}
+
+export interface PartComponent<
+    Descriptor extends string = string,
```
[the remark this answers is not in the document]
ComLock: The thinking was that descriptor was the discriminating "id". So for a layout it could theoretically limit the shape of regions. But in the example above there is no "link" between descriptor and regions.

modules/lib/core/index.d.ts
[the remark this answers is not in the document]
ComLock: One "drawback" is that the XpPartMap only contains the PartConfig, meaning if parts are extended in the future, it's becomes a breaking type change.

If the XpPartMap contained the full Component, this would not be the case...

modules/lib/core/index.d.ts
```diff
@@ -52,23 +63,71 @@ export interface PublishInfo {
...
+    fragment?: string;
+    path?: string;
+    regions?: Regions;
+    type: 'fragment' | 'layout' | 'page' | 'part' | 'text';
+    text?: string
+}
+
+export interface Region<Components extends Component[] = Component[]> {
     name: string;
-    components: Component<Config>[];
```
ComLock, line 65: @tajakobsen I think this is the only "major" change. A region can have multiple different components, so it makes no sense to send in only a single config. To describe the full shape, one needs to pass in all the potential components.

I'm using the very generic Component here, I guess one could use 

```typescript
(Fragment | Layout | Part | TextComponent)[]
```

Instead?

ComLock: I still like the idea of PageRegion and LayoutRegion, since they are not the same.

modules/lib/core/index.d.ts
```diff
@@ -88,7 +147,7 @@ export interface Content<
     originProject?: string;
     childOrder?: string;
     _sort?: object[];
-    page: Type extends 'portal:fragment' ? undefined : _Component;
+    page: Type extends 'portal:fragment' ? undefined : Page;
```
tajakobsen, line 150: We are missing something here...

This would not give a union of all Page-types automatically.

We would need something like:

```typescript
type EveryPageDescriptor = keyof XpPartMap
type EveryPage = EveryPageDescriptor extends any // this lets us iterate over every member of the union
 ? Part<EveryPageDescriptor>
 : never;
```

Here comes this line:
```suggestion
    page: Type extends 'portal:fragment' ? undefined : EveryPage;
```

modules/lib/core/index.d.ts
```diff
@@ -52,23 +63,71 @@ export interface PublishInfo {
...
+export type PageDescriptor = keyof XpPageMap;
+export interface Page<
+    Descriptor extends PageDescriptor = PageDescriptor,
+> {
+    type: 'page'
+    descriptor: Descriptor
+    config: XpPageMap[Descriptor]
+    path: '/'
+    regions: Record<string, Region<(Fragment | Layout | Part | TextComponent)[]>>;
+}
```
tajakobsen, line 102: I think I would prefer to make these types completely independent of `XpPageMap` etc, and Instead create a new type which iterates over the `XpPageMap` and uses this one to to build up that union. (See `EveryPage` below.)

```suggestion
export interface Page<
    Descriptor extends string,
    Config
> {
    type: 'page'
    descriptor: Descriptor
    config: Config
    path: '/'
    regions: Record<string, Region<(Fragment | Layout | Part | TextComponent)[]>>;
}
```

modules/lib/core/index.d.ts
```diff
@@ -52,23 +63,71 @@ export interface PublishInfo {
...
+
+export interface TextComponent {
+    type: 'text'
+    path: string
+    text: string
+}
+
+export interface Component<
```
tajakobsen, line 110: I don't think `Component` will be very useful anymore. Maybe it could be a union of the possible types it can be, and only used as a constraint on type parameters. 

```typescript
type Component =
  | Fragment
  | TextComponent 
  | Page<string, unknown> 
  | Part<string, unknown>
  | Layout<string, unknown>; 
```

tajakobsen (changes requested):
Hi, this one is pretty tricky. Maybe we can have a look together on it tomorrow?

modules/lib/core/index.d.ts
ComLock: Also one looses the ability to specify a specific page, so perhaps removing the third generic param is a mistake

modules/lib/core/index.d.ts
```diff
@@ -1,4 +1,15 @@
+export type ComponentDescriptor = `${string}:${string}`;
+
 declare global {
+    interface XpLayoutMap {
+        [layoutDescriptor: ComponentDescriptor]: object;
```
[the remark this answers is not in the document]
ComLock: How bout NestedRecord?

modules/lib/core/index.d.ts
ComLock: Yeah I think we are mixing to extremes, instead of providing both options...

modules/lib/core/index.d.ts
ComLock: Yeah, but it's used a lot, so it has to keep it's generics.

modules/lib/core/index.d.ts
```diff
@@ -52,23 +67,115 @@ export interface PublishInfo {
...
+export interface TextComponent {
+    type: 'text'
+    path: string
+    text: string
+}
+
+export type Component<
+    Config extends NestedRecord = NestedRecord,
+    Regions extends Record<string, Region> = Record<string, Region>,
+    Descriptor extends ComponentDescriptor = LayoutDescriptor | PageDescriptor | PartDescriptor,
```
ComLock, line 135: @tajakobsen Had to add a third param here, to match the generics for LayoutComponent, PageComponent and PartComponent

modules/lib/core/index.d.ts
```diff
@@ -52,23 +67,115 @@ export interface PublishInfo {
...
+> {
+    type: 'layout'
+    descriptor: Descriptor
+    config: Config
+    path?: string // Missing in fragmentPreview https://github.com/enonic/xp/issues/10116
+    regions: Regions;
+}
+type LayoutDescriptor = keyof XpLayoutMap;
+type Layout = LayoutDescriptor extends any // this lets us iterate over every member of the union
+    ? LayoutComponent<LayoutDescriptor>
```
ComLock, line 91: Ok, this wasn't thought through. Since LayoutComponent doesn't look up config from XpLayoutMap, this won't work

modules/lib/core/index.d.ts
tajakobsen: I would reorder the type parameters. Since `Regions` are not used for every type in the union it should come last. 
 1. Config
 2. Descriptor
 3. Regions

modules/lib/core/index.d.ts
```diff
@@ -52,23 +67,115 @@ export interface PublishInfo {
...
     path: string;
-    type: 'page' | 'layout' | 'part';
+}
+
+export interface LayoutComponent<
+    Descriptor extends ComponentDescriptor = ComponentDescriptor,
+    Config extends NestedRecord = NestedRecord,
+    Regions extends
+        Record<string, Region<(FragmentComponent | PartComponent | TextComponent)[]>> = 
+        Record<string, Region<(FragmentComponent | Part          | TextComponent)[]>>
```
tajakobsen, line 81: I think you should (for all Component-types) reorder the type parameters so that `Config` comes first. In many cases `Config` is the only thing you need, so it's nice to only have to pass in `Config`. I suggest `Regions` still come last to make this as similar as possible to `PartComponent`
 1. Config
 2. Descriptor
 3. Regions

modules/lib/core/index.d.ts
tajakobsen: This is also consistent with the order of type parameters in `Content`

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -471,9 +471,10 @@ interface GetCurrentComponentHandler<Config extends object = object,
  *
  * @returns {object|null} The current component as JSON.
  */
-export function getComponent<Config extends object = object,
+export function getComponent<
+    Config extends NestedRecord = NestedRecord,
     Regions extends Record<string, Region> = Record<string, Region>,
-    >(): Component<Config, Regions> | null {
+>(): Component<Config, Regions> | null {
```
tajakobsen, line 477: I suggest we let this function take 1 type parameter that extends  `Component`. This would be consistent with how we work with content.

```typescript
export function getComponent<
    _Component extends Component = Component
>(): _Component | null {
```

ComLock: I agree, it's a breaking change though.

modules/lib/core/index.d.ts
ComLock: Well, if we have a registry where descriptor is the key, the descriptor should be all you need, not config?
With only descriptor the "full" shape is described (via registry).
With only config, descriptor remains `${string}:${string}`

modules/lib/core/index.d.ts
ComLock: I would definetly have Regions as the third param, it's a breaking change though.

(I would make descriptor the first key though). Just a config would not tell whether the component is Layout, Page or Part...

modules/lib/core/index.d.ts
ComLock: With descriptor first, this is now possible:

```typescript
getComponent<Component<'app.name:partDescriptor'>>()
```

Fragment and text doesn't have descriptors, but it doesn't matter because they don't have Config or Regions either.

```typescript
getComponent<FragmentComponent>()
```

alansemenov (approved):

modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
edloidas: I suggest you postpone this until XP 8 types then.

PR #10119, opened 2023-04-14
Fix 10118 Typing for Content doesn't match fragment content

ComLock:
This one might need some discussion...

Should the Content Generic cover everything, or just the most common cases?
Since getContent Hit parameter extends Content it must support FragmentContent too...

Do we have more "base" ContentTypes that are different from FragmentContent and PageContent?

### DIFF
--- a/modules/lib/core/index.d.ts
+++ b/modules/lib/core/index.d.ts
@@ -151,30 +151,25 @@ export interface Region<Components extends Component[] = Component[]> {
 // ─────────────────────────────────────────────────────────────────────────────
 // Content
 // ─────────────────────────────────────────────────────────────────────────────
-export interface Content<
-    Data = Record<string, unknown>,
-    Type extends string = string,
-    Page extends Component = Component,
-    > {
+export interface CommonContent {
     _id: string;
     _name: string;
     _path: string;
-    _score: number;
+    _score?: number;
     creator: UserKey;
-    modifier: UserKey;
+    modifier?: UserKey;
     createdTime: string;
-    modifiedTime: string;
+    modifiedTime?: string;
     owner: string;
-    data: Data;
-    type: Type;
+    
     displayName: string;
     hasChildren: boolean;
     language?: string;
     valid: boolean;
-    originProject: string;
+    originProject?: string;
     childOrder?: string;
     _sort?: object[];
-    page: Page;
+    
     x: XpXData;
     attachments: Record<string, Attachment>;
     publish?: PublishInfo;
@@ -185,6 +180,27 @@ export interface Content<
     inherit?: ('CONTENT' | 'PARENT' | 'NAME' | 'SORT')[];
     variantOf?: string;
 }
+
+export interface FragmentContent extends CommonContent {
+    fragment: LayoutComponent | PartComponent
+    type: 'portal:fragment'
+}
+
+export interface PageContent<
+    Data = Record<string, unknown>,
+    Type extends string = string,
+    Page extends Component = Component,
+> extends CommonContent {
+    data: Data;
+    page: Page;
+    type: Type;
+}
+
+export type Content<
+    Data = Record<string, unknown>,
+    Type extends string = string,
+    Page extends Component = Component,
+> = FragmentContent | PageContent;
 
 // Compliant with npm module ts-brand
 type Brand<
--- a/modules/lib/core/index.d.ts
+++ b/modules/lib/core/index.d.ts
@@ -200,7 +200,7 @@ export type Content<
     Data = Record<string, unknown>,
     Type extends string = string,
     Page extends Component = Component,
-> = FragmentContent | PageContent;
+> = FragmentContent | PageContent<Data, Type, Page>;
 
 // Compliant with npm module ts-brand
 type Brand<
--- a/modules/lib/core/index.d.ts
+++ b/modules/lib/core/index.d.ts
@@ -151,56 +151,110 @@ export interface Region<Components extends Component[] = Component[]> {
 // ─────────────────────────────────────────────────────────────────────────────
 // Content
 // ─────────────────────────────────────────────────────────────────────────────
-export interface CommonContent {
+export interface CommonContent<
+    Type extends string = string
+> {
+    // Literal member to discriminate on
+    type: Type;
+    // Required properties
     _id: string;
     _name: string;
     _path: string;
-    _score?: number;
-    creator: UserKey;
-    modifier?: UserKey;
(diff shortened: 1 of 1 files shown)


modules/lib/core/index.d.ts
```diff
@@ -140,6 +135,27 @@ export interface Content<
...
+    data: Data;
+    page: Page;
+    type: Type;
+}
+
+export type Content<
+    Data = Record<string, unknown>,
+    Type extends string = string,
+    Page extends Component = Component,
+> = FragmentContent | PageContent;
```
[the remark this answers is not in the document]
tajakobsen: ## Problematic union

I think this union will be problematic. I think it will force us to check `content.type !== "portal:fragment"` every time we use `Content`.

How about you just build the fragment support into the existing `Content`?

```diff
export interface Content<
    Data = Record<string, unknown>,
    Type extends string = string,
    Page extends Component = Component,
> {
    _id: string;
    _name: string;
    _path: string;
    _score?: number;
    creator: UserKey;
    modifier?: UserKey;
    createdTime: string;
    modifiedTime?: string;
    owner: string;
    data: Data;
    type: Type;
    displayName: string;
    hasChildren: boolean;
    language?: string;
    valid: boolean;
    originProject?: string;
    childOrder?: string;
    _sort?: object[];
    page: Page;
    x: XpXData;
    attachments: Record<string, Attachment>;
    publish?: PublishInfo;
    workflow?: {
        state: 'IN_PROGRESS' | 'PENDING_APPROVAL' | 'REJECTED' | 'READY';
        checks?: Record<string, 'PENDING' | 'REJECTED' | 'APPROVED'>;
    };
    inherit?: ('CONTENT' | 'PARENT' | 'NAME' | 'SORT')[];
    variantOf?: string;
+   fragment: Type extends "portal:fragment" 
+      ? LayoutComponent | PartComponent
+      : undefined;
}
```

If someone needs to handle fragments in their code, they can do this:

```typescript
type ContentFragment = Content<{}, "portal:fragment">; // This should have `fragment` on it.
```

----

## Problematic complexity

I think the type `Content` is the most important TypeScript-type in all of XP. It is imperative that developers can understand it and work with it easily.

I would be very careful to make it more complex (like a union).

tajakobsen (changes requested):
I suggest you extend the existing `Content` with a `fragment` property instead of turning it into a union.

modules/lib/core/index.d.ts
```diff
@@ -140,6 +135,27 @@ export interface Content<
...
+    Data = Record<string, unknown>,
+    Type extends string = string,
+    Page extends Component = Component,
+> extends CommonContent {
+    data: Data;
+    page: Page;
+    type: Type;
+}
+
+export type Content<
```
tajakobsen, line 153: This will be a problem.

If I say I want an article-content by doing `contentLib.get<Content<Article>>()`, this will not give an `Content<Article>` anymore. It will give you the type `FragmentContent | Content<Article>`.

It will insist to make me **always** have to check `Content.type` if something is `FragmentContent` even if I told it the result is `Content<Article>`.

I will repost my solution from my last review here for you to consider again:

```diff
export interface Content<
    Data = Record<string, unknown>,
    Type extends string = string,
    Page extends Component = Component,
> {
    _id: string;
    _name: string;
    _path: string;
    _score?: number;
    creator: UserKey;
    modifier?: UserKey;
    createdTime: string;
    modifiedTime?: string;
    owner: string;
    data: Data;
    type: Type;
    displayName: string;
    hasChildren: boolean;
    language?: string;
    valid: boolean;
    originProject?: string;
    childOrder?: string;
    _sort?: object[];
    page: Page;
    x: XpXData;
    attachments: Record<string, Attachment>;
    publish?: PublishInfo;
    workflow?: {
        state: 'IN_PROGRESS' | 'PENDING_APPROVAL' | 'REJECTED' | 'READY';
        checks?: Record<string, 'PENDING' | 'REJECTED' | 'APPROVED'>;
    };
    inherit?: ('CONTENT' | 'PARENT' | 'NAME' | 'SORT')[];
    variantOf?: string;
+   fragment: Type extends "portal:fragment" 
+      ? LayoutComponent | PartComponent
+      : undefined;
}
```

tajakobsen (changes requested):
@alansemenov The issues from my previous code review has not been fixed or commented on yet.

I will repost my proposed solution from my previous review again here.

modules/lib/core/index.d.ts
ComLock: What about this madness? Not certain I like it...
Screenshot 2023-05-02 at 15 04 13

tajakobsen (changes requested):
I think the tradeoff where a lot more complexity is added to support a little used feature – fragments – and isn't worth it here.  From a developer standpoint we don't encounter fragments that often.

I did really like your suggestion of using one field to configure either `page` or `fragment`. It is a really good one, since they are mutually exclusive.

So for simplicity I would keep just one type for `Content`!

I would not change the order of the type-parameters of `Content` it would break pretty much every project, and create a whole lot of pain. Also, for many cases `Data` is the only field you need, and it's nice to only need to pass that one field in.

---

So here is my suggestion. Let `_Component` be wide enough to cover both `page` and `fragment`. It will be possible to pass in the wrong shape as `_Component`, but it is worth the tradeoff to keep `Content` very simple.

```typescript
export interface Content<
    Data = Record<string, unknown>,
    Type extends string = string,
    _Component extends Component = Component,
> {
    _id: string;
    _name: string;
    _path: string;
    _score?: number;
    creator: UserKey;
    modifier?: UserKey;
    createdTime: string;
    modifiedTime?: string;
    owner: string;
    data: Data;
    type: Type;
    displayName: string;
    hasChildren: boolean;
    language?: string;
    valid: boolean;
    originProject?: string;
    childOrder?: string;
    _sort?: object[];
    page: Type extends "portal:fragment" 
      ? undefined
      : _Component;
    x: XpXData;
    attachments: Record<string, Attachment>;
    publish?: PublishInfo;
    workflow?: {
        state: 'IN_PROGRESS' | 'PENDING_APPROVAL' | 'REJECTED' | 'READY';
        checks?: Record<string, 'PENDING' | 'REJECTED' | 'APPROVED'>;
    };
    inherit?: ('CONTENT' | 'PARENT' | 'NAME' | 'SORT')[];
    variantOf?: string;
   fragment: Type extends "portal:fragment" 
      ? _Component
      : undefined;
}
```

ComLock:
I want to close this PR in favour of https://github.com/enonic/xp/pull/10141/files

Issue #10135, opened 2023-05-04
add WebappResource to the management endpoint


vbradnitski:
add list() method - returns web app names from installed applications.

PR #10136, opened 2023-05-04
add WebappResource to the management endpoint #10135

vbradnitski:


anatol-sialitski (approved):

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/WebappResource.java
```diff
@@ -0,0 +1,58 @@
...
+import com.enonic.xp.app.Application;
+import com.enonic.xp.app.ApplicationService;
+import com.enonic.xp.impl.server.rest.model.WebappJson;
+import com.enonic.xp.jaxrs.JaxRsComponent;
+import com.enonic.xp.resource.Resource;
+import com.enonic.xp.resource.ResourceKey;
+import com.enonic.xp.resource.ResourceService;
+import com.enonic.xp.security.RoleKeys;
+
+@Path("/content/webapps")
```
rymsha, line 25: Is it content? I don't think so

rymsha (changes requested):

PR #10177, opened 2023-06-14
Fix #10176 Avoid having to set fragment: undefined everywhere

### LINKED ISSUE #10176: Typings: Avoid having to set fragment: undefined everywhere

ComLock:
Both fragment and page has to be optional, or we have to convert the interface to a type which has a conditional union or something similar:
https://github.com/sindresorhus/type-fest/blob/main/source/require-exactly-one.d.ts
https://stackoverflow.com/questions/49524414/typescript-interface-conditional-optional-parameters

This also support 'base:folder' by allowing page to be an empty object.

And fixes the fact that data is an empty object for portal:fragment, not undefined.

### DIFF
--- a/modules/lib/core/index.d.ts
+++ b/modules/lib/core/index.d.ts
@@ -178,7 +178,7 @@ export interface Content<
     _Component extends (
         Type extends 'portal:fragment'
             ? LayoutComponent | PartComponent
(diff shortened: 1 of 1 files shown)


tajakobsen (approved):
This looks coorect

PR #10208, opened 2023-06-29
Fix 10207 Access property, not use Generics

ComLock:


tajakobsen (approved):
It looks correct now! Good job :+1:

edloidas (approved):

PR #10214, opened 2023-07-18
Page resolving: add flag to resolve empty regions in Content API

ashklianko:
- Introduced xp.content config variable resolveEmptyRegions (set to 'false' by default) that allows to switch between previous content page resolving behavior that forces injection of all descriptor's regions into page/layout and new one where content's page object is returned as it is 
- Updated core-content classes to read config variable and build page object according to it 
- Updated portal-impl classes to read descriptor and inject empty regions into page if it is not being made by content api, and not to read and inject regions if it is done by content api

### DIFF
--- a/modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/ContentOutboundDependenciesIdsResolver.java
+++ b/modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/ContentOutboundDependenciesIdsResolver.java
@@ -18,12 +18,9 @@ class ContentOutboundDependenciesIdsResolver
 {
     private final ContentService contentService;
 
-    private final ContentDataSerializer contentDataSerializer;
-
     ContentOutboundDependenciesIdsResolver( final ContentService contentService )
     {
         this.contentService = contentService;
-        this.contentDataSerializer = new ContentDataSerializer();
     }
 
     public ContentIds resolve( final ContentId contentId )
@@ -40,7 +37,7 @@ class ContentOutboundDependenciesIdsResolver
         final PropertySet contentPageData = new PropertyTree().getRoot();
         if ( content.getPage() != null )
         {
-            contentDataSerializer.toPageData( content.getPage(), contentPageData );
+            new ContentDataSerializer().toPageData( content.getPage(), contentPageData );
         }
 
         final Stream<Property> extraDataDependencies = content.hasExtraData() ? content.getAllExtraData().
--- a/modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/CreateContentCommand.java
+++ b/modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/CreateContentCommand.java
@@ -35,9 +35,6 @@ import com.enonic.xp.node.Node;
 import com.enonic.xp.node.NodeAccessException;
 import com.enonic.xp.node.NodeAlreadyExistAtPathException;
 import com.enonic.xp.node.RefreshMode;
-import com.enonic.xp.page.PageDescriptorService;
-import com.enonic.xp.region.LayoutDescriptorService;
-import com.enonic.xp.region.PartDescriptorService;
 import com.enonic.xp.schema.content.ContentType;
 import com.enonic.xp.schema.content.GetContentTypeParams;
 import com.enonic.xp.security.PrincipalKey;
@@ -56,21 +53,12 @@ final class CreateContentCommand
 
     private final FormDefaultValuesProcessor formDefaultValuesProcessor;
 
-    private final PageDescriptorService pageDescriptorService;
-
-    private final PartDescriptorService partDescriptorService;
-
-    private final LayoutDescriptorService layoutDescriptorService;
-
     private CreateContentCommand( final Builder builder )
     {
         super( builder );
         this.params = builder.params;
         this.mediaInfo = builder.mediaInfo;
         this.formDefaultValuesProcessor = builder.formDefaultValuesProcessor;
-        this.pageDescriptorService = builder.pageDescriptorService;
-        this.partDescriptorService = builder.partDescriptorService;
-        this.layoutDescriptorService = builder.layoutDescriptorService;
     }
 
     static Builder create()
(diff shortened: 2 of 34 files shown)


modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/page/region/GetComponentByKeyCommand.java
```diff
@@ -38,21 +36,11 @@ public Component execute()
         {
             return LayoutComponent.create().
                 descriptor( layoutDescriptor.getKey() ).
-                regions( buildLayoutRegions( layoutDescriptor ) ).
                 build();
         }
         return null;
     }
 
-    private LayoutRegions buildLayoutRegions( final LayoutDescriptor layoutDescriptor )
```
rymsha, line 47: why this method is no longer needed?

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/FullContentNodeTranslator.java
```diff
@@ -0,0 +1,127 @@
...
+import com.enonic.xp.content.ContentPath;
+import com.enonic.xp.content.Contents;
+import com.enonic.xp.context.ContextAccessor;
+import com.enonic.xp.core.impl.content.serializer.ContentDataSerializer;
+import com.enonic.xp.node.Node;
+import com.enonic.xp.node.NodePath;
+import com.enonic.xp.node.NodeService;
+import com.enonic.xp.node.Nodes;
+
+public class FullContentNodeTranslator
```
rymsha, line 20: this class looks like ContentNodeTranslator duplicate with minor changes. Can it be one class with `if` instead?

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/serializer/FullComponentDataSerializerProvider.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.enonic.xp.region.LayoutDescriptorService;
+
+public class FullComponentDataSerializerProvider extends ComponentDataSerializerProvider
+{
+
+    public FullComponentDataSerializerProvider( final LayoutDescriptorService layoutDescriptorService )
+    {
+        super();
+        this.layoutDataSerializer = new FullLayoutComponentDataSerializer( layoutDescriptorService, regionDataSerializer );
```
rymsha, line 11: this looks a bit strange that we change filed already assigned. in parent constructor. probably better to have protected constructor with an argument

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/rendering/FragmentRenderer.java
```diff
@@ -32,14 +37,18 @@ public final class FragmentRenderer
...
     private final ContentService contentService;
 
+    private final LayoutDescriptorService layoutDescriptorService;
+
     private final RendererDelegate rendererDelegate;
 
     private final FragmentPageResolver fragmentPageResolver = new FragmentPageResolver();
 
-    public FragmentRenderer( final ContentService contentService, final RendererDelegate rendererDelegate )
+    public FragmentRenderer( final ContentService contentService, final LayoutDescriptorService layoutDescriptorService,
```
rymsha, line 46: Why LayoutDscriptorService is needed, but PartDescriptorService is not? it is needed in other paces

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/CreateContentCommand.java
```diff
@@ -108,6 +96,7 @@ private Content doExecute()
             .xDataService( this.xDataService )
             .partDescriptorService( this.partDescriptorService )
             .layoutDescriptorService( this.layoutDescriptorService )
+            .contentDataSerializer( this.translator.getContentDataSerializer() )
```
rymsha, line 99: maybe doe to this change we'll start to save empty regions into data. Need to verify

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/ContentOutboundDependenciesIdsResolver.java
```diff
@@ -20,10 +20,10 @@ class ContentOutboundDependenciesIdsResolver
 
     private final ContentDataSerializer contentDataSerializer;
 
-    ContentOutboundDependenciesIdsResolver( final ContentService contentService )
+    ContentOutboundDependenciesIdsResolver( final ContentService contentService, final ContentDataSerializer contentDataSerializer )
```
rymsha, line 23: it looks strange that Outbound IdsResolver needs data serializer

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/serializer/ComponentDataSerializerProvider.java
```diff
@@ -1,34 +1,56 @@
...
+        this( Optional.empty() );
+    }
+
+    protected ComponentDataSerializerProvider( final LayoutDescriptorService layoutDescriptorService )
+    {
+        this( Optional.ofNullable( layoutDescriptorService ) );
+    }
 
-    public ComponentDataSerializerProvider( )
+    private ComponentDataSerializerProvider( final Optional<LayoutDescriptorService> optionalLayoutDescriptorService )
```
rymsha, line 37: probably can be simplified without need to use Optional. null check works as good

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/serializer/ComponentDataSerializerProvider.java
```diff
@@ -1,34 +1,56 @@
...
-public final class ComponentDataSerializerProvider
+public class ComponentDataSerializerProvider
 {
-    private final PartComponentDataSerializer partDataSerializer;
+    protected final PartComponentDataSerializer partDataSerializer;
+
+    protected final TextComponentDataSerializer textDataSerializer;
 
-    private final TextComponentDataSerializer textDataSerializer;
+    protected LayoutComponentDataSerializer layoutDataSerializer;
```
rymsha, line 19:
```suggestion
    protected final LayoutComponentDataSerializer layoutDataSerializer;
```

rymsha (approved):

Issue #10263, opened 2023-09-08
Incorrect type for `siteConfig` parameter in Project API


alansemenov:
`siteConfig` is an array of objects ([create](https://developer.enonic.com/docs/xp/stable/api/lib-project#create), [modify](https://developer.enonic.com/docs/xp/stable/api/lib-project#modify)), however in our current types it's defined as a single object ([create](https://github.com/enonic/xp/blob/master/modules/lib/lib-project/src/main/resources/lib/xp/project.ts#L40), [modify](https://github.com/enonic/xp/blob/master/modules/lib/lib-project/src/main/resources/lib/xp/project.ts#L50)).

PR #10264, opened 2023-09-08
Incorrect type for siteConfig parameter in Project API #10263

alansemenov:


edloidas (approved):

ComLock (approved):
I have not testes, but that seems consistent with the docs
https://developer.enonic.com/docs/xp/stable/api/lib-project#create

rymsha (changes requested):
I don't think it is correct, either

Because portal has
```
export type Site<Config> = Content<{
    description?: string;
    siteConfig: SiteConfig<Config> | SiteConfig<Config>[];
}, 'portal:site'>;

export interface SiteConfig<Config> {
    applicationKey: string;
    config: Config;
}

```

or maybe it is correct, because we do expect an array in handler, but very inconvenient. 
And `siteConfig` have a specific structure, too: it must nave applicationKey and config, not just random fields.

alansemenov:
@rymsha updated

ComLock (approved):

Issue #10192, opened 2023-06-19
Vhost Config


rymsha:
Custom config per vhost

Example: `mapping.mymapping.config.portal.mediaBaseURL = https://cdn.enonic.com/`

Config is available from Context

PR #10267, opened 2023-09-11
Vhost Config #10192

anatol-sialitski:


rymsha (changes requested):

modules/web/web-vhost/src/main/java/com/enonic/xp/web/vhost/impl/mapping/VirtualHostMapping.java
```diff
@@ -19,10 +20,18 @@ public final class VirtualHostMapping
 
     private final int order;
 
+    private final Map<String, String> context;
```
rymsha, line 23: How you guarantee it is immutable?

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/api/MediaApiHandler.java
```diff
@@ -0,0 +1,204 @@
...
+        worker.contentSecurityPolicy = this.contentSecurityPolicy;
+        worker.contentSecurityPolicySvg = this.contentSecurityPolicySvg;
+        return worker.execute();
+    }
+
+    private PortalResponse doHandleAttachment( final PortalRequest portalRequest, final String restPath )
+        throws Exception
+    {
+        Matcher matcher = ATTACHMENT_REST_PATH_PATTERN.matcher( restPath );
+        if ( !matcher.find() )
```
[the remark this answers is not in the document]
rymsha: It is either wrong both places of there the patern starts-ends with `^$`

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/AttachmentUrlBuilder.java
```diff
@@ -14,17 +15,29 @@ final class AttachmentUrlBuilder
     protected void buildUrl( final StringBuilder url, final Multimap<String, String> params )
     {
         super.buildUrl( url, params );
-        appendPart( url, this.portalRequest.getContentPath().toString() );
-        appendPart( url, "_" );
-        appendPart( url, "attachment" );
 
-        if ( this.params.isDownload() )
+        boolean isSlashAPI = portalRequest.getRawPath() != null && portalRequest.getRawPath().startsWith( "/api/" );
```
rymsha, line 19: Get rid of all null checks with rawPath

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/UrlContextHelper.java
```diff
@@ -0,0 +1,18 @@
+package com.enonic.xp.portal.impl.url;
+
+import com.enonic.xp.context.ContextAccessor;
+
+public class UrlContextHelper
```
rymsha, line 5: Entire class to deduplicate one line of code. Me not like

rymsha (changes requested):

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/UrlContextHelper.java
anatol-sialitski: I am not agree with you. That is used at least in two places in AttachentUrlBuilder and ImageUrlBuilder.

modules/web/web-vhost/src/main/java/com/enonic/xp/web/vhost/impl/mapping/VirtualHostMapping.java
anatol-sialitski: https://github.com/enonic/xp/pull/10267/files#diff-de7ef834b70f3087be55fce7e3e1ce37a58ad5976c4b95f0dfc129d2684a6fffR49

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/AttachmentUrlBuilder.java
```diff
@@ -14,17 +15,29 @@ final class AttachmentUrlBuilder
...
+
+        if ( isSlashAPI )
         {
-            appendPart( url, "download" );
+            appendPart( url, "attachment" );
+            appendPart( url, RepositoryUtils.getContentRepoName( this.portalRequest.getRepositoryId() ) );
+            appendPart( url, this.portalRequest.getBranch().toString() );
+            if ( this.params.isDownload() )
+            {
+                params.put( "download", "" );
```
rymsha, line 28: Does it make query string`?download=` ?

modules/web/web-impl/src/main/java/com/enonic/xp/web/impl/context/ContextFilter.java
```diff
@@ -30,6 +32,12 @@ protected void doHandle( final HttpServletRequest req, final HttpServletResponse
         context.getLocalScope().setAttribute( "__currentTimeMillis", System.currentTimeMillis() );
         context.getLocalScope().setSession( new SessionWrapper( req ) );
 
+        final VirtualHost virtualHost = VirtualHostHelper.getVirtualHost( req );
+        if ( virtualHost != null )
+        {
+            virtualHost.getContext().forEach( ( key, value ) -> context.getLocalScope().setAttribute( key, value ) );
```
rymsha, line 38: One annoying thing,  context contains typed values for branch, repository, etc. 
They have to be converted

rymsha (changes requested):

modules/web/web-impl/src/main/java/com/enonic/xp/web/impl/context/ContextFilter.java
anatol-sialitski: How that is related with my changes?

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/AttachmentUrlBuilder.java
anatol-sialitski: Sure

modules/web/web-impl/src/main/java/com/enonic/xp/web/impl/context/ContextFilter.java
rymsha: This pr introduces vhost context feature.
Context supports default repositoryId, branch and authenticationInfo - but it is not possible to set them up using just string.

AuthenticationInfo is probably too complex, but repositoryid and branch should be supported

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/AttachmentUrlBuilder.java
rymsha: I was hoping for `?download` which is equivalent of value `null` instead of empty string

modules/web/web-impl/src/main/java/com/enonic/xp/web/impl/context/ContextFilter.java
anatol-sialitski: Ok, I agree. But I don't want to do it in this PR. I suggest having a separate issue for that because it will be more transparent, and it requires more work than one might think at first.

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/AttachmentUrlBuilder.java
anatol-sialitski: Unfortunately, we cannot do it another way because `params` will be applied after the base part of the URL is built. We can only add params to the map, at least for now.

rymsha (approved):

Issue #10255, opened 2023-08-30
RichTextProcessor must use styles from all installed applications


anatol-sialitski:
The `RichTextProcessor` must use styles from all installed applications.
The current implementation only uses styles from `com.enonic.xp.app.system` and applications that are associated with a site.
This is the reason why the Guillotine app returns empty styles for the `RichText` type.

PR #10284, opened 2023-10-10
RichTextProcessor must use styles from all installed apps #10255

anatol-sialitski:


rymsha (changes requested):
We think it is the wrong place to fix the issue. It is Guillotine specific and must only be fixed there

Issue #10326, opened 2023-11-08
The `/api/<app-name>/...` endpoint must ignore `media` and `idprovider`


anatol-sialitski:
The `/api/<app-name>/...` endpoint must ignore `media` and `idprovider`

PR #10327, opened 2023-11-08
The `/api/<app>/...` endpoint must ignore `media` and `idprovider` #10326

anatol-sialitski:


rymsha (commented):
It would be better to handle all /api/media calls in media handler

rymsha (approved):

PR #10335, opened 2023-11-17
RichTextProcessor must use styles from all installed apps #10255

anatol-sialitski:


rymsha (changes requested):

rymsha (changes requested):
This PR changes from where url gets repo/branch parts. It should not do so.

rymsha (approved):

Issue #10351, opened 2023-12-01
Config to enable multilayer


rymsha:
configured via `com.enonic.xp.project.cfg`
```
multiInheritance = true | false
```

It should not be documented 
It is disabled by default 

When it is set to false, and project/layer is created with more than one parent, exceptions should be thrown

PR #10359, opened 2023-12-05
Config to enable multilayer #10351

vbradnitski:


modules/core/core-project/src/main/java/com/enonic/xp/core/impl/project/ProjectConfig.java
```diff
@@ -0,0 +1,6 @@
+package com.enonic.xp.core.impl.project;
+
+public @interface ProjectConfig
+{
+    boolean multipleParents_allowed() default false;
```
rymsha, line 5:
```suggestion
    boolean multiInheritance() default false;
```

modules/runtime/src/home/config/com.enonic.xp.project.cfg
```diff
@@ -0,0 +1 @@
+#multipleParents.allowed = false
```
rymsha, line 1: this is a "secret" config. we must not make it visible for now.
Delete the file

rymsha (changes requested):

rymsha (approved):

Issue #10417, opened 2024-02-13
Lib `i18n` must not allow foreign application 


anatol-sialitski:
Lib `i18n` must not allow to provide application to read resources

rymsha:
For similar reasons like in https://github.com/enonic/xp/issues/10419#issuecomment-1943771492 we will make an opposite change: remove applicationKey argument from `localize` method. 
- it is not a good idea to read localization messages from an app which is not even guaranteed to exist at the time localization is done (remember, there is no startup order)

PR #10418, opened 2024-02-13
Lib i18n must allow to specify application to read resources #10417

anatol-sialitski:


rymsha:
This is only half-done. getsupportedlocales has the same issue

rymsha (changes requested):
getsupportedlocales has the same issue

anatol-sialitski:
This pr contains fix for getsupportedlocales too

rymsha (approved):

rymsha:
We decided that it should be vice versa - dependencies between different applications should not be allowed.

PR #10519, opened 2024-04-15
Support multiple APIs in one app

anatol-sialitski:
#10514


modules/core/core-api/src/main/java/com/enonic/xp/api/ApiDescriptor.java
```diff
@@ -35,9 +33,21 @@ public boolean isAccessAllowed( final PrincipalKeys principalKeys )
...
     {
-        return ResourceKey.from( applicationKey, API_DESCRIPTOR_PATH );
+        return toResourceKey( this.key, extension );
+    }
+
+    public static ResourceKey toResourceKey( final DescriptorKey key, final String extension )
+    {
+        if ( "api".equals( key.getName() ) )
+        {
+            return ResourceKey.from( key.getApplicationKey(), "api/api." + extension );
```
rymsha, line 45:
```suggestion
            return ResourceKey.from( key.getApplicationKey(), "apis/api." + extension );
```

modules/core/core-api/src/main/java/com/enonic/xp/api/ApiDescriptor.java
```diff
@@ -35,9 +33,21 @@ public boolean isAccessAllowed( final PrincipalKeys principalKeys )
...
+        }
+        else
+        {
+            return ResourceKey.from( key.getApplicationKey(), "api/" + key.getName() + "/" + key.getName() + "." + extension );
```
rymsha, line 49:
```suggestion
            return ResourceKey.from( key.getApplicationKey(), "apis/" + key.getName() + "/" + key.getName() + "." + extension );
```

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/api/ApiDescriptorKeyLocator.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.Set;
+
+import com.google.common.collect.ImmutableSet;
+
+import com.enonic.xp.app.ApplicationKey;
+import com.enonic.xp.page.DescriptorKey;
+import com.enonic.xp.resource.ResourceService;
+
+public class ApiDescriptorKeyLocator
```
rymsha, line 11: why DescriptorKeyLocator is not sufficient?

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/api/ApiAppHandler.java
```diff
@@ -37,7 +38,8 @@
 public class ApiAppHandler
     extends BaseWebHandler
 {
-    public static final Pattern PATTERN = Pattern.compile( "^/api/(?<appKey>(?!media|idprovider)[^/]+)/(?<restPath>.*)?$" );
+    public static final Pattern PATTERN =
+        Pattern.compile( "^/api/(?<appKey>(?!media|idprovider)[^/]+)/(?<apiKey>[^/]+)/(?<restPath>.*)?$" );
```
rymsha, line 42: idprovider is no longer in /api

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/api/ApiDescriptorKeyLocator.java
anatol-sialitski: We must support both, for example, `apis/api.xml` and `apis/myapi/myapi.xml`.

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/api/ApiAppHandler.java
anatol-sialitski: Yes, but that is within `(?!...)` in other word it must not match. I think that part will be extended the rest of reserved words.

rymsha (approved):

PR #10557, opened 2024-05-14
Fix #10555 Make global types importable

### LINKED ISSUE #10555: Make the global types importable
LongTitle: Make the types in @enonic-types/global importable from @enonic-types/core

--

The types in @enonic-types/global are working as intended for Enonic XP server-side code files in src/main/resources/*.ts

However when writing tests the global functions and object must be replaced by mocks.
Since the globals are defined as "const" in @enonic-types/global, it's a type error if one tries to replace them.

Changing the type to var (let) is not a good idea since that's actually wrong when it comes to the actual "runtime" types.

My suggestion is to
1. export the global types in @enonic-types/core
2. make @enonic-types/global dependant on @enonic-types/core
3. import the global types from @enonic-types/core into @enonic-types/global and use them to define the const's

Then if I need any of those global types anywhere, I can manually import only the specific ones I need from @enonic-types/core.

One example is when implementing the classes in Mock-XP, I would like to use the type for the global log object as an interface:

```
import type {Log as LogInterface} from '@enonic-types/core';

export class Log implements LogInterface {
  public debug(...args: unknown[]): void {
    //...
  }
}
```

The same goes then modifying the type of globalThis when mocking in test files:

```
import type {Log} from '@enonic-types/core';

declare module globalThis {
   let log: Log
}

log.debug = () => {
  // no-op silence :)
}
```

I can probably make it even prettier with a local Global Type Library, or a local Global Modifying Module.

But I need the actual types. I don't want to duplicate those types locally.
Define once, use everywhere.

ComLock:


modules/lib/core/index.d.ts
```diff
@@ -1,3 +1,5 @@
+export type { ResourceKey } from '@enonic-types/global';
```
tajakobsen, line 1: Is it possible to make this dependency go the other way (so `ResourceKey` would still live in core)?

I think having core not depend on anything makes it really flexible. Now we would automatically be pulling in everything from global too.

modules/lib/global.d.ts
```diff
@@ -1,130 +1,163 @@
...
+export type ScriptValue = ScriptValueDefinition;
 
-    getList(): object[];
+export interface App {
+    /**
+     * The name of the application.
+     *
+     * @type string
+     */
+    name: string;
```
tajakobsen, line 143: I have always preferred to have `app.name` be a string literal in my projects. This is the main reason I usually roll my own global-types.

```typescript

declare const app: {
  name: "no.item.www";
  version: string;
  config: Record<string, string | undefined>;
};
```

Because then I can do:
```typescript
type ArticleViewPart = PartComponent<`${typeof app.name}:article-view`>;

export function get() {
  const part = getComponent<ArticleViewPart>();
  ...
}
```

---

Could a solution be to not declare `app.name` in globals, but rely on interface merging with a file in the developers' code instead?

modules/lib/core/index.d.ts
ComLock: I was the other way around yesterday (aka global depended upon core) but we flipped it after a meeting.

> Now we would automatically be pulling in everything from global too.
So since global now is a global modifying module, simply importing from it, will add globals. Hmm.

modules/lib/global.d.ts
ComLock: I do see the use for and like your making app.name a string literal in projects.

I'm assuming this PR currently breaks merging, because
```const app``` is a global const, not an interface type.

I should read this:
https://www.typescriptlang.org/docs/handbook/declaration-merging.html

Let's say I make a
interface AppDefinition {...}

And
declare global {
 type App = AppDefinition
 declare const app: App;
}

export type App = AppDefinition

--

That should allow you to "interface merge" global.App?

File: modules/lib/core/index.d.ts (hunk 1 above)
Comment (ComLock) [reply modules/lib/core/index.d.ts:1]: Yeah I don't like the side-effect of importing from core, and getting the global scope modified more than expected.

File: modules/lib/global.d.ts (hunk 2 above)
Comment (ComLock) [reply modules/lib/global.d.ts:143]: One cannot use declaration merging to change app.name from string to 'com.example'.

https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces

Non-function members of the interfaces should be unique. If they are not unique, they must be of the same type. The compiler will issue an error if the interfaces both declare a non-function member of the same name, but of different types.

File: modules/lib/core/index.d.ts (hunk 1 above)
Comment (ComLock) [reply modules/lib/core/index.d.ts:1]: By flipping it back so that global depends on core, then XpLayoutMap, XpPageMap, XpPartMap and XpXData will also exist when setting typeRoot to @enonic-types or including @enonic-types/global in types.

But that seems ok to me.

Comment (ComLock) [reply modules/lib/core/index.d.ts:1]: Can anyone think of a reason why @enonic-types/global can't depend upon @enonic-types/core?

Comment (ComLock) [reply modules/lib/core/index.d.ts:1]: I think as soon as global.d.ts does an import, it can no longer be an "global declaration file" but needs to be a "global modifying module", (which it currently is in this PR) (and which core is too).

ASFAIK "global modifying modules" aren't supposed to declare new globals, only augment existing ones, but it seems to work though.

Comment (ComLock) [reply modules/lib/core/index.d.ts:1]: According to Gemini:

> Can TypeScript global modifying modules declare new global constants?
No

> Can TypeScript global modifying modules declare the shape of new global constants?
Yes

Since global.d.ts only declares the shape of `app` not the value, it's ok.

Comment (rymsha) [reply modules/lib/core/index.d.ts:1]: > Can anyone think of a reason why @enonic-types/global can't depend upon @enonic-types/core?

You - as you said it would e breaking.

Comment (ComLock) [reply modules/lib/core/index.d.ts:1]: It's "breaking" in the sense that a new dependency is added. In such that all @enonic-types packages must be of the same version.

It was also "breaking" because the build system tsconfig.json file "depended" upon the previous release of @enonic-types/core

Comment (rymsha) [reply modules/lib/core/index.d.ts:1]: `ScriptValue` and `ResourceKey` technically belong in core - they represent Java classes exactly and there is always a Java object behind them.

`XpLayoutMap`, `XpPageMap`, `XpPartMap` and `XpXData` - they have similar classes in Java, but in JS they are `Map`, not actual Java objects - but core uses this types for communication in multiple methods

` XpBeans`, `XpLibraries` abstractions don't have functional value fore core - they are not used in core and can be left in global.


The more I look it this, the more I agree, that global needs core types, not core needs global types. But this is how it should be, I have no idea, if it is possible to do without breaking compatibility.

BTW ResourceKey.getApplication return type is `ApplicationKey`, `string`. 
`ApplicationKey` type has one method `getName()` - which is string

Comment (ComLock) [reply modules/lib/core/index.d.ts]: I've flipped it back, and it seems to work.

File: modules/lib/tsconfig.json [hunk 3]
```diff
@@ -1,17 +1,46 @@
...
+        // which are left intact. Changing target also changes the default value
+        // of lib. You may “mix and match” target and lib settings as desired,
+        // but you could just set target for convenience.
+        // NOTE: This is probably correct since we're still building for Nashorn.
         "target": "es5",
+
+        // NOTE: This is probably correct since we're still building for Nashorn.
         "module": "commonjs",
+
+        // TODO: Maybe use "node16" or "bundler" since we're not using file
```
edloidas, line 20: Option `bundler` can only be used when `module` is set to `preserve` or to `es2015` or later. Option `module` must be set to `NodeNext` when option `moduleResolution` is set to `NodeNext`. Also, it's not like we are even targeting `NodeNext`, as the Node.js is not our target platform.
Just let's keep `node`.

modules/lib/tsconfig.json
```diff
@@ -1,17 +1,46 @@
...
+        // TODO: This only matters at buildtime, perhaps move to
+        // tsconfig.build.json
         "sourceMap": false,
+
         "esModuleInterop": true,
         "forceConsistentCasingInFileNames": true,
         "skipLibCheck": true,
+
+        // TODO deprecated TypeScript 5.5
+        // Also why was it ever here?
```
edloidas, line 43: `"noImplicitUseStrict": true` were added to prevent tsc adding "use strict" to the top of the file to match our old JS files, before they were compiled from TS. I think that it doesn't actually matter if we add it or not, so we can drop this property.

modules/lib/tsconfig.json
```diff
@@ -1,17 +1,46 @@
...
-        ],
+
+        "paths": {
+            // So they can be imported from without installing old ones into
+            // node_modules:
+            "@enonic-types/global": ["./global.d.ts"],
+            "@enonic-types/core": ["./core/index.d.ts"]
+        },
+
+        // TODO: This only matters at buildtime, perhaps move to
```
edloidas, line 34: You are free to move this and similar build time options to `tsconfig.build.json`

modules/lib/tsconfig.build.json
```diff
@@ -1,9 +1,11 @@
-{
+{ // This file is used to build typescript declaration files
```
edloidas, line 1: I hope you delete such comments before merging, as they are pretty obvious and only pollute the code. More to that, comments are often forgotten to update after a change in the project concept or structure.

modules/lib/global.d.ts
```diff
@@ -1,210 +1,112 @@
...
-declare const resolve: (path: string) => {
-    getApplicationKey(): string;
-    getPath(): string;
-    getUri(): string;
-    isRoot(): boolean;
-    getName(): string;
-    getExtension(): string;
-};
+//──────────────────────────────────────────────────────────────────────────────
+// Making sure the file is a module
```
edloidas, line 110: Same goes for such comments. Keep it in the commit message. If it's super important to keep it here, one line of code would be enough.

modules/lib/global.d.ts
```diff
@@ -1,210 +1,112 @@
...
+    Log,
+    NewBean as NewBeanType,
+    Resolve,
+    ScriptValue as ScriptValueInterface,
+    XpRequire,
+} from '@enonic-types/core';
+
+
+//──────────────────────────────────────────────────────────────────────────────
+// Declare global types
```
edloidas, line 13: Isn't it obvious what `declare global` is doing? Do we really need three-line comments here? We don't even separate several large code blocks in one large file with those comments.

modules/lib/global.d.ts
```diff
@@ -1,210 +1,112 @@
...
+//──────────────────────────────────────────────────────────────────────────────
+declare global {
+    // TODO: Remove these two? They are now exported by @enonic-types/core and
```
edloidas, line 16: Yes, I think we should remove them, but it will be potential breaking change.

modules/lib/core/index.d.ts
```diff
@@ -356,9 +484,9 @@ export type QueryDsl = {
     exists: ExistsDslExpression;
 };
 
-type SortDirection = 'ASC' | 'DESC';
+export type SortDirection = 'ASC' | 'DESC';
```
edloidas, line 487: Do we actually need these types to be imported? They are only used inside `FieldSortDsl`, so why not just accessing them via `FieldSortDsl['direction']` or `GeoDistanceSortDsl['unit']`, like we do with some other types?

modules/lib/global.d.ts
edloidas: A small addition to what Christian mentioned. Unfortunately, TypeScript does not allow changing the type of parameter during declaration merging. While it is technically possible to do this if `"skipLibCheck: true"` is enabled, the error won't disappear, and I cannot recommend this approach.

Sure, using a specific value as the type for `app.name` works well with custom global types. I do this myself in some projects. However, if the variable's value is used as a type, I don't see any reason why you shouldn't globally define `type AppName = 'my.app'` and use it instead of `typeof app.name`. I haven't encountered scenarios where we need to check the value of `app.name` and where we specifically need to know its exact type (exact string literal).

modules/lib/core/index.d.ts
```diff
@@ -19,6 +21,132 @@ declare global {
...
+//──────────────────────────────────────────────────────────────────────────────
+// Types that are imported by @enonic-types/global
+//──────────────────────────────────────────────────────────────────────────────
+export interface App {
+    /**
+     * The name of the application.
+     *
+     * @type string
+     */
+    name: string;
```
edloidas, line 33: I think we should define all these fields (in global variables) as `readonly`. Do you have any objections? Did I miss something, and did we have a reason not to do so?

modules/lib/tsconfig.json
ComLock: I believe it's about module resolution at build time, not in the produced target code.

node means node9 or older, which version of node are you using in the compile time environment again?...

https://www.typescriptlang.org/tsconfig/#moduleResolution

'node10' (previously called 'node') for Node.js versions older than v10, which only support CommonJS require. You probably won’t need to use node10 in modern code.

modules/lib/tsconfig.build.json
ComLock: > pretty obvious and only pollute the code
I disagree, I find such comments very useful.
Not everyone is a wizard when it comes to tsconfig.json.
It wasn't clear to me how the multiple tsconfig files worked, before I had studied the "whole" build setup.

These files aren't even part of the finished npm packages.
So I don't see the problem.

modules/lib/global.d.ts
ComLock: > Keep it in the commit message
Yes, I like keeping commit message, however commit messages can "disappear" since some people want to squash branches.

modules/lib/core/index.d.ts
ComLock: Yes, they are used in our other type packages.

modules/lib/core/index.d.ts
ComLock: Dunno, seems outside what I'm trying to do in this branch anyway.

modules/lib/tsconfig.json
ComLock: Hmm, when trying node16 I get: Option 'module' must be set to 'Node16' when option 'moduleResolution' is set to 'Node16'.

ComLock: So I guess keeping "node", is the way for now.

modules/lib/global.d.ts
ComLock: Removed

modules/lib/global.d.ts
ComLock: I kept the comment, but removed the lines

modules/lib/global.d.ts
ComLock: I believe we decided a while ago the "breaking changes" in the type system, was ok.

So I'm removing them.

modules/lib/core/index.d.ts
ComLock: If we want to limit the number of exports and imports, I think we should go for namespaces instead. Much nicer syntax and autocompletion UI in IDE's.

modules/lib/tsconfig.build.json
edloidas: **TL;DR:** Project-specific comments should be placed in one place, while technical comments become unnecessary once you learn how things work by reading the docs or talking with your team.

I believe we should maintain a certain level of expertise. For example, there are many things in Gradle build that aren't obvious at first glance, but we don't comment much there. You just go and learn it. If you didn't know something before, well, now you do.

Comments related to the build process can be placed in the `BUILDING.md` file, where the entire process can be explained, not just a small part in one of the files.

As for technical comments, like the one about the "target" property, as I said, we expect a certain level of expertise (I mean any developer in any project). If a developer doesn't know something, they go and read the docs and learn it. I expect that you, I, and everyone else working with this file knows how it works. If they don't, then why should they be making changes based on a few comments?

modules/lib/core/index.d.ts
edloidas: It looks like these are simply re-exports from two other packages. While namespaces are useful, that’s not the issue that they'll fix, as we’re not adding these types to the global scope.

My main concern is whether we really need to export these. It’s easier to add them later if needed, but removing them could be problematic (backward compatibility). So, do you think developers will actually need these?

modules/lib/tsconfig.build.json
ComLock: Good luck when you get older and your brain starts forgetting... FIFO

modules/lib/core/index.d.ts
ComLock: Re-export could be skipped, however that's how we've already done things.

When I write jests tests, mocks, etc, I can import all the types which are used in a library, directly from that library (/lib/xp/name), rather than some from the list, and some from a type package (@enonic-types/root)

modules/lib/tsconfig.build.json
ComLock: TL;DR Applies more to BUILDING.md than the file I'm directly working on.

ComLock: Also who I going to write and maintain BUILDING.md, and how would I know it exists? Should it be pointed to from all the files what it documents? Developers are notoriously "bad" at writing and maintaining documentation. Commenting seems less formal.

modules/lib/core/index.d.ts
edloidas: If we stop re-exporting these types, the only place they’ll be used is in the core, right? Are you certain that these specific types need to be separate exports? We have plenty of other literal unions (like Content['workflow']['state'], Content['inherit'][number], Highlight['encoder'], etc.) that aren’t exported.

If you're using these types or believe others might need them, then it's fine to keep them. Otherwise, I’d prefer not to export them and not to have a dedicated type for them within the code.

modules/lib/tsconfig.build.json
edloidas: No offense meant. I just find that excessive comments can be more of a distraction than a help, often cluttering the code without adding real value.

As for the BUILDING.md file, it can be placed next to the README, so it shouldn’t be hard to find. However, even with comments scattered across different files, understanding how things work might still require switching between them and reading all those comments, which doesn’t guarantee a clear understanding. If the goal is to provide a brief explanation of how things work, it’s better to do it in one place. A simple step-by-step guide, like those usually found under the "Usage"/"Installation" sections of many frameworks and tools, would be more effective, in my opinion. It may not be the task for this issue, of course.

Comments can also become outdated. It's rarely a problem to have comments in our code, but things get complicated with the configurations. For example, some properties in the TSConfig are now deprecated (and you even left a note there), but who’s responsible for updating or removing those comments when things no longer work the way they used to?  It falls on the same developers, and often, that doesn’t work well.

ComLock: > you even left a note there

No, I didn't :)

ComLock: That was a Monty Python reference btw :)
https://www.youtube.com/watch?v=uLlv_aZjHXc

modules/lib/core/index.d.ts
tajakobsen: > ...will also exist when setting typeRoot to @enonic-types

I don't believe that there is any reason to set `compilerOptions.typeRoot` now, since we rely on `compilerOptions.paths` to be able to resolve all library types. (I also just use `compilerOptions.types` to add my global types).

modules/lib/global.d.ts
tajakobsen: I think keeping it the way it is is the easiest for new developers, so that is the way it should be! :)

It's better to widen the types if the model is easier to understand for new developers.

---

> I don't see any reason why you shouldn't globally define type `AppName = 'my.app'`

Very true! This is really a luxury I'm asking for. :)

tajakobsen (approved):
Great job!

I just had one comment regarding activating `strict`-mode.

modules/lib/tsconfig.json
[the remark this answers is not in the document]
ComLock: Yup, that opened up a bunch of things to look at

alansemenov (approved):

tajakobsen (approved):
Looks good

Issue #10545, opened 2024-05-06
Limit mount to site, project, webapp, admin, global


alansemenov:
```
<?xml version="1.0" encoding="UTF-8"?>
<api xmlns="urn:enonic:xp:model:1.0">
<mounts xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:noNamespaceSchemaLocation="example.xsd">
    <mount>global</mount>
    <mount>any_site_endpoint</mount>
    <mount>site_endpoint</mount>
</mounts>
</api>
```

```
<xs:simpleType name="mountType">
        <xs:restriction base="xs:string">
            <xs:enumeration value="api"/>
            <xs:enumeration value="all_sites"/>
            <xs:enumeration value="site"/>
            <xs:enumeration value="all_webapps"/>
            <xs:enumeration value="webapp"/>
            <xs:enumeration value="admin"/>
        </xs:restriction>
    </xs:simpleType>
```

`api` - available on `/api/`
`aal_site`  - available on any site `/[admin/]site/<project>/<branch>/.../some-site/.../_/...` even if application is not set on a site. It does include project root.
`site` - available on sites (and projects) where application is set. It does not include project root. Ignored if `all_sites` is also listed.
`all_webapps` - available on all webapps `/webapp/<app-name>/.../_/...`
`webapp` - available on application's webapp. Ignored if `any_webapp_endpoint` is also listed
`admin` - available on  `/admin/tool/.../_/...` NOTE: Does not enable it on `/admin/site`, `/admin`, /admin/rest`, etc

sigdestad:
Can we add the XML descriptor spec to this task?

PR #10575, opened 2024-05-24
Limit mount to site, project, webapp, admin, global #10545

anatol-sialitski:


modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/SlashApiHandler.java
```diff
@@ -111,6 +140,105 @@ public WebResponse handle( final WebRequest webRequest )
...
+                }
+                case ANY_PROJECT_ENDPOINT, PROJECT_ENDPOINT ->
+                {
+                    Matcher projectMatcher = MOUNT_ENDPOINT_PATTERN.matcher( rawPath );
+                    if ( !projectMatcher.find() )
+                    {
+                        return false;
+                    }
+
+                    ProjectName projectName = ProjectName.from( projectMatcher.group( "project" ) );
```
rymsha, line 164: exception may be thrown or null returned. need to return false if invalid name

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/SlashApiHandler.java
```diff
@@ -111,6 +140,105 @@ public WebResponse handle( final WebRequest webRequest )
...
+                {
+                    Matcher webappMatcher = WEBAPP_ENDPOINT_PATTERN.matcher( rawPath );
+                    if ( !webappMatcher.find() )
+                    {
+                        return false;
+                    }
+
+                    if ( !apiDescriptor.hasMount( ApiMount.ANY_WEBAPP_ENDPOINT ) && apiMount == ApiMount.WEBAPP_ENDPOINT )
+                    {
+                        ApplicationKey baseAppKey = ApplicationKey.from( webappMatcher.group( "baseAppKey" ) );
```
rymsha, line 227: exception may be thrown or null returned. need to return false if invalid name

modules/core/core-api/src/main/java/com/enonic/xp/api/ApiDescriptor.java
```diff
@@ -11,10 +13,13 @@ public final class ApiDescriptor
 
     private final PrincipalKeys allowedPrincipals;
 
+    private final List<ApiMount> mounts;
+
     private ApiDescriptor( final Builder builder )
     {
         this.key = builder.key;
         this.allowedPrincipals = builder.allowedPrincipals;
+        this.mounts = builder.mounts;
```
rymsha, line 22: List.copyOf

modules/core/core-api/src/main/java/com/enonic/xp/api/ApiMount.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.util.Locale;
+
+public enum ApiMount
+{
+    GLOBAL, ANY_SITE_ENDPOINT, SITE_ENDPOINT, ANY_PROJECT_ENDPOINT, PROJECT_ENDPOINT, ANY_WEBAPP_ENDPOINT, WEBAPP_ENDPOINT, ADMIN_ENDPOINT;
+
+    public static ApiMount from( final String value )
+    {
+        return value != null ? ApiMount.valueOf( value.toUpperCase( Locale.ROOT ) ) : null;
```
rymsha, line 11: why making it null save? we are reading it from valid XML
why returning null - this will make user's code responsible to check for null

rymsha (changes requested):

PR #10584, opened 2024-05-29
Limit mount to root context-path #10546

anatol-sialitski:


modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/SlashApiHandler.java
```diff
@@ -55,11 +55,16 @@ public class SlashApiHandler
     private static final Pattern MOUNT_SITE_ENDPOINT_PATTERN =
         Pattern.compile( "^/(admin/)?site/(?<project>[^/]+)/(?<branch>[^/]+)(?<contentPath>/.*?)/_/" );
 
-    private static final Pattern MOUNT_WEBAPP_ENDPOINT_PATTERN = Pattern.compile( "^/webapp/(?<baseAppKey>[^/]+)(?<restPath>/.*?)/_/" );
+    private static final Pattern MOUNT_WEBAPP_ENDPOINT_PATTERN = Pattern.compile( "^/webapp/(?<baseAppKey>[^/]+)(?<restPath>.*?)/_/" );
 
     private static final List<String> RESERVED_APP_KEYS =
         List.of( "attachment", "image", "error", "idprovider", "service", "asset", "component", "widgets", "media" );
 
+    private static final String WEBAPP_DEFAULT_CONTEXT_PATH_PATTERN = "^/webapp/([^/]+)/_/.*";
```
rymsha, line 63: Compile it, and give group a name

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/SlashApiHandler.java
```diff
@@ -55,11 +55,16 @@ public class SlashApiHandler
     private static final Pattern MOUNT_SITE_ENDPOINT_PATTERN =
         Pattern.compile( "^/(admin/)?site/(?<project>[^/]+)/(?<branch>[^/]+)(?<contentPath>/.*?)/_/" );
```
rymsha, line 56: How about project path? I'm not sure it is captured by this regex

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/SlashApiHandler.java
```diff
@@ -213,12 +227,21 @@ private boolean verifyRequestMountedOnSites( final ApiDescriptor apiDescriptor,
...
+            final Project project = projectService.get( ProjectName.from( portalRequest.getRepositoryId() ) );
+            if ( project == null || project.getSiteConfigs().get( apiDescriptor.key().getApplicationKey() ) == null )
+            {
+                return false;
+            }
+        }
+
+        if ( apiDescriptor.getContextPath() == ApiContextPath.DEFAULT )
+        {
+            final String siteRelativePath = (String) ContextAccessor.current().getAttribute( SITE_RELATIVE_PATH_CTX_ATTRIBUTE );
```
rymsha, line 243: If we only use it in one place it is better to propagate it via method parameters

rymsha (changes requested):

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/SlashApiHandler.java
anatol-sialitski: We do not have requirements to support project path at least for now.

rymsha (approved):

Issue #10607, opened 2024-06-24
Media scope


anatol-sialitski:
There is a security implication: some intranet sites may be protected only by VHost - which is actually a bad practice. Due to this fact we need to introduce VHost config with a list of allowed projects (media projects)

`mediaService.scope = <project1:draft>, <project2>`

If not specified:

For endpoint mount limit the project to the one specified in site's path (allow from self) `/(admin)/site/<project>/<branch>/site-path/_/media/.../<project>(:branch))`

For Slash API - allow any project and branch.

Note that content permissions will further limit the access to specific media.

PR #10613, opened 2024-06-26
Media scope #10607

anatol-sialitski:


rymsha (changes requested):
if media context matches site context and mediaService.scope is not defined - allow media endpoint

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/MediaHandler.java
```diff
@@ -122,22 +124,53 @@ public WebResponse handle( final WebRequest webRequest, final WebResponse webRes
...
+                Arrays.stream( mediaServiceScope.split( ",", -1 ) ).map( String::trim ).collect( Collectors.toSet() );
+
+            if ( scopes.stream().noneMatch( scope -> scope.equals( projectContext ) ) )
+            {
+                throw WebException.notFound( "Not a valid media url pattern" );
+            }
+        }
+        else
+        {
+            if ( webRequest.getRawPath().startsWith( "/site/" ) || webRequest.getRawPath().startsWith( "/admin/site/" ) )
```
rymsha, line 162: check that it is PortalRequest - if not  throw 404
in case it is - use method `PortalRequest.isSiteBase`

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/MediaHandler.java
```diff
@@ -122,22 +124,53 @@ public WebResponse handle( final WebRequest webRequest, final WebResponse webRes
...
+            {
+                // the webRequest was already handled by SiteHandler/AdminSiteHandler
+                final PortalRequest portalRequest = (PortalRequest) webRequest;
+                if ( !( repositoryId.equals( portalRequest.getRepositoryId() ) && branch.equals( portalRequest.getBranch() ) ) )
+                {
+                    throw WebException.notFound( "Not a valid media url pattern" );
+                }
+            }
```
rymsha, line 170: it is confusing that `/api` is not explicitly coded. Handler is supposed to work on `/api` and `(/admin)/site`. I can't tell if it actually covered.

rymsha (changes requested):

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/MediaHandler.java
anatol-sialitski: testMediaEndpointDoesNotAllowedIfSiteContextDoesNotMatch

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/MediaHandler.java
anatol-sialitski: Great idea to check that this is a `PortalRequest` and that `PortalRequest.isSiteBase` is true. However, if we throw a 404 error, it would imply that it works only for the site, which is incorrect.

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/MediaHandler.java
```diff
@@ -117,27 +120,61 @@ public WebResponse handle( final WebRequest webRequest, final WebResponse webRes
...
+        if ( webRequest.getEndpointPath() == null )
+        {
+            return;
+        }
+
+        final String mediaServiceScope = VirtualHostContextHelper.getMediaServiceScope();
+        if ( mediaServiceScope != null )
+        {
+            final List<String> scopes =
+                MEDIA_SCOPE_DELIMITER_PATTERN.splitAsStream( mediaServiceScope ).map( String::trim ).collect( Collectors.toList() );
```
rymsha, line 157: use noneMatch instead of collect

Issue #10624, opened 2024-07-04
Invert how harmonized API is mounted


rymsha:
This task makes changes to #10545 and #10546

Three mount types are completely abandoned: `all_sites`, `all_webapps` and `admin`
`api` mount is implicit - all apis are always mounted on `/api`
`mount` section of api.xml descriptor should be removed
Instead, API mount should be defined via `site.xml` and `webapp.xml` similar to how mounting is described in #10612 

It is no longer possible to mount API with `context-path=any` - the exceptions are  legacy APIs (for instance `component`)

site.xml, webapp.xml and tool.xml can mount both internal, and reference external APIs
```
<apis>
    <api>com.enonic.xp.app.main:launcher</api>
    <api>com.enonic.xp.myapp:</api> <-- unnamed api mount of external app requires colon  
    <api/> <-- mount unnamed api of  application itself (appkey is optional and name is not specified)
    <api>myapi</api> <-- mount named api of  application itself (appkey is optional)
  <apis>
```
When `apis` section exists in application's site.xml, site with such application set can use the listed APIs.
When `apis` section exists in application's tool.xml this admin tool can use the listed APIs
When `apis` section exists in application's webapp.xml this webapp can use the listed APIs. Note that we need to introduce webap.xml

It is not an error to mount APIs that do not exist. Access to such APIs returns 404.

PR #10632, opened 2024-07-08
Invert how harmonized API is mounted #10624

anatol-sialitski:


modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/tool/XmlAdminToolDescriptorParser.java
```diff
@@ -41,5 +53,58 @@ protected void doParse( final DomElement root )
...
+        return apisElement.getChildren( API_DESCRIPTOR_TAG_NAME ).stream().map( this::toApiMountDescriptor ).collect( Collectors.toList() );
+    }
+
+    private ApiMountDescriptor toApiMountDescriptor( final DomElement apiElement )
+    {
+        final ApiMountDescriptor.Builder builder = ApiMountDescriptor.create();
+
+        final String apiMount = apiElement.getValue().trim();
+
+        if ( !apiMount.contains( ":" ) )
```
rymsha, line 76: Don't we have common code that does thin logic already?

modules/core/core-api/src/main/java/com/enonic/xp/xml/parser/ApiMountResolver.java
```diff
@@ -0,0 +1,40 @@
+package com.enonic.xp.xml.parser;
+
+import com.enonic.xp.app.ApplicationKey;
+
+public final class ApiMountResolver
```
rymsha, line 5: move to core-shared

rymsha (approved):

Issue #10627, opened 2024-07-04
Widgets as API


rymsha:
instead of `/admin/tool/_/widgets/com.enonic.app.guillotine/guillotine` there should be 
`/admin/tool/com.enonic.app.contentstudio/main/_/widget/com.enonic.app.guillotine/guillotine`

For this API Dispatcher should allow to register dynamic APIs via Java (`ApiHandler` interface)
And `com.enonic.xp.admin-impl` module should register an API with `Component` properties `app-name = widget` and no api-name

`WidgetHandler` code is basically reused with some minimal changes

rymsha:
there might be some changers necessary in ContentStudio and app-main in order to support correct prefix. Probably should use #10628

PR #10637, opened 2024-07-12
Widgets as API #10627

anatol-sialitski:


modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/portal/ApiWidgetHandler.java
```diff
@@ -0,0 +1,121 @@
...
+import com.enonic.xp.web.WebException;
+import com.enonic.xp.web.WebRequest;
+import com.enonic.xp.web.WebResponse;
+import com.enonic.xp.web.api.ApiHandler;
+
+@Component(immediate = true, service = ApiHandler.class, property = ApiHandler.APPLICATION_KEY_PROPERTY + "=widget")
+public class ApiWidgetHandler
+    implements ApiHandler
+{
+    private static final Pattern WIDGET_API_PATTERN = Pattern.compile( "^/(_|api)/widget/(?<appKey>[^/]+)/(?<widgetKey>[^/]+)" );
```
anatol-sialitski, line 32: How about mapping it like this `/_/widget/<appKey>:<widgetKey>`?  Where `apiKey` is combined and in this case it fits  good to common approach with Slash API and Harmonized API.

Issue #10625, opened 2024-07-04
Remove `tool` path segment for admin tools


rymsha:
in XP7 every admin tool is located on
`/admin/tool/{app-name}/{tool-name}` path

We remove `tool` path segment so it becomes
`/admin/{app-name}/{tool-name}` path 

without ability to restore previous behavior

app names `site`, `rest`, `tool`, `event` are reserved. applications with such names do not override existing handlers.
all urls under /admin/tool return 404

PR #10671, opened 2024-09-02
Remove tool path segment for admin tools #10625

anatol-sialitski:


modules/admin/admin-api/src/main/java/com/enonic/xp/admin/widget/WidgetDescriptor.java
```diff
@@ -37,7 +37,7 @@ public class WidgetDescriptor
 
     private final ImmutableMap<String, String> config;
 
-    private static final String URL_PREFIX = "_/widgets/";
+    private static final String URL_PREFIX = "_/widget/";
```
rymsha, line 40: Deprecate/remove getUrl method?

rymsha (approved):

modules/admin/admin-api/src/main/java/com/enonic/xp/admin/widget/WidgetDescriptor.java
anatol-sialitski: It will be done in https://github.com/enonic/xp/issues/10672

Issue #10628, opened 2024-07-04
apiUrl frunction


rymsha:
Similar to how serviceUrl function works apiUrl should generate urls

with site engine
`/(admin/)site/repo/branch</path-to-nearest-site>/_/<app-name>/<api-name>/...`
`/(admin/)site/repo/branch</path-to-nearest-site>/_/<app-name>/...` <- for unnamed APIs

`/path-to-nearest-site` - path to site, if exists. or `/` if there is not site, but just project

with webapp engine
`/webapp/<webapp-name>/_/<app-name>/<api-name>/...`
`/webapp/<webapp-name>/_/<app-name>/...` <- for unnamed APIs

with admin tool engine

`/admin/<admin-tool-path>/_/<app-name>/<api-name>/...`
`/admin/<admin-tool-path>/_/<app-name>/...` <- for unnamed APIs

`admin-tool-path` - consists of app mane holding the tool, and tool name. Ex: `com.enonic.app.main/home`

otherwise 
`/api/<app-name>/<api-name>/...`
`/api/<app-name>/...` <- for unnamed APIs

Just like serviceUrl works - apiUrl should rewrite URL based on vhost source-target information
In case there is no request available fallback to `/api` url structure (exception is #10349)

PR #10683, opened 2024-09-12
apiUrl function #10628

anatol-sialitski:


modules/lib/lib-portal/src/main/resources/lib/xp/portal.ts
```diff
@@ -600,3 +600,36 @@ export function imagePlaceholder(params: ImagePlaceholderParams): string {
     bean.setHeight(params?.height ?? 0);
     return bean.createImagePlaceholder();
 }
+
+export interface ApiUrlParams {
+    toolApplication?: string;
+    toolKey?: string;
```
rymsha, line 606: tool is implicit when requested from/admin 
In XP8 tools will unlikely be able to generate urls to each other as it is not secure. 
Exception is main tool (launcher) and wellcome - but they may have their own custom implementation

rymsha (changes requested):

rymsha (approved):

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/UrlBuilderHelper.java
```diff
@@ -0,0 +1,90 @@
...
+        {
+            str.append( "/" );
+        }
+
+        str.append( normalized );
+    }
+
+    public static String normalizePath( final String value )
+    {
+        if ( value == null )
```
rymsha, line 40: I don't think this function should ever be called with null

rymsha (approved):

PR #10719, opened 2024-10-02
Linter errors after turning on strict mode

alansemenov:


modules/lib/lib-admin/src/main/resources/lib/xp/admin.ts
```diff
@@ -13,16 +13,34 @@ declare global {
...
-const portal = require('/lib/xp/portal');
+interface ApiUrlParams {
+    application: string;
+    api?: string;
+    type?: 'server' | 'absolute' | 'websocket';
+    params?: object;
+    path?: string | string[];
+}
+
+interface Portal {
```
ComLock, line 24: Why not use the types from the source, which is available in the same code repo?

ComLock (changes requested):

modules/lib/lib-io/src/main/resources/lib/xp/io.ts
```diff
@@ -65,7 +65,7 @@ class ResourceImpl
     private res: JavaResource;
 
     constructor(key: string | ResourceKey) {
-        this.res = bean.getResource(key);
+        this.res = bean.getResource(key) ;
```
ComLock, line 68: whitespace oupsie?

anatol-sialitski (approved):

ComLock (approved):

anatol-sialitski:
Awesome work, Alan!

rymsha (approved):

PR #10724, opened 2024-10-11
Add type definitions for Request and Response

ComLock:


modules/lib/core/index.d.ts
```diff
@@ -142,6 +142,318 @@ export interface ScriptValue {
...
+     * Indicates whether the cookie should only be sent over HTTPS (optional).
+     *
+     * @type string
+     */
+    secure?: boolean
+}
+
+export type RequestBranch = 'draft' | 'master';
+export type RequestGetHeaderFunction = (headerName: string) => string | null;
+export type RequestMethod = 'GET' | 'POST' | ' HEAD' | 'OPTIONS' | ' PUT' | 'DELETE' | ' TRACE' | 'CONNECT' | ' PATCH' | 'PROPFIND' | ' PROPPATCH' | 'MKCOL' | ' COPY' | 'MOVE' | ' LOCK' | 'UNLOCK';
```
rymsha, line 213: We don't support many of these methods. but we don't limit this list either if developer wants some. 
My recommendation is to suggest `'GET' | 'POST' | ' HEAD' | 'OPTIONS' | ' PUT' | 'DELETE'` and allow any string, too

modules/lib/core/index.d.ts
```diff
@@ -142,6 +142,318 @@ export interface ScriptValue {
...
+export type RequestScheme = 'http' | 'https';
+
+export type RequestCookies = Record<string, string | undefined>;
+export type ResponseCookies = Record<string, string | ComplexCookie | undefined>;
+
+export type RequestHeaders = Record<string, string | undefined>;
+export type ResponseHeaders = Record<string, string | number | (string | number)[] | undefined>;
+
+export interface DefaultRequestCookies extends RequestCookies {
+    enonic_xp_tour?: string
```
rymsha, line 225: this is only for admin and we deprecate this cookie. Not a good idea to mention it

modules/lib/core/index.d.ts
```diff
@@ -142,6 +142,318 @@ export interface ScriptValue {
...
+     */
+    secure?: boolean
+}
+
+export type RequestBranch = 'draft' | 'master';
+export type RequestGetHeaderFunction = (headerName: string) => string | null;
+export type RequestMethod = 'GET' | 'POST' | ' HEAD' | 'OPTIONS' | ' PUT' | 'DELETE' | ' TRACE' | 'CONNECT' | ' PATCH' | 'PROPFIND' | ' PROPPATCH' | 'MKCOL' | ' COPY' | 'MOVE' | ' LOCK' | 'UNLOCK';
+export type RequestMode = 'edit' | 'inline' | 'live' | 'preview' | 'admin';
+export type RequestParams = Record<string, string | string[]>;
+export type RequestScheme = 'http' | 'https';
```
rymsha, line 216: how about ws/wss ? not sure they are ever visible - should be tested.

modules/lib/core/index.d.ts
```diff
@@ -142,6 +142,318 @@ export interface ScriptValue {
...
+    Host?: string
+    'If-None-Match'?: string
+    Referer?: string
+    'sec-ch-ua'?: string
+    'sec-ch-ua-mobile'?: string
+    'sec-ch-ua-platform'?: string
+    'Sec-Fetch-Dest'?: string
+    'Sec-Fetch-Mode'?: string
+    'Sec-Fetch-Site'?: string
+    'Sec-Fetch-User'?: string
```
rymsha, line 250: some experimental headers whe really never use. Why are they int default? Why they are different case BTW?

modules/lib/core/index.d.ts
```diff
@@ -142,6 +142,318 @@ export interface ScriptValue {
...
+> & {
+    body?: unknown[] | Record<string, unknown> | boolean | number | string | null;
+};
+
+export interface DefaultResponseHeaders extends ResponseHeaders {
+    'Cache-Control'?: string
+    'Content-Encoding'?: string
+    'Content-Type'?: string
+    'Content-Security-Policy'?: string
+    'Date'?: string
```
rymsha, line 306: I don't think XP allows to set this header. it is always overridden

modules/lib/test/RequestImplementation.ts
```diff
@@ -0,0 +1,133 @@
...
+        this.contextPath = contextPath;
+        this.contentType = contentType;
+        this.repositoryId = repositoryId;
+        this.validTicket = validTicket;
+
+        this.body = body;
+    }
+
+    getHeader(header: string): string | null {
+        const value = this.headers[header];
```
rymsha, line 78: actual implementation is case-insensitive

modules/lib/core/index.d.ts
```diff
@@ -142,6 +142,318 @@ export interface ScriptValue {
...
+    // connect?: RequestHandler
+    delete?: RequestHandler
+    get?: RequestHandler
+    head?: RequestHandler
+    options?: RequestHandler
+    // patch?: RequestHandler
+    post?: RequestHandler
+    put?: RequestHandler
+    // trace?: RequestHandler
+    // TODO what about propfind, proppatch, mkcol, copy, move, lock and unlock?
```
rymsha, line 369: yes, what about them? I don't think they are supported out of the box

modules/lib/core/index.d.ts
ComLock: ws/wss are allowed, just not autocompleted

modules/lib/core/index.d.ts
ComLock: Default optional, only for autocomplete. These are the ones in chrome.

Dunno what to use them for on the server-side, so probably can be removed.

Are they in the way though?

modules/lib/core/index.d.ts
ComLock: Which one? Date? Cause some of the other ones I do set in lib-static

modules/lib/test/RequestImplementation.ts
ComLock: RIght, that file is just a test, but good to know. That class could be useful in mock-xp.

modules/lib/core/index.d.ts
ComLock: I'm happy to remove them. They where mentioned in the previous pull request discussions.

modules/lib/core/index.d.ts
ComLock: Date is forbidden in fetch requests https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date

These are response headers though.

And Date is one of those that Chrome uses.

modules/lib/core/index.d.ts
ComLock: I tested the chat in app features, the websocket request doesn't go to the "get" method, but directly to the "webSocketEvent" method, which doesn't receive the request object, but an event object.

ComLock: The first request has webSocket: false, the second has webSocket: true, but is still has scheme: http, the third and ongoing are just a events, not passing through the get function.

ComLock: Added webSocketEvent

ComLock: Now, more like WebSocketEventMapper

modules/lib/test/Controller.test-d.ts
```diff
@@ -0,0 +1,187 @@
...
+    delete?: PageRequestHandler
+    get?: PageRequestHandler
+    head?: PageRequestHandler
+    options?: PageRequestHandler
+    post?: PageRequestHandler
+    put?: PageRequestHandler
+}
+
+const pageRequest /* : PageRequest */ = {
+    method: 'GET' as PageRequest['method'], // Avoid flattening down to string
```
tajakobsen, line 83: Doesn't the use of `as` go against the purpose of checking if this object satisfies a specific type?

modules/lib/test/tsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
+    "compilerOptions": {
+        "lib": ["DOM"]
```
tajakobsen, line 3: Is this correct. This is server side code we are looking at, so the DOM probably shouldn't be available (even for the tests).

tajakobsen (changes requested):
Good job!  I can see that there has gone in a ton of effort into this!

Is it possible for me to be able to test this in Beta, because I'm interested in seeing if there is a lot of friction when changing from "my" controller types to these.

I would also be interested in seeing how you use these types in your application projects. When you have a beta out, can you try to put them in, so that I can see an example of the intended use from your side?

modules/lib/test/tsconfig.json
ComLock: I probably did that because I was mocking log with console, but since I've removed all console mocks and I'm only testing types, not actually running the code, there is no need for that, so I've removed the file entirely.

modules/lib/test/Controller.test-d.ts
ComLock: Yeah, but. I've added some comments to explain what's going on.

I'm specifically avoiding setting the type on the pageRequest object to PageRequest, so the assign test isn't just a dummy comparing PageRequest to PageRequest.

If you just copied the whole object into the assign test that would also work.

It's about how TypeScript "autogenerates" types for untyped things.

In normal code these problems shouldn't exist as the request object comes in from outside (doesn't even exist at compiletime), so the first thing we do is actually say that the shape of request is `AS Request`.

modules/lib/core/index.d.ts
```diff
@@ -142,6 +142,339 @@ export interface ScriptValue {
...
+    body?: string
+    branch?: LiteralUnion<RequestBranch>
+    contentType?: string
+    contextPath?: string
+    cookies: RequestCookies
+    headers: RequestHeaders
+    host: string
+    method: LiteralUnion<RequestMethod>
+    mode: LiteralUnion<RequestMode>
+    params: RequestParams
```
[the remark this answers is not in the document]
ComLock: It's definitely possible to define the shape of params, however how it works with type magic down the line, remains to be tested.

modules/lib/core/index.d.ts
ComLock: I removed the experimental ones and added autocompletion to the Sec-Fetch ones. 

I also added Priority since both Chrome and Firefox sends it.

There are tons of headers:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers
https://en.wikipedia.org/wiki/List_of_HTTP_header_fields

Not really sure which ones should be part of DefaultRequestHeaders.

modules/lib/core/index.d.ts
```diff
@@ -355,46 +356,47 @@ export type HttpFilterNext<
...
-    user: Omit<User,'type'>
+    id: string;
+    params: Record<string, string | string[]>;
+    path: string;
+    user: Omit<User,'type'>;
 }
 
 export type WebSocketEventType = 'open' | 'message' | 'error' | 'close';
 
 export interface WebSocketEvent<T> {
```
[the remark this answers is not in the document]
ComLock: My first implementation was similar to that, because it was based on the data I was able to generate in real life. I just simplified after looking at the Java class behind it.

alansemenov (approved):

tajakobsen (approved):
Good job.

alansemenov:
Any additional comments before this is merged @edloidas?

edloidas (approved):

rymsha (approved):

PR #10745, opened 2024-10-31
Fix #10555 Make global types importable (#10557)

### LINKED ISSUE #10555: Make the global types importable

anatol-sialitski:
* Fix #10555 Make global types importable

* Define and reuse definitions both as global and export

* XpBeans and XpLibraries belongs in globals

* Make global depend on core (not the other way around)

* Remove noImplicitUseStrict which is deprecated in TypeScript 5.5

* Remove NewBean and ScriptValue from global scope. Can be imported from core instead.

* Enable strict and fix most type errors

* Fix TypeScript errors in grid.ts

---------

Co-authored-by: Mikita Taukachou <edloidas@gmail.com>

(cherry picked from commit 1a48ac121aef8d94f6e2e71d3d1b462158fb73d3)

### DIFF
--- a/modules/lib/lib-node/src/main/resources/lib/xp/node.ts
+++ b/modules/lib/lib-node/src/main/resources/lib/xp/node.ts
@@ -24,6 +24,7 @@ import type {
     HighlightResult,
     PrincipalKey,
     QueryDsl,
+    ScriptValue,
     SortDsl,
 } from '@enonic-types/core';
 
@@ -72,6 +73,7 @@ export type {
     QueryDsl,
     RangeDslExpression,
     RoleKey,
+    ScriptValue,
     SingleValueMetricAggregationResult,
     SingleValueMetricAggregationsUnion,
     SortDirection,
@@ -91,6 +93,16 @@ type WithRequiredProperty<T, K extends keyof T> = T & { [P in K]-?: T[P] };
 function checkRequired<T extends object>(obj: T, name: keyof T): void {
     if (obj == null || obj[name] == null) {
         throw `Parameter '${String(name)}' is required`;
+    }
+}
+
+function assertStringArray(value: unknown, name: string): asserts value is string[] {
+    if (!Array.isArray(value)) {
+        throw new TypeError(`${name} must be an array of strings! Isn't even an array!`);
+    }
+
+    if (!value.every(item => typeof item === 'string')) {
+        throw new TypeError(`${name} must be an array of strings! Is an array, but contains non-string elements!`);
     }
 }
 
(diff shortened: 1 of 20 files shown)


modules/lib/lib-node/src/main/resources/lib/xp/node.ts
```diff
@@ -487,9 +520,11 @@ export interface CommitParams {
     message?: string;
 }
 
+/**
```
rymsha, line 523: it is not deprecated in 7.x

modules/lib/lib-node/src/main/resources/lib/xp/node.ts
```diff
@@ -253,6 +265,9 @@ interface NodeHandler {
 
     setRootPermissions<NodeData>(v: ScriptValue): Node<NodeData>;
 
+    applyPermissions(key: string, permissions: ScriptValue, addPermissions: ScriptValue, removePermissions: ScriptValue, branches: string[],
```
rymsha, line 268: it doe not exist in 7.x

modules/lib/lib-node/src/main/resources/lib/xp/node.ts
```diff
@@ -1003,13 +1042,14 @@ class RepoConnectionImpl
     }
 
     /**
+     * @deprecated
```
rymsha, line 1045: it is not deprecated in 7.x

rymsha (changes requested):

Issue #10755, opened 2024-11-04
ApiDescriptor rename `slash-api` to `mount`


anatol-sialitski:
ApiDescriptor rename `slash-api` to `mount`

PR #10756, opened 2024-11-04
ApiDescriptor rename `slash-api` to `mount` #10755

anatol-sialitski:


rymsha (changes requested):

modules/core/core-api/src/main/java/com/enonic/xp/api/ApiDescriptor.java
```diff
@@ -23,18 +22,19 @@ public final class ApiDescriptor
 
     private final String documentationUrl;
 
-    private final boolean slashApi;
+    private final boolean mount;
 
     private ApiDescriptor( final Builder builder )
     {
         Preconditions.checkNotNull( builder.key, "key cannot be null" );
+        Preconditions.checkNotNull( builder.allowedPrincipals, "allowedPrincipals cannot be null" );
```
rymsha, line 30: empty must not be allowed either

modules/core/core-api/src/main/java/com/enonic/xp/api/ApiDescriptor.java
```diff
@@ -62,15 +62,14 @@ public String getDocumentationUrl()
...
     {
-        return slashApi;
+        return mount;
     }
 
     public boolean isAccessAllowed( final PrincipalKeys principalKeys )
     {
-        return allowedPrincipals == null || principalKeys.contains( RoleKeys.ADMIN ) ||
-            allowedPrincipals.stream().anyMatch( principalKeys::contains );
+        return allowedPrincipals.stream().anyMatch( principalKeys::contains );
```
rymsha, line 72: admin role should always have access, regardless of what descriptor says

rymsha (changes requested):

modules/core/core-api/src/main/java/com/enonic/xp/api/ApiDescriptor.java
anatol-sialitski: Good catch

rymsha (approved):

Issue #10759, opened 2024-11-11
WebSocket: message is sent via non-async RemoteEndpoint


ashklianko:
- Reproduced this problem when multiple threads try sending messages to a client:
"...if two threads attempt to call a send method concurrently, or if a developer attempts to send a new message while in the middle of sending an existing one, the send method called while the connection is already busy may throw an [IllegalStateException](http://docs.oracle.com/javase/7/docs/api/java/lang/IllegalStateException.html?is-external=true)."
- Exception was ignored by empty catch block, thus it was not reflected in server log. 

Probably this issue is one of the reasons of issues with our websockets

PR #10760, opened 2024-11-11
WebSocket: message is sent via non-async RemoteEndpoint #10759

ashklianko:
- Using async RemoteEndpoint for sending a message via Websocket


anatol-sialitski:
Thank you @ashklianko. I will apply it for 7.15. For master it was already done [here](https://github.com/enonic/xp/pull/10763/files#diff-8910f6a76c990f501c4ca5bfe7739872f7e67132a47a4286bbb62dfdb6d58d0d)

Issue #10762, opened 2024-11-13
Migrate Admin Event API to Universal API


anatol-sialitski:
Migrate Admin Event API to Universal API

PR #10763, opened 2024-11-14
Migrate Admin Event API to Universal API #10762

anatol-sialitski:


modules/admin/admin-event/src/main/java/com/enonic/xp/admin/event/impl/EventApiHandler.java
```diff
@@ -0,0 +1,75 @@
...
+        this.webSocketService = webSocketService;
+        this.webSocketManager = webSocketManager;
+    }
+
+    @Override
+    public WebResponse handle( final WebRequest request )
+    {
+        final WebResponse.Builder<?> responseBuilder = WebResponse.create();
+
+        if ( !webSocketService.isUpgradeRequest( request.getRawRequest(), null ) )
```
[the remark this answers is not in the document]
anatol-sialitski: Maybe, but I think it would be better to do it in a separate issue to maintain transparency of changes

modules/admin/admin-event/src/main/java/com/enonic/xp/admin/event/impl/EventApiHandler.java
```diff
@@ -0,0 +1,75 @@
...
+import com.enonic.xp.portal.websocket.WebSocketManager;
+import com.enonic.xp.web.HttpStatus;
+import com.enonic.xp.web.WebRequest;
+import com.enonic.xp.web.WebResponse;
+import com.enonic.xp.web.websocket.WebSocketConfig;
+import com.enonic.xp.web.websocket.WebSocketEvent;
+import com.enonic.xp.web.websocket.WebSocketEventType;
+
+@Component(property = {"applicationKey=admin", "apiKey=event", "displayName=Event API", "allowedPrincipals=role:system.admin.login",
+    "mount=true"})
```
rymsha, line 22:
```suggestion
    "mount=false"})
```

Until this API is stable we don't want to expose it

modules/admin/admin-event/src/main/java/com/enonic/xp/admin/event/impl/EventApiHandler.java
```diff
@@ -0,0 +1,75 @@
...
+public class EventApiHandler
+    implements UniversalApiHandler, EventListener
+{
+    private static final String GROUP_NAME = "com.enonic.xp.admin.event.api";
```
rymsha, line 26:
```suggestion
    private static final String GROUP_NAME = "com.enonic.xp.admin.event";
```

rymsha (approved):

PR #10772, opened 2024-11-27
Fix #10771 PageComponent type is missing template?: string

### LINKED ISSUE #10771: PageComponent type is missing template?: string
When a specific page template is selected, the PageComponent contains a template property with the content id of the page template.

ComLock:


alansemenov (approved):

modules/lib/test/PageComponent.test-d.ts
```diff
@@ -0,0 +1,29 @@
+import type {
+    PageComponent,
+} from '../core/index';
+import {expectAssignable} from 'tsd';
+
+// On a page using automatic page template, the page component is empty.
+expectAssignable<PageComponent>({});
```
ComLock, line 7: One could argue that a page component without type: 'page' is a bug in xp. Cause it means when type is missing on a generic component, I have to assume it's a page.

modules/lib/core/index.d.ts
```diff
@@ -578,11 +578,12 @@ export interface PageComponent<
         Record<string, Region<(FragmentComponent | LayoutComponent | PartComponent | TextComponent)[]>> = 
         Record<string, Region<(FragmentComponent | Layout          | Part          | TextComponent)[]>>
 > {
-    type: 'page'
-    descriptor: Descriptor
-    config: Config
-    path: '/'
-    regions: Regions;
+    type?: 'page'
```
tajakobsen, line 581: What is the background of making every field optional here?

ComLock:
It reflects real data.

modules/lib/core/index.d.ts
ComLock: I can make it better, gimme moment

ComLock: How bout dat?

modules/lib/test/PageComponent.test-d.ts
tajakobsen: > One could argue that a page component without type: 'page' is a bug in xp

I was going to say something similar

If you put yourself in the application-developers shoes; is it useful to know that an empty object can be a `PageComponent`? 

If the app-developer writes a new component-object, do you want to enable it to be without `type`?

ComLock: > If you put yourself in the application-developers shoes; is it useful to know that an empty object can be a PageComponent?
No, but it's useful to know that a PageComponent can be an empty object :)

> If the app-developer writes a new component-object
This typically only happens in the context of testing.

> do you want to enable it to be without type?
I'm not too interested in control, I'm jut trying to represent reality.

ComLock: Seems to work

edloidas (approved):
Great job. Nothing to add, not with my expertise in page components at least.

tajakobsen (approved):
It looks good

Issue #10800, opened 2024-12-10
lib-portal.getContent doesn't resolve page templates


ComLock:
Does not work in site mapping controllers (and probably filters too)

But does work on normal page controllers

ashklianko:
Confirm, MappingHandlerHelper.java doesn't resolve Page object as in for example PageHandlerWorker.java

PR #10807, opened 2024-12-13
lib-portal.getContent doesn't resolve page templates #10800

ashklianko:
- Added Page resolving for the mappers, thus request get content with injected page/regions etc.


ashklianko:
One more question: should we resolve for all RenderMode(s) or for only Edit/Live etc.?

rymsha (approved):

Issue #10794, opened 2024-12-06
Do not remove node and blob versions for non-transient repositories


rymsha:
Do not remove node and blob versions for non-transient repositories

PR #10818, opened 2024-12-18
Don't remove node and blob versions for non-transient repositories #10794

anatol-sialitski:
Do not remove node and blob versions for non-transient repositories #10794

(cherry picked from commit f6a9cc57914fd0e02bf78be9e712fb7b42226f8c)


rymsha:
cherry picked 3361b34006137cc175604fc712d9e9a01f479e92

Issue #8804, opened 2021-05-06
Management port returns 403 when no authentication is provided


gbbirkisson:
403 implies that you are authenticated but not authorized, it should return 401

rymsha:
There is a similar case
https://github.com/enonic/xp/issues/6900

Yes, https://tools.ietf.org/html/rfc2616 got obsolete
so now we should try to follow https://tools.ietf.org/html/rfc7231

That actually ref to
https://tools.ietf.org/html/rfc7231#section-6.5.3 and 
https://tools.ietf.org/html/rfc7235#section-3.1

But 401 MUST send a `WWW-Authenticate` header... sure but it requires some non-trivial code to implement correctly (mind you #8668)
So, what's exactly gets broken for you when 403 is sent instead of 401?

gbbirkisson:
I figured this out when I was trying out reactive authentication in OkHttp:

```
When reactive authentication is requested by an origin web server, the response code is 401 and the implementation should respond with a new request that sets the "Authorization" header
```

PR #10819, opened 2024-12-18
#8804 management port returns 403 when no authentication is provided

rymsha:


modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/idprovider/IdProviderResponseWrapper.java
```diff
@@ -124,43 +129,30 @@ public void sendError( final int sc, final String msg )
     }
 
     private void handleError( final int sc )
+        throws IOException
```
anatol-sialitski, line 132: Why should we handle this in an external method if we could have thrown the  `UncheckedIOException`  error right here?

rymsha: most usages already can handle IOException, only one that doesn't - handled with UncheckedIOException

vbradnitski (approved):

anatol-sialitski (approved):

anatol-sialitski:
Tests?

Issue #10821, opened 2024-12-18
all admin paths should have Cache-control private


rymsha:
To prevent CDNs from caching private pages - Cache-Control must be always private for `/admin/...`

PR #10825, opened 2024-12-20
all admin paths should have Cache-control private #10821

rymsha:


modules/admin/admin-impl/src/main/java/com/enonic/xp/admin/impl/app/NoCacheAdminFilter.java
```diff
@@ -0,0 +1,76 @@
...
+                {
+                    if ( value.contains( "public" ) )
+                    {
+                        value = value.replaceAll( "public", "private" );
+                    }
+                    else
+                    {
+                        value = "private, " + value;
+                    }
+                    value = value.replaceAll( "public", "private" );
```
anatol-sialitski, line 56: We can remove it

anatol-sialitski (approved):

Issue #10848, opened 2025-01-08
Implement `hasUserPassword` function


anatol-sialitski:
Implement the `hasUserPassword` function for the auth library

PR #10849, opened 2025-01-08
Implement hasUserPassword function #10848

anatol-sialitski:


rymsha (changes requested):
it should be property Of User JS object |hasPassword|

No need to modify Java User

Issue #7301, opened 2019-08-15
Page Contributions from components are not injected until page reload


Slideshow776:
# Bug
Contributed style is not included in preview when `apply` is clicked in layout.
Contributed style is included, however, upon a refresh.

Controller code is shown below.
```
var thymeleaf = require("/lib/xp/thymeleaf");
var portal = require("/lib/xp/portal");
var util = require("/lib/enonic/util");
var view = resolve("banner.html");

exports.get = function () {
    var component = portal.getComponent();

    var backgroundClass =  "";   

    if (component.config.bgVariant) { backgroundClass = component.config.bgVariant+"-bg"; }

    var model = {
        mainRegion: component.regions['main'],
        title: component.config.title || '',
        description: component.config.description || '',
        backgroundClass: backgroundClass,
    };

    var cssUrl = portal.assetUrl({ path: "css/extentionsMain.css" });

    return {
        body: thymeleaf.render(view, model),
        pageContributions: {
            headEnd: [
                "<link rel='stylesheet' href='" + cssUrl + "'>",
            ],
        },
    };
};
```

sigdestad:
Contributions are not injected when adding New components to the page editor, it would require a full Page reload

alansemenov:
This is not simply when adding a new component, this is after applying changes in the component which trigger save of the content and reload of the page. Looks like a bug to me.

sigdestad:
Solution to this problem is:
- If a component has pageContributions, we must simply refresh the entire page. If not we keep current behaviour.

alansemenov:
Related task: https://github.com/enonic/app-contentstudio/issues/2034

sigdestad:
It would be really nice to get this fixed, as it is severly affecting usability in the case where a component needs more JS available. This is increasingly relevant with frameworks such as React and friends.

Bellfalasch:
What's the status on this one? When using React we need to refresh the pages since our assets (js-files) changes for each edit inside a part. We serve them using pageContributions.

rymsha:
the idea is to send some info via http headers from renderer in EDIT mode. In this case: inform that component introduces pageContributions.
CS can read custom headers and decide to reload entire page, instead of injecting the component's html.

sigdestad:
Feedback from customers indicates we can often inject javascripts directly within the Page in such cases, and the browser will be able to handle it, Even if the same files already exist

alansemenov:
[...]

PR #10852, opened 2025-01-09
Page Contributions are only injected after page reload #7301

ashklianko:


modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/render/ComponentHandlerWorker.java
```diff
@@ -159,4 +162,8 @@ private Component getFragmentComponent( final FragmentComponent component )
         }
     }
 
+    private PortalResponse addHasContributionsHeader( final PortalResponse source) {
+        return PortalResponse.create(source).header( "Has-Contributions", "true" ).build();
```
rymsha, line 166:
```suggestion
        return PortalResponse.create(source).header( "X-Has-Contributions", "true" ).build();
```

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/render/ComponentHandlerWorker.java
```diff
@@ -103,12 +103,15 @@ else if ( fragmentComponent instanceof LayoutComponent )
             trace.put( "componentPath", component.getPath() );
             trace.put( "type", component.getType().toString() );
         }
-        final PortalResponse response = rendererDelegate.render( component, this.request );
+        final PortalResponse result = rendererDelegate.render( component, this.request );
+        final PortalResponse response = result.hasContributions() ? addHasContributionsHeader( result ) : result;
```
rymsha, line 107: only relevant for Edit mode

rymsha (changes requested):
Shorten Issue title, please

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/render/ComponentHandlerWorker.java
ashklianko: My first thought was also to use `X-` but there's a catch:
https://datatracker.ietf.org/doc/html/rfc6648

rymsha: Then, according to this RFC it should have a prefix `Enonic-`. 
I prefer `X-` though - because static analysis tools are picky and warn about unknown headers, if they don't start with `X-`

ashklianko: Let's stick to `X-` then

rymsha (approved):

PR #10874, opened 2025-01-28
Update lib-event doc

edloidas:
Added explicit info about wildcards in type parameter.


rymsha (changes requested):
It is not "only" that
`Pattern.compile( pattern.replace( ".", "\\." ).replace( "*", ".*" ) );`

rymsha (approved):

PR #10886, opened 2025-02-04
#10885 site edit mode response is 503 instead of 418

rymsha:


anatol-sialitski (approved):

rymsha:
fixed by https://github.com/enonic/xp/pull/10887

Issue #10898, opened 2025-02-12
Wrong default Content-type for json response


rymsha:
```
exports.get = function (req) {
    return {
        body: {a: 1, b: 2}
    }
};
```

should have `application/json; charset=utf-8` in response automatically set

PR #10899, opened 2025-02-12
Wrong default Content-type for json response #10898

rymsha:


modules/portal/portal-impl/src/test/java/com/enonic/xp/portal/impl/controller/AbstractControllerTest.java
```diff
@@ -105,21 +109,41 @@ protected final void execute( final String script )
...
+            try
+            {
+                return MAPPER.writeValueAsString( body );
+            }
+            catch ( final Exception e )
+            {
+                throw new RuntimeException( e );
+            }
+        }
+        return ( body != null ) ? body.toString() : null;
```
anatol-sialitski, line 124:
```suggestion
        return body != null ? body.toString() : null;
```

anatol-sialitski (approved):

PR #10927, opened 2025-02-20
Deprecate dryRun #10926

vbradnitski:


modules/core/core-api/src/main/java/com/enonic/xp/content/ImportContentParams.java
```diff
@@ -153,10 +149,10 @@ public Builder originProject( ProjectName originProject )
             return this;
         }
 
+        @Deprecated
         public Builder dryRun( boolean dryRun )
         {
-            this.dryRun = dryRun;
-            return this;
+            throw new UnsupportedOperationException( "dryRun is not supported" );
```
anatol-sialitski, line 155: Just ignore, otherwise the line https://github.com/enonic/xp/pull/10927/files#diff-1a51303cc84f0d4b00bb82f4871df06a0264fd43616bfce10fadfafdb404f88eR80 does not have sense

rymsha: don't ignore. It is to prevent accidental set of dryRun to true. and still apply changes

rymsha (approved):

Issue #10981, opened 2025-03-26
Portal URL functions for image and attachment


anatol-sialitski:
```
portalLib.imageUrl({
    type: : absolute | server
    id | path : string
    project: string? <- new parameter
    branch: string? <- new parameter
    baseUrl: string? <- new parameter
})
```

PR #10982, opened 2025-03-26
Portal URL functions for image and attachment #10981

anatol-sialitski:


rymsha (changes requested):

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/UrlGeneratorParamsAdapter.java
```diff
@@ -55,120 +47,51 @@ public UrlGeneratorParamsAdapter( @Reference final ContentService contentService
...
-            .setDownload( params.isDownload() )
-            .setName( params.getName() )
-            .setLabel( params.getLabel() )
-            .addQueryParams( params.getParams().asMap() )
-            .build();
+        if ( baseUrl != null )
+        {
+            return noRequestMediaBaseUrlStrategy( baseUrl );
+        }
+        else if ( "/api".equals( portalRequest.getBaseUri() ) )
```
rymsha, line 70: perhaps api baseUri should be `/api/<api-descriptor>` to simplify rewrite logic

rymsha (changes requested):

rymsha (approved):

Issue #10990, opened 2025-03-31
XP8 Tech Stack upgrade


rymsha:
- [x]  Java21

- [x]  Jetty  enonic/xp#10947

- [x]  Hazelcast  enonic/xp#10947

- [x] slf4j, logback  enonic/xp#10947

- [x] Jackson 

- [x] Guava enonic/xp#7414

- [x] Jakarta

- [x] #10997

- [x] Tika

- [x] OSGi

- [x] Resteasy

- [x] OSGi Logging - logback logging was remove from Felix https://github.com/apache/felix-dev/commit/22cdf0e5d03da839e81bffd15194281899d92129

PR #10996, opened 2025-04-01
XP8 Tech Stack upgrade #10990

rymsha:


modules/core/core-internal/src/main/java/com/enonic/xp/core/internal/Exceptions.java
```diff
@@ -0,0 +1,31 @@
...
+    public static void throwIfUnchecked( final Throwable throwable )
+    {
+        if ( throwable instanceof RuntimeException )
+        {
+            throw (RuntimeException) throwable;
+        }
+        if ( throwable instanceof Error )
+        {
+            throw (Error) throwable;
+        }
```
anatol-sialitski, line 20: What about 
```suggestion
        }
        throw new IllegalStateException(throwable);
```

rymsha: This method only throws unchecked exceptions

modules/portal/portal-impl/src/test/java/com/enonic/xp/portal/impl/controller/ControllerScriptImpl_requestTest.java
```diff
@@ -76,12 +75,9 @@ public void testCookiesParsedFromResponse()
         assertEquals( "value2", this.portalResponse.getCookies().get( 1 ).getValue() );
         assertEquals( 2000, this.portalResponse.getCookies().get( 2 ).getMaxAge() );
         assertEquals( "/valid/path", this.portalResponse.getCookies().get( 2 ).getPath() );
-        assertThat( this.portalResponse.getCookies().get( 2 ).getComment() ).contains( "Some cookie comments" );
```
vbradnitski, line 79: comment was replace by attribute? , it would be nice to test it

modules/server/server-status/src/main/java/com/enonic/xp/server/impl/status/MetricsReporter.java
```diff
@@ -26,8 +21,8 @@ public final class MetricsReporter
 
     public MetricsReporter()
     {
-        final MetricsModule module = new MetricsModule( TimeUnit.SECONDS, TimeUnit.SECONDS, false, MetricFilter.ALL );
-        this.mapper = new ObjectMapper().registerModule( module );
+        //final MetricsModule module = new MetricsModule( TimeUnit.SECONDS, TimeUnit.SECONDS, false, MetricFilter.ALL );
+        this.mapper = new ObjectMapper();//.registerModule( module );
```
vbradnitski, line 25: Is it still in work?

vbradnitski (changes requested):

Issue #10611, opened 2024-06-25
processHtml for media


rymsha:
`media://` url post-processing should generate mediaUrl #10610 functions
In first version there is no need to support custom media project for `/_/media`, but `/api/meida` urls should be generated effortlessly

PR #11005, opened 2025-04-04
processHtml for links #10611

anatol-sialitski:


modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/ApiUrlBaseUrlResolver.java
```diff
@@ -19,12 +18,80 @@ final class ApiUrlBaseUrlResolver
 {
     private final ContentService contentService;
 
-    private final ApiUrlGeneratorParams params;
+    private final Supplier<String> application;
```
rymsha, line 21: Can we instead require api descriptor?

rymsha (approved):

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/ApiUrlBaseUrlResolver.java
anatol-sialitski: It will be done in https://github.com/enonic/xp/issues/11007

Issue #10972, opened 2025-03-21
rename should not do update


rymsha:
after the rename - rename command does an extra content update. This is not necessary if new patch method is used

rymsha:
this would also affect layers - verify they still work

PR #11021, opened 2025-04-15
rename should not do update #10972

vbradnitski:


modules/core/core-api/src/main/java/com/enonic/xp/node/AggregatedNodeDataProcessor.java
```diff
@@ -0,0 +1,26 @@
+package com.enonic.xp.node;
+
+import java.util.List;
+
+import com.google.common.base.Preconditions;
+
+import com.enonic.xp.data.PropertyTree;
+
+public final class AggregatedNodeDataProcessor
```
rymsha, line 9:
```suggestion
public final class ComposeteNodeDataProcessor
```

But do we really need it as an API? it should be be trivial to implement whenever needed

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/RenameContentCommand.java
```diff
@@ -47,63 +55,57 @@ Content execute()
...
+                protected EnumSet<ContentInheritType> getTypesToProceed()
+                {
+                    return EnumSet.of( ContentInheritType.NAME );
+                }
+            } );
         }
 
-        final Node node = nodeService.rename( builder.build() );
+        processors.add( ( data, newNodePath ) -> {
+            final Node persistedNode = nodeService.getById( NodeId.from( params.getContentId().toString() ) );
```
rymsha, line 75: back to drawing board....
processor must not fetch node as it operates on not yet persisted data.

rymsha (changes requested):

sigdestad:
Just a note. Remember that diff still needs to detect a rename.

rymsha (approved):

PR #11036, opened 2025-04-23
Epic node upgrading rebased

vbradnitski:


modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/ContentAuditLogSupportImpl.java
```diff
@@ -602,9 +605,24 @@ private void addContents( final PropertySet targetSet, final ContentIds contents
         targetSet.addStrings( name, contents.stream().map( ContentId::toString ).collect( Collectors.toList() ) );
     }
 
-    private void addContents( final PropertySet targetSet, final ContentPaths contents, final String name )
+    private void addCreateAttachments( final PropertySet targetSet, final CreateAttachments attachments )
```
rymsha, line 608: why

modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/node/PatchNodeCommand.java
```diff
@@ -0,0 +1,227 @@
...
+        this.params = builder.params;
+        this.binaryService = builder.binaryService;
+    }
+
+    public static Builder create()
+    {
+        return new Builder();
+    }
+
+    public static Builder create( final AbstractNodeCommand source )
```
rymsha, line 51: is it n neded?

rymsha (changes requested):

Issue #11071, opened 2025-05-13
autologin function cannot return response


rymsha:
`autoLogin?: RequestHandler;` but `autologin` result is ignored

PR #11088, opened 2025-05-20
autologin function cannot return response #11071

edloidas:


rymsha:
I tis not a big deal for 7.15 . In fact it might hurt a bit there - as type checking will insist on `void` return type. Even though we agreed to break types in bugfix versions - I holding the fix till 8.0

PR #11089, opened 2025-05-20
autologin function cannot return response #11071

edloidas:


rymsha (changes requested):
autologin cannot return anything but can accept request

modules/lib/core/index.d.ts
```diff
@@ -449,7 +449,7 @@ export interface ErrorController {
 }
 
 export interface IdProviderController extends Controller {
-    autoLogin?: RequestHandler;
+    autoLogin?: void;
```
alansemenov, line 452: Type cannot be `void`. It should have a special new type where function still accepts `request` but returns `void` instead of `Response`.

modules/lib/test/IdProviderController.test-d.ts
```diff
@@ -12,9 +12,6 @@ const log = {
 const myIdProviderController = {
```
alansemenov, line 12: Here you need to provide correct type for `const myIdProviderController` (`const myIdProviderController: IdProviderController`), then you will immediately see that type of `autoLogin` cannot be just `void`.

alansemenov (changes requested):
Type cannot be `void`. It should have a special new type where function still accepts `request` but returns `void` instead of `Response`.

modules/lib/core/index.d.ts
edloidas: Yeah, was my initial intent, but got distracted and treated autoLogin as response itself.

alansemenov (approved):

rymsha (approved):

Issue #11097, opened 2025-05-23
Universal API should resolve content and site for PortalRequest


rymsha:
legacy service does set Content and Site for PortalRequest - so they are visible in `portal.getContent`
Universal API `/_/` should also support it

sigdestad:
I'm assuming this only applies to the /site engine?

PR #11100, opened 2025-05-23
Universal API should resolve content and site for PortalRequest #11097

anatol-sialitski:


modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/SlashApiHandler.java
```diff
@@ -252,15 +252,18 @@ private boolean verifyPathMountedOnWebapps( final DescriptorKey descriptorKey, f
 
     private boolean verifyRequestMountedOnSites( final DescriptorKey descriptorKey, final PortalRequest portalRequest )
     {
+        final ContentResolverResult contentResolverResult = new ContentResolver( contentService ).resolve( portalRequest );
```
rymsha, line 255: I don't think media handlers ever had access to site/content

rymsha (changes requested):

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/SlashApiHandler.java
anatol-sialitski: This is done in the site context block, which will not be applicable to the slash API part.

rymsha (approved):

PR #11142, opened 2025-06-16
add new repo in matrix

sgauruseu:


.github/workflows/gradle.yml
```diff
@@ -56,7 +56,7 @@ jobs:
     needs: [ publish_vars, build ]
     strategy:
       matrix:
-        repo: ["enonic/app-admin-home", "enonic/app-applications", "enonic/app-users", "enonic/app-standardidprovider", "enonic/app-xp-welcome" ]
+        repo: ["enonic/app-admin-home", "enonic/app-applications", "enonic/app-users", "enonic/app-standardidprovider", "enonic/app-xp-welcome", "enonic/uitest-contenttypes" , "enonic/uitest-myfirstapp", "enonic/uitest-simple-page", "enonic/uitest-myfirstapp", "enonic/uitest-adfs-provider", "enonic/uitest-auth0-provider"]
```
anatol-sialitski, line 59: The `enonic/uitest-myfirstapp` app is used twice.

rymsha:
The system apps are part of XP - hance they are triggered by the build. testing apps are not different from all other (100) apps we manage - and we don't list them here. There should be another way to rebuild the ui-test apps

Issue #10556, opened 2024-05-14
GetCurrentSiteConfigHandler resolve project and branch correctly


anatol-sialitski:

rymsha:
Set it earlier to PortalRequest

PR #11157, opened 2025-06-23

anatol-sialitski:


modules/lib/lib-content/src/main/java/com/enonic/xp/lib/content/mapper/SiteMapper.java
```diff
@@ -1,15 +1,15 @@
...
+import com.enonic.xp.content.Content;
 import com.enonic.xp.script.serializer.MapGenerator;
 import com.enonic.xp.script.serializer.MapSerializable;
-import com.enonic.xp.site.Site;
 
 public final class SiteMapper
     implements MapSerializable
 {
-    private final Site site;
+    private final Content site;
```
rymsha: I don't think this issue is about treating all projects as sites . A bit too early

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/ServiceHandler.java
```diff
@@ -127,10 +132,19 @@ private PortalRequest createPortalRequest( final WebRequest webRequest, final St
...
 
-        final Site site = resolvedContent.getNearestSite();
-
-        //Checks if the application is set on the current site
-        if ( site != null && site.getSiteConfigs().get( descriptorKey.getApplicationKey() ) == null )
+        final Content siteOrProject = resolvedContent.getNearestSite() != null
+            ? resolvedContent.getNearestSite()
+            : ContextBuilder.from( ContextAccessor.current() )
+                .authInfo( AuthenticationInfo.copyOf( ContextAccessor.current().getAuthInfo() )
+                               .principals( RoleKeys.CONTENT_MANAGER_ADMIN )
```
rymsha: Why content manager admin?
Also NOW how do you avoid failure for non site engine requests?

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/BaseUrlExtractor.java
```diff
@@ -61,6 +57,11 @@ BaseUrlMetadata extract( final BaseUrlParams params )
             UrlBuilderHelper.appendSubPath( str, projectName.toString() );
             UrlBuilderHelper.appendSubPath( str, branch.toString() );
 
+            final Content site = portalRequest.getSite();
+            if ( site != null && !site.getPath().isRoot() )
```
rymsha, line 61: It should be possible to have baseUrl for project

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/mapping/MappingHandlerHelper.java
```diff
@@ -114,19 +108,18 @@ public WebResponse handle( final WebRequest webRequest, final WebResponse webRes
...
+        final Content siteOrProject = resolvedContent.getNearestSite() != null
+            ? resolvedContent.getNearestSite()
+            : ContextBuilder.from( ContextAccessor.current() )
+                .repositoryId( request.getRepositoryId() )
+                .branch( request.getBranch() )
+                .authInfo( AuthenticationInfo.copyOf( ContextAccessor.current().getAuthInfo() )
+                               .principals( RoleKeys.CONTENT_MANAGER_ADMIN )
+                               .build() )
+                .build()
+                .callWith( () -> contentService.getByPath( ContentPath.ROOT ) );
```
rymsha, line 120: This should be in ContentResolver

rymsha (changes requested):

modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/BaseUrlExtractor.java
anatol-sialitski: It is done for a project too, but I can improve it.

PR #11161, opened 2025-06-25

anatol-sialitski:


modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/url/ApiUrlBaseUrlResolver.java
```diff
@@ -127,7 +139,7 @@ else if ( PortalRequestHelper.isSiteBase( portalRequest ) )
                 .build();
 
             final Site nearestSite = context.callWith( () -> {
-                final ContentResolver contentResolver = new ContentResolver( contentService );
+                final ContentResolver contentResolver = new ContentResolver( contentService, projectService );
```
rymsha: Is it needed here? At this point site is already set on request

rymsha (approved):

PR #11163, opened 2025-06-26
Don't merge permissions for children nodes

vbradnitski:
`inheritPermissions` field is deprecated and removed from node, so there is no need to go through a tree and merge on SINGLE `applyPermissions` mode


modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/node/ApplyNodePermissionsCommand.java
```diff
@@ -175,6 +175,12 @@ private void doApplyOnNode( final Node node, final Branch branch )
 
         final AccessControlList permissions = mergingStrategy.mergePermissions( node.getPermissions(), params.getPermissions() );
 
+        if ( permissions.equals( node.getPermissions() ) )
+        {
+            results.addResult( node.id(), branch, node );
```
rymsha, line 180: don't we need to add this to "ignored" results and not fire a permissions changed event ? (because in reality no new version is created)

Also, how does it fix the problem with permissions being applied on children?

PR #11165, opened 2025-06-26

anatol-sialitski:


modules/portal/portal-api/src/main/java/com/enonic/xp/portal/handler/BaseSiteHandler.java
```diff
@@ -81,7 +82,8 @@ private static String findPathAfterBranch( final String baseSubPath )
         return index >= 0 ? repoSubPath.substring( index ) : "";
     }
 
-    protected PortalRequest doCreatePortalRequest( final WebRequest webRequest, final String baseUri, final String baseSubPath )
+    protected PortalRequest doCreatePortalRequest( final WebRequest webRequest, final String baseUri, final String baseSubPath,
```
rymsha: Remove this class. There are only 2 usages and they should have their own url parsing logic

rymsha (changes requested):

rymsha (approved):

PR #11212, opened 2025-07-25
Remove FullContentSerializer functionality #11211

vbradnitski:


modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/ContentServiceImpl.java
```diff
@@ -167,7 +166,7 @@ public ContentServiceImpl( @Reference final NodeService nodeService, @Reference
         this.pageDescriptorService = pageDescriptorService;
```
rymsha, line 166: Remove services

vbradnitski: they are in use by CreateSiteCommand

Issue #11213, opened 2025-07-25
Use new collector methods


rymsha:
Continue on #10429

PR #11214, opened 2025-07-25
Use new collector methods #11213

rymsha:


modules/core/core-api/src/main/java/com/enonic/xp/blob/BlobKeys.java
```diff
@@ -30,9 +29,9 @@ public static BlobKeys from( final BlobKey... blobKeys )
         return fromInternal( ImmutableSet.copyOf( blobKeys ) );
     }
 
-    public static BlobKeys from( final Collection<BlobKey> blobKeys )
+    public static BlobKeys from( final Iterable<? extends BlobKey> blobKeys )
```
vbradnitski: isn't BlobKey final?

modules/core/core-api/src/main/java/com/enonic/xp/content/CompareContentResults.java
```diff
@@ -13,11 +11,11 @@
 public final class CompareContentResults
     implements Iterable<CompareContentResult>
 {
-    private final Map<ContentId, CompareContentResult> compareContentResultsMap;
+    private final ImmutableMap<ContentId, CompareContentResult> compareContentResultsMap;
 
-    private CompareContentResults( Builder builder )
+    private CompareContentResults( ImmutableMap<ContentId, CompareContentResult> builder )
```
vbradnitski, line 16: Do we want to use 3d part specific lib class in a constructor of core-api class?

modules/core/core-api/src/main/java/com/enonic/xp/macro/MacroDescriptors.java
```diff
@@ -30,9 +29,9 @@ public static MacroDescriptors from( final MacroDescriptor... macroDescriptors )
         return fromInternal( ImmutableList.copyOf( macroDescriptors ) );
     }
 
-    public static MacroDescriptors from( final Collection<MacroDescriptor> macroDescriptors )
+    public static MacroDescriptors from( final Iterable<? extends MacroDescriptor> macroDescriptors )
```
vbradnitski: MacroDescriptor is final

modules/core/core-api/src/main/java/com/enonic/xp/content/CompareContentResults.java
rymsha: for private constructor - yes

vbradnitski (approved):

Issue #8700, opened 2021-03-11
remove XData from ContentType


rymsha:
XData on ContentType is a rudiment from XP6 - not mentioned in documentation

rymsha:
also deprecate related ContentTypeFilter class

PR #11221, opened 2025-07-30
remove XData from ContentType #8700

rymsha:


modules/core/core-api/src/main/java/com/enonic/xp/attachment/CreateAttachments.java
```diff
@@ -51,23 +59,31 @@ public static Builder create()
...
+            map.put(  value.getName(), value );
             return this;
         }
 
         public Builder addAll( final Iterable<CreateAttachment> values )
         {
-            contents.addAll( values );
+            for ( CreateAttachment value : values )
+            {
+                map.put( value.getName(), value );
```
vbradnitski, line 77: didn't you want to deduplicate values by 'name+label'?

modules/core/core-api/src/main/java/com/enonic/xp/content/ExtraDatas.java
```diff
@@ -14,11 +17,14 @@
 public final class ExtraDatas
     extends AbstractImmutableEntityList<ExtraData>
 {
-    private static final ExtraDatas EMPTY = new ExtraDatas( ImmutableList.of() );
+    private static final ExtraDatas EMPTY = new ExtraDatas( ImmutableMap.of() );
 
-    private ExtraDatas( final ImmutableList<ExtraData> list )
+    private final ImmutableMap<XDataName, ExtraData> map;
```
vbradnitski, line 21: do we need a mep only for 'getMetadata' method?

rymsha: Mostly used to deductate. But since we have this map - why not using for quick lookup.
Although existence of this getMetadata is doubtful.

Issue #11236, opened 2025-08-04
Form API cleanup


rymsha:
Certain aspects of Form in schema are suboptimal for performance
 
- Form can be immutable  - this will avoid copying when schema is used
- Mixins can be mixed in earlier to avoid to avoid form traversal when property sets are validated
- .,.

PR #11237, opened 2025-08-04
Form API cleanup #11236

rymsha:


anatol-sialitski (approved):

modules/core/core-api/src/main/java/com/enonic/xp/form/FieldSet.java
```diff
@@ -23,18 +27,26 @@ public final class FieldSet
...
+    @Override
+    public String getName()
+    {
+        return this.name;
+    }
+
+    @Override
+    public FormItemType getType()
+    {
+        return FormItemType.LAYOUT;
```
vbradnitski, line 49: backward compatibility?

rymsha: a bit. may change in next iteration

PR #11306, opened 2025-08-27
Improving InputType ContentSelector from YAML #11305

### LINKED ISSUE #11305: Improving InputType ContentSelector from YAML
Improving InputType ContentSelector from YAML

It should be possible to use either `allowContentType` or `allowType`, where `allowType` serves as an alias for `allowContentType`.

anatol-sialitski:


modules/core/core-schema/src/main/java/com/enonic/xp/core/impl/schema/mapper/ContentSelectorYml.java
```diff
@@ -10,6 +13,8 @@
 public class ContentSelectorYml
     extends InputYml
 {
+    @JsonAlias("allowType")
```
rymsha: I see no point from supporting both `allowType` and `allowContentType`

rymsha (changes requested):

rymsha (approved):

PR #11315, opened 2025-08-28
Support TextArea in YAML #11314

anatol-sialitski:


modules/core/core-schema/src/main/java/com/enonic/xp/core/impl/schema/mapper/InputYml.java
```diff
@@ -36,6 +43,12 @@ public final Input convertToInput()
             builder.occurrences( occurrences );
         }
 
+        if ( defaultValue != null )
```
rymsha, line 46: Not all default values are strings

anatol-sialitski: Sure, that is why the default has the type Object. But in the input property, all values can only be strings

rymsha (approved):

Issue #11304, opened 2025-08-27
Remove setActiveVersion method


rymsha:
setActiveVersion is harmful and must be removed. New version should be generated instead

PR #11316, opened 2025-08-29
Remove setActiveVersion method #11304

rymsha:


modules/core/core-repo/src/main/java/com/enonic/xp/repo/impl/storage/StoreNodeParams.java
```diff
@@ -47,6 +47,10 @@ public static Builder create()
         return new Builder();
     }
 
+    public static StoreNodeParams newVersion( final Node node )
```
anatol-sialitski, line 50: We usually use `from`

anatol-sialitski (approved):

Issue #11307, opened 2025-08-27
sort for childOrder


rymsha:
setChildOrder and reorderChildren should be just sort.

current API generates too many versions for manual sorting.

PR #11325, opened 2025-08-29
Child order #11307

rymsha:


modules/core/core-api/src/main/java/com/enonic/xp/blob/NodeVersionKey.java
```diff
@@ -12,9 +12,9 @@ public final class NodeVersionKey
 
     private NodeVersionKey( final Builder builder )
     {
-        nodeBlobKey = builder.nodeBlobKey;
-        indexConfigBlobKey = builder.indexConfigBlobKey;
-        accessControlBlobKey = builder.accessControlBlobKey;
+        nodeBlobKey = Objects.requireNonNull(builder.nodeBlobKey);
```
anatol-sialitski, line 15: Should we keep the error message?

modules/core/core-api/src/main/java/com/enonic/xp/content/SortContentParams.java
```diff
@@ -0,0 +1,115 @@
...
+    {
+        return manualOrderSeed;
+    }
+
+    public List<ReorderChildContentParams> getReorderChildContents()
+    {
+        return reorderChildContents;
+    }
+
+    public boolean stopInherit()
```
anatol-sialitski, line 52:
```suggestion
    public boolean isStopInherit()
```

modules/core/core-api/src/main/java/com/enonic/xp/blob/NodeVersionKey.java
rymsha: Too intrrnal really to give any value

modules/core/core-api/src/main/java/com/enonic/xp/content/SortContentParams.java
rymsha: Not sure. @vbradnitski call

vbradnitski (approved):

Issue #11346, opened 2025-09-03
Support `IdproviderDescriptor` using YAML


anatol-sialitski:
```
mode: "MIXED"
form:
  - type: "TextLine"
    name: "title"
    label: "Title"
    occurrences:
      minimum: 0
      maximum: 1
```

PR #11352, opened 2025-09-04
Support `IdproviderDescriptor` using YAML #11346

anatol-sialitski:


modules/core/core-schema/src/main/java/com/enonic/xp/core/impl/schema/YmlParserBase.java
```diff
@@ -40,6 +41,11 @@ public final class YmlParserBase
     private final ObjectMapper mapper = new ObjectMapper( new YAMLFactory() );
 
     public YmlParserBase()
+    {
+        initialize();
```
rymsha: there is no point to call private method from constructor.

PR #11355, opened 2025-09-04
Support `WebappDescriptor` using YAML #11349

anatol-sialitski:


modules/core/core-schema/src/main/java/com/enonic/xp/core/impl/schema/mapper/DescriptorKeyDeserializer.java
```diff
@@ -21,15 +21,15 @@ public DescriptorKey deserialize( final JsonParser jsonParser, final Deserializa
         final ObjectMapper mapper = (ObjectMapper) jsonParser.getCodec();
         final JsonNode node = mapper.readTree( jsonParser );
 
-        final String rawValue = node.asText();
-        if ( rawValue.contains( ":" ) )
+        final String trimmedValue = node.asText().trim();
```
rymsha: I see no pint in yaml/json to trim values. it might have had some explanation n XML only

modules/core/core-schema/src/main/java/com/enonic/xp/core/impl/schema/mapper/PrincipalKeyMapper.java
```diff
@@ -4,11 +4,13 @@
...
 
+import static com.google.common.base.Strings.nullToEmpty;
+
 public class PrincipalKeyMapper
 {
     @JsonCreator
     public static PrincipalKey from( String value )
     {
-        return PrincipalKey.from( value );
+        return PrincipalKey.from( nullToEmpty( value ).trim() );
```
rymsha: see above

rymsha (changes requested):

Issue #11281, opened 2025-08-20
set content admin access to 'patch' method


vbradnitski:
Only users with content admin access should be allowed to call publish method

PR #11382, opened 2025-09-11
set content admin access to 'patch' method #11281

vbradnitski:


modules/core/core-api/src/main/java/com/enonic/xp/project/ProjectAccessVerifier.java
```diff
@@ -0,0 +1,10 @@
+package com.enonic.xp.project;
+
+import java.util.Set;
+
+import com.enonic.xp.security.auth.AuthenticationInfo;
+
+public interface ProjectAccessVerifier
```
rymsha: Need more discussion

Issue #11219, opened 2025-07-28
`applyPermissions` return permissions in result


vbradnitski:
ContentMapper in JS content lib doesn't map `permissions` from Java object to JS. So a result of `applyPermissions` is useless in JS.
We should either return just permissions instead of full Content ( together with Java API) or extend JS result with Permissions.

PR #11395, opened 2025-09-18
applyPermissions js API doesn't return permissions in result #11219

vbradnitski:


modules/lib/lib-content/src/main/resources/lib/xp/content.ts
```diff
@@ -16,7 +16,7 @@ declare global {
     interface XpXData {}
 }
 
-import type {
+import {
```
rymsha: why?

modules/core/core-api/src/main/java/com/enonic/xp/content/ApplyContentPermissionsResult.java
```diff
@@ -54,9 +54,9 @@ private Builder()
         {
         }
 
-        public Builder addResult( ContentId contentId, Branch branch, Content content )
+        public Builder addResult( ContentId contentId, Branch branch, AccessControlList permissions )
         {
-            results.put( contentId, new BranchResult( branch, content ) );
+            results.put( contentId, new BranchResult( branch, permissions ) );
```
rymsha, line 59: Aren't permissions supposed to be equal in all branches? Except for the case where permissions were not applied

rymsha (approved):

mkansara23:
Hi @vbradnitski , 

I noticed 6 rename refactoring in your [recent commit](https://github.com/enonic/xp/commit/5b9939a4033bdb61f2469dc195453f6f7c25aff7):

- Single Variable Declaration content -> permissions.
- Field Declaration content -> permissions.
- Method Declaration addContentWithPermissions -> addPermissions.

I had a few questions on your approach to these refactorings:
- How did you identify that these six elements were connected and should be renamed together?
- Did you find this cluster-rename challenging in any way?
- Did you perform the renames manually or with tools?
- Would it be useful to have tools that automatically identify and rename clusters in one step?

Issue #10650, opened 2024-08-15
Move `parsers` from `core-api`


anatol-sialitski:
Move `parsers` from `core-api`

rymsha:
We are considering to remove xml - and replace them with yaml syntax

PR #11429, opened 2025-10-02
Remove XML parsers from core-api #10650

anatol-sialitski:


modules/core/core-api/src/main/java/com/enonic/xp/site/CmsDescriptor.java
```diff
@@ -0,0 +1,118 @@
...
+    {
+        return xDataMappings;
+    }
+
+    public Instant getModifiedTime()
+    {
+        return modifiedTime;
+    }
+
+    public static ResourceKey toResourceKey( final ApplicationKey applicationKey )
```
rymsha, line 52: I think it is a misconception in API that descriptors are always resources. This is an implementation detail and should be moved to impl

modules/core/core-api/src/main/java/com/enonic/xp/site/CmsDescriptor.java
```diff
@@ -0,0 +1,118 @@
...
+    {
+        return ResourceKey.from( applicationKey, CMS_DESCRIPTOR_PATH );
+    }
+
+    public static Builder create()
+    {
+        return new Builder();
+    }
+
+    public static Builder copyOf( final CmsDescriptor cmsDescriptor )
```
rymsha, line 62: copyOf is only needed because we have form-fragments. Maybe ove this logic where it belongs?

Issue #10920, opened 2025-02-18
content.update - allow to update siteConfig


rymsha:
Fields with special behavior 
`siteConfig`

PR #11430, opened 2025-10-02
content.update - allow to update siteConfig #10920

vbradnitski:


modules/core/core-api/src/main/java/com/enonic/xp/site/SiteConfigsDataSerializer.java
```diff
@@ -2,35 +2,44 @@
...
         {
-            final SiteConfig siteConfig = siteConfigSerializer.fromData( siteConfigAsProperty.getSet() );
+            final SiteConfig siteConfig = getConfig( siteConfigAsProperty.getSet() );
             builder.add( siteConfig );
         }
-        return builder;
+        return builder.build();
+    }
+
+    public SiteConfig getConfig( final PropertySet siteConfigAsSet )
```
rymsha: Maybe all methods should be just static?

getConfig is it a good name?

rymsha (changes requested):
validation on role with non-empty site config is the main part of requested changes. I haven't found it in changes

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/CreateContentCommand.java
```diff
@@ -153,6 +153,12 @@ private void validatePropertyTree( final CreateContentParams params )
         {
             final ContentType contentType = contentTypeService.getByName( new GetContentTypeParams().contentTypeName( params.getType() ) );
 
+//            if ( contentType.getName().isSite() )
```
rymsha: check project access

rymsha (changes requested):

PR #11468, opened 2025-10-10
X-data is not added to content when it's created #9952

vbradnitski:


modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/XDataMappingServiceImpl.java
```diff
@@ -0,0 +1,160 @@
...
+                    else
+                    {
+                        return false;
+                    }
+
+                } );
+            }
+        } );
+
+        return result.entrySet().stream().map( e -> new XDataOption( e.getKey(), e.getValue() ) ).collect( toList() );
```
rymsha:
```suggestion
        return result.entrySet().stream().map( e -> new XDataOption( e.getKey(), e.getValue() ) ).toList() ;
```

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/XDataMappingServiceImpl.java
```diff
@@ -0,0 +1,160 @@
...
+        final Map<XData, Boolean> result = new LinkedHashMap<>();
+
+        xDataMappings.stream().filter( xDataMapping -> {
+            final String pattern = xDataMapping.getAllowContentTypes();
+            final ApplicationKey applicationKey = xDataMapping.getXDataName().getApplicationKey();
+
+            return nullToEmpty( pattern ).isBlank() || new ApplicationWildcardMatcher<>( applicationKey, ContentTypeName::toString,
+                                                                                         ApplicationWildcardMatcher.Mode.MATCH ).matches(
+                pattern, contentTypeName );
+        } ).forEach( xDataMapping -> {
```
rymsha, line 148: No foreach for huge lambdas

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/CreateContentCommand.java
```diff
@@ -75,6 +91,49 @@ Content execute()
         return doExecute();
     }
 
+    private ExtraDatas mergeExtraData()
+    {
+        final ExtraDatas.Builder result = ExtraDatas.create();
+
+        final List<XDataOption> allowedXData = xDataMappingService.fetch( params.getParent(), params.getType() );
```
rymsha: If site is created, XData Mapping happens from self, not from parent

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/CreateContentCommand.java
```diff
@@ -75,6 +91,49 @@ Content execute()
...
+        {
+            final boolean isOptional = xDataOption.optional();
+            final XData xData = xDataOption.xdata();
+            final ExtraData extraData = params.getExtraDatas().getMetadata( xData.getName() );
+
+            if ( extraData == null )
+            {
+                if ( !isOptional )
+                {
+                    throw new IllegalArgumentException( "Missing xData: " + xData.getName() );
```
rymsha: We allow to save invalid data, but it is valid = false

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/XDataMappingServiceImpl.java
```diff
@@ -0,0 +1,160 @@
...
+
+        this.contentTypeService = contentTypeService;
+        this.eventPublisher = eventPublisher;
+        this.translator = new ContentNodeTranslator();
+    }
+
+    @Override
+    public List<XDataOption> fetch( final ContentPath parentPath, final ContentTypeName type )
+    {
+        final Node parent = this.nodeService.getByPath( ContentNodeHelper.translateContentPathToNodePath( parentPath ) );
```
rymsha: Nearest site, as well as entire xdata mapping is not relevant it archive.

There is a way to find Nearest Site by path. It is more efficient that to get node and then get path

modules/core/core-api/src/main/java/com/enonic/xp/site/XDataMappingService.java
```diff
@@ -0,0 +1,13 @@
+package com.enonic.xp.site;
+
+import java.util.List;
+
+import com.enonic.xp.content.ContentPath;
+import com.enonic.xp.schema.content.ContentTypeName;
+
+public interface XDataMappingService
+{
+    List<XDataOption> fetch( ContentPath path, ContentTypeName type );
```
rymsha, line 10: single method
return `XDataOptions`

Issue #11417, opened 2025-09-26
Split DateTime input into two input types


rymsha:
DateTime sometimes represents Temporal Instant and sometimes Local DateTime value.
These are two different data types and and must be represented by two different input types

PR #11469, opened 2025-10-10
Split DateTime input into two #11417

anatol-sialitski:


modules/core/core-api/src/main/java/com/enonic/xp/inputtype/DateTimeHelper.java
```diff
@@ -0,0 +1,32 @@
+package com.enonic.xp.inputtype;
+
+import java.time.Instant;
+import java.time.Period;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+
+final class DateTimeHelper
```
rymsha: Just remove this code. It requires offset from client (browser)

modules/core/core-api/src/main/java/com/enonic/xp/inputtype/InputTypeName.java
```diff
@@ -22,6 +22,8 @@ public final class InputTypeName
 
     public static final InputTypeName DATE_TIME = InputTypeName.from( "DateTime" );
 
+    public static final InputTypeName LOCAL_DATE_TIME = InputTypeName.from( "LocalDateTime" );
```
rymsha: DateTime. And another is Instant

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/processor/ImageContentProcessor.java
```diff
@@ -364,7 +363,7 @@ private ExtraDatas extractMetadata( final MediaInfo mediaInfo )
                 if ( formItem instanceof Input input )
                 {
                     final ExtraData extraData = getOrCreate( metadataMap, xData.getName() );
-                    if ( InputTypeName.DATE_TIME.equals( input.getInputType() ) )
+                    if ( InputTypeName.INSTANT.equals( input.getInputType() ) )
```
rymsha:
```suggestion
                    if ( InputTypeName.DATE_TIME.equals( input.getInputType() ) )
```

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/schema/BuiltinXDataTypes.java
```diff
@@ -90,7 +90,7 @@ private static Input.Builder createLong( final String name, final String label,
 
     private static Input.Builder createDate( final String name, final String label, final String i18n )
     {
-        return Input.create().inputType( InputTypeName.DATE_TIME ).label( label ).name( name ).
+        return Input.create().inputType( InputTypeName.INSTANT ).label( label ).name( name ).
```
rymsha:
```suggestion
        return Input.create().inputType( InputTypeName.DATA_TIME ).label( label ).name( name ).
```

Issue #11495, opened 2025-10-23
Replace InputTypeConfig with GenericValue


anatol-sialitski:
Replace InputTypeConfig with GenericValue

PR #11500, opened 2025-10-27
Replace InputTypeConfig with GenericValue #11495

anatol-sialitski:


modules/core/core-api/src/main/java/com/enonic/xp/util/GenericValue.java
```diff
@@ -0,0 +1,250 @@
...
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+
+public final class GenericValue
```
rymsha, line 18: move to `com.enonic.xp.util` package

Issue #11523, opened 2025-11-07
Fix OSGi dependencies


anatol-sialitski:
Fix OSGi dependencies.

PR #11524, opened 2025-11-07
Fix OSGi dependencies #11523

anatol-sialitski:


modules/core/core-content/build.gradle
```diff
@@ -12,6 +12,7 @@ dependencies {
 
 jar {
     bundle {
-        bnd( 'Bundle-Name': 'Enonic XP Core - Content' )
+        bnd( 'Bundle-Name': 'Enonic XP Core - Content',
+             'Export-Package': 'com.enonic.xp.core.impl.content.parser')
```
rymsha, line 16: What is it for?

anatol-sialitski: `DynamicResourceParser` from the `core-app` module uses parsers from the core-content module, such as `Page`, `Part`, `Layout` , `Content` ...

PR #11566, opened 2025-12-10
test `allTextLanguages` js property

vbradnitski:


vbradnitski:
@copilot add tests for the new property and validate the name on the index codebase

Issue #11521, opened 2025-11-05
Full repository exports


rymsha:
Export loads all nodes into heap - causing OOME when there are many nodes.

related to #8029

rymsha:
it might be an alternative to #11492

PR #11677, opened 2026-01-16
Full repository exports #11521

vbradnitski:


modules/core/core-export/src/main/java/com/enonic/xp/core/impl/export/NodeExporter.java
```diff
@@ -124,38 +132,45 @@ private void doWriteNode( final Node node, final Path baseFolder )
...
-
-    private FindNodesByParentResult doExport( final NodePath nodePath )
-    {
-        final FindNodesByParentResult children =
-            nodeService.findByParent( FindNodesByParentParams.create().parentPath( nodePath ).build() );
+        int from = 0;
+        boolean hasMore = true;
 
-        final Nodes childNodes = this.nodeService.getByIds( children.getNodeIds() );
+        final String childrenPattern = parentPath.isRoot() ? "/?*" : parentPath + "/*";
```
vbradnitski, line 138: fake news

Issue #10473, opened 2024-03-20
Different error message when importing content that is in archive


danhul:
I worked on a support issue to Brreg to import a layer from one environment to another. The import failed for some nodes and the error message was "Node already exists". The content on the destination environment was cleaned before the import but the failure was due to contents that were archived. 

Can the error message be updated to inform users that the contents exists in "archive" when that is the case?

PR #11703, opened 2026-01-26
Different error message when importing content that is in archive #10473

vbradnitski:


modules/core/core-api/src/main/java/com/enonic/xp/node/NodeIdExistsException.java
```diff
@@ -1,13 +1,66 @@
 package com.enonic.xp.node;
 
 import com.enonic.xp.annotation.PublicApi;
+import com.enonic.xp.branch.Branch;
+import com.enonic.xp.repository.RepositoryId;
 
 @PublicApi
 public class NodeIdExistsException
     extends RuntimeException
```
rymsha:
```suggestion
    extends DuplicateElementException
```

Issue #8815, opened 2021-05-12
Improve error message when non-content node is fetched by id


rymsha:
Improve error message when content.getById called on a non-content object (for instance issues node) 

Currently error message it too abstract
```
java.lang.RuntimeException: Error getting node
Caused by: java.lang.NullPointerException
at com.enonic.xp.core.impl.content.serializer.ContentDataSerializer.fromData(ContentDataSerializer.java:216)
at com.enonic.xp.core.impl.content.ContentNodeTranslator.doTranslate(ContentNodeTranslator.java:77)
at com.enonic.xp.core.impl.content.ContentNodeTranslator.fromNode(ContentNodeTranslator.java:49)
at com.enonic.xp.core.impl.content.GetContentByIdCommand.execute(GetContentByIdCommand.java:32)
... 39 more
```

PR #11704, opened 2026-01-27
Improve error message when non-content node is fetched by id

vbradnitski:
#8815


modules/core/core-content/src/test/java/com/enonic/xp/core/impl/content/CreateContentCommandTest.java
```diff
@@ -379,8 +379,7 @@ void createTemplateFolderOutsideSite()
         final Node parentNode = Node.create()
             .id( NodeId.from( "id1" ) )
             .name( "parent" )
-            .parentPath( ContentConstants.CONTENT_ROOT_PATH )
-            .data( parentNodeData )
+            .parentPath( ContentConstants.CONTENT_ROOT_PATH ).data( parentNodeData ).nodeType( ContentConstants.CONTENT_NODE_COLLECTION )
```
rymsha: in reality content root has `default`type

Issue #11701, opened 2026-01-26
NPE during site validation if layout/part components are not selected in dropdown


sgauruseu:
1. Open a site , insert a component , part or layout
2. Click on `Save` button

**BUG** - NPE after saving the site.

<img width="1853" height="949" alt="Image" src="https://github.com/user-attachments/assets/222b3812-1248-43a2-8cc2-b31c3791f881" />

```

java.lang.NullPointerException: key is required
        at java.base/java.util.Objects.requireNonNull(Objects.java:246)
        at com.enonic.xp.resource.ResourceProcessor$Builder.build(ResourceProcessor.java:86)
        at com.enonic.xp.core.impl.app.descriptor.DescriptorFacetImpl.newProcessor(DescriptorFacetImpl.java:97)
        at com.enonic.xp.core.impl.app.descriptor.DescriptorFacetImpl.get(DescriptorFacetImpl.java:42)
        at com.enonic.xp.core.impl.app.descriptor.DescriptorServiceImpl.get(DescriptorServiceImpl.java:37)
        at com.enonic.xp.core.impl.content.page.region.PartDescriptorServiceImpl.getByKey(PartDescriptorServiceImpl.java:30)
        at com.enonic.xp.core.impl.content.AbstractCreatingOrUpdatingContentCommand.validateRegions(AbstractCreatingOrUpdatingContentCommand.java:257)
        at com.enonic.xp.core.impl.content.AbstractCreatingOrUpdatingContentCommand.validatePage(AbstractCreatingOrUpdatingContentCommand.java:245)
        at com.enonic.xp.core.impl.content.UpdateContentCommand.validate(UpdateContentCommand.java:241)
        at com.enonic.xp.core.impl.content.UpdateContentCommand.doExecute(UpdateContentCommand.java:93)
        at com.enonic.xp.core.impl.content.UpdateContentCommand.execute(UpdateContentCommand.java:64)
        at com.enonic.xp.core.impl.content.ContentServiceImpl.update(ContentServiceImpl.java:275)
        at com.enonic.app.contentstudio.rest.resource.content.page.PageResource.update(PageResource.java:57)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
        at java.base/java.lang.reflect.Method.invoke(Method.java:565)
        at org.jboss.resteasy.core.MethodInjectorImpl.invoke(MethodInjectorImpl.java:154)
[...]

PR #11710, opened 2026-01-29
#11701 NPE during site validation if layout/part components are not selected

vbradnitski:
… in dropdown #11701


modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/AbstractCreatingOrUpdatingContentCommand.java
```diff
@@ -252,15 +253,21 @@ private void validateRegions( final Regions regions )
         {
             for ( Component component : region.getComponents() )
             {
-                if ( component instanceof PartComponent part )
+                if ( !( component instanceof DescriptorBasedComponent descriptorBasedComponent ) ||
+                    !descriptorBasedComponent.hasDescriptor() )
+                {
+                    continue;
```
rymsha, line 259: is it a root cause? isn't it strange that descriptor-based component doesn't have a descriptor? Maybe it should not have been created like that at first place and actually an error should be more descriptive, instead of just allowing it and not validating

alansemenov: no, it's a normal use case to add one or several components and save the content without selecting descriptors

Issue #11732, opened 2026-02-09
publish.time


rymsha:
`publish.time` that is set to current time on publish and reset on every update

if content is modified in-place ( both in master and draft) value is not modified (kept as-is, it might be null in draft, but has some value in master)

rymsha:
workflow appears to be a bad fit 
- workflow is used for layers sync and any additional state on top of ready can brake the logic
- workflow can be changed via updateWorkflow to any state. 

added an option in #11734 fix - instead of workflow use publish data `publish.published`. it fits perfectly and does not break layers sync

PR #11741, opened 2026-02-12
#11732 new workflow published

rymsha:


modules/core/core-api/src/main/java/com/enonic/xp/content/GetActiveContentVersionsResult.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.Map;
+
+import com.google.common.collect.ImmutableMap;
+
+import com.enonic.xp.annotation.PublicApi;
+import com.enonic.xp.branch.Branch;
+
+@PublicApi
+public final class GetActiveContentVersionsResult
```
rymsha:
```suggestion
public record GetActiveContentVersionsResult
```

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/MoveContentCommand.java
```diff
@@ -90,6 +90,7 @@ private MoveContentsResult doExecute()
             if ( !layersSync )
             {
                 processors.add( InheritedContentDataProcessor.PARENT );
+                processors.add( PublishDataProcessor::removePublishTime );
```
vbradnitski, line 93: why only for "!layersSync" ?

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/SortContentCommand.java
```diff
@@ -59,7 +59,10 @@ SortContentResult execute()
             }
             else
             {
-                paramsBuilder.processor( InheritedContentDataProcessor.SORT );
+                paramsBuilder.processor(
+                    CompositeNodeDataProcessor.create().add( InheritedContentDataProcessor.SORT ).add(
```
vbradnitski, line 63: Why is sorting changing "publishTime"?

rymsha: to show that after sort you need to publish again

modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/MoveContentCommand.java
rymsha: i dont thik it is strictly necessary because layers sync removes entire publish field, but didnt bother checking. it is sort of a safeguard

Issue #11750, opened 2026-02-14
Switch from NPM_TOKEN to OIDC Trusted Publishers


edloidas:
NPM says: 
> Classic tokens have been revoked. Granular tokens are now limited to 90 days and require 2FA by default. Update your CI/CD workflows
  to avoid disruption. 

Our old way of publishing npm packages no longer works. We need to migrate from classic tokens to trusted publishes via GitHub's OIDC identity, like was already done in miltiple other repos.

PR #11751, opened 2026-02-14
Switch from NPM_TOKEN to OIDC Trusted Publishing #11750

edloidas:
Moved from classic token for npm publishing to Trusted publishing. 
Added permissions for "contents: write" (not "read" to guarantee artifacts will be saved, had issues with that in multiple repos) and "id-token: write" (for publishing). 
Upgraded Node (and bundled npm), since Trusted publishing requires npm >= 11.5.1

_All packages on npm were also configured to use Trusted publishing, so this should work now._

---

If 7.16.2 wasn't actually released, I'd suggest rolling back last 2 commit in 7.16, dropping 7.16.2 tag, merge these changes and repeat.
Also, `NPM_TOKEN` can be removed for this repo.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,6 +40,10 @@ jobs:
 
     needs: release_notes
 
+    permissions:
+      contents: write
+      id-token: write
+
     steps:
       - uses: actions/checkout@v5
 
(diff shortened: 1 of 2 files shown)


jsi (approved):
Let's try this with and RC-release

Issue #11764, opened 2026-02-16
Missing tag argument when publishing with npm


edloidas:
npm version 11+ requires passing `--tag` argument when publishin prerelease versions. Right now we are missing this functionality in our Gradle script.

PR #11765, opened 2026-02-16
Missing tag argument when publishing with npm #11764

edloidas:


modules/lib/build.gradle
```diff
@@ -10,6 +10,10 @@ node {
 
 def coreDir = "$projectDir/core/"
 
+def versionString = version.toString()
```
alansemenov, line 13: Not sure if this is needed - version is already a string

modules/lib/build.gradle
```diff
@@ -10,6 +10,10 @@ node {
...
+def npmPublishTag = versionString.contains( '-' ) && !versionString.endsWith( '-SNAPSHOT' ) ? 'next' : 'latest'
```
alansemenov, line 14: This will set the value to `latest` for a snapshot version. A bit scary :) but I guess it's ok since we have `onlyIf` checks inside the actual publishing tasks

edloidas: We use `onlyIf` guard and don't publish SNAPSHOTS.

Issue #11773, opened 2026-02-17
NPE in ScriptValueTranslator.handleElement() on Nashorn


edloidas:
`ScriptValueTranslator.handleElement()` has no `null` check. Its final else branch calls `value.toString()`, which throws `NullPointerException` when a null value reaches it.                                                                                                                                                              
                                                                                                                                                                    
In practice, GraalJS's `GraalObjectConverter.toMap()` filters out `null` values before they reach `ScriptValueTranslator`, masking the bug. However, **Nashorn's** `JsObjectConverter.toMap()` preserves nulls in the map, exposing it. GraalJS also converts JS `undefined` to Java `null`, so both `null` and `undefined` property values trigger the NPE. Ironically, Nashorn's `toList()` does filter nulls (same if (converted != null) pattern). The map path just missed it.

We should add an early return in `handleElement()` when `value == null`, consistent with GraalJS behavior (skip nulls entirely).

Connected to the https://github.com/enonic/app-standardidprovider/issues/1206

PR #11774, opened 2026-02-17
NPE in ScriptValueTranslator.handleElement() on Nashorn #11773

edloidas:


rymsha:
#11797 is essentially the same thing, but with Java 25 syntax

PR #11784, opened 2026-02-18
Improve handling of instant input type #11783

### LINKED ISSUE #11783: Improve handling of instant input type
Improve handling of instant input type

anatol-sialitski:


modules/core/core-api/src/main/java/com/enonic/xp/data/ValueTypes.java
```diff
@@ -74,6 +74,10 @@ private static void register( ValueType valueType )
 
     public static ValueType getByName( final String name )
     {
+        if ( "instant".equalsIgnoreCase( name ) )
```
rymsha: why ignore case?

rymsha (changes requested):

rymsha (changes requested):
There is no Instant value type. It is lib-admin-ui that fails to lookup a correct ValueTupe

Issue #11464, opened 2025-10-09
Remove resource resolving logic from descriptor


anatol-sialitski:
Remove resource resolving logic from descriptor

PR #11800, opened 2026-02-24
Remove resource resolving logic from descriptor #11464

anatol-sialitski:


modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/SlashApiHandler.java
```diff
@@ -356,9 +356,9 @@ private void applyWebSocketIfPresent( final WebSocketContext webSocketContext, f
         }
     }
 
-    private ControllerScript getScript( final DescriptorKey descriptorKey )
+    private ControllerScript getScript( final DescriptorKey key )
```
rymsha, line 358: Converting an X descriptor key to an X ResourceKey is a duty of an XDescriptorService

rymsha (changes requested):

PR #11832, opened 2026-03-04
Json Schema for descriptors #11668

anatol-sialitski:


modules/core/core-jsonschema/src/main/schema/fragments/longDef.json
rymsha, line 1: I don't think json supports biggest integers Long supports.

PR #11861, opened 2026-03-16
type def improvements

alansemenov:


modules/lib/core/index.d.ts
rymsha: it does not cover the case than parameter developer asked for is not in the request - undefined

modules/lib/lib-auth/src/main/resources/lib/xp/auth.ts
rymsha: there is no way to say `as PricnipalType` because there is no PrincipalType

Issue #11872, opened 2026-03-19
Support include for FormFragment


anatol-sialitski:
Support include for FormFragment

PR #11873, opened 2026-03-20
Support `include` for FormFragment #11872

anatol-sialitski:


modules/core/core-schema/src/main/java/com/enonic/xp/core/impl/schema/mapper/FormItemDeserializer.java
```diff
@@ -30,6 +32,16 @@ public FormItem deserialize( final JsonParser jsonParser, final DeserializationC
 
         final JsonNode node = mapper.readTree( jsonParser );
 
+        final String include = node.path( "include" ).asText( null );
+
+        if ( include != null && node.size() == 1 )
+        {
+            final FormFragmentName fragmentName = include.contains( ":" )
```
rymsha: Form fragments from other apps are prohibited

Issue #10430, opened 2024-03-01
Content js API support validationErrors


ComLock:
Something like this, someone should probably check java?

```
validationErrors?: {
	errorCode: string
}
```

PR #11950, opened 2026-04-17
Content js API support validationErrors #10430

ashklianko:


modules/lib/core/index.d.ts
```diff
@@ -565,16 +565,11 @@ export interface AttachmentValidationError
...
     i18n: string;
-    errorCode: ValidationErrorCode;
+    errorCode: string;
     /* eslint-disable @typescript-eslint/no-explicit-any*/
     args: any[];
 }
 
-export interface ValidationErrorCode {
-    applicationKey: string;
-    code: string;
```
rymsha, line 575: where did the code go? inside the string? app:code ?

ashklianko: yes, ValidationErrorsMapper calls `gen.value( "errorCode", error.getErrorCode().toString() );` that becomes 'app:code'

modules/lib/lib-content/src/main/resources/lib/xp/content.ts
```diff
@@ -185,7 +187,7 @@ export interface PatchableContent<
     modifyAttachments: Attachment;
     removeAttachments: string[];
     createAttachments: AddAttachmentParam[];
-    validationErrors: ValidationError[];
+    validationErrors?: ValidationError[];
```
rymsha: None of other fields are optional. It is a different level problem

rymsha:
superseded by #11964

PR #11990, opened 2026-05-02
Add JS API to create id providers #9026

### LINKED ISSUE #9026: Add possibility to create an userstore from javascript
I don't think there currently is any library available to create a userstore. 

The use case is for idproviders to create a userstore when the app is started. 
Also need a function to get all userstores from the system. This is to not create duplicate userstores.

rymsha:
Closes #9026.

Adds a JS API for creating and listing id providers from server-side scripts.

## Use case

Id provider apps that self-provision their id provider on app startup currently have no way to do that from JS — they need to either rely on `system.properties` bootstrapping or duplicate provisioning logic. This API closes that gap.

## API (`/lib/xp/auth`)

```ts
createIdProvider(params: {
    key: string;
    displayName: string;
    description?: string;
    idProviderConfig?: { applicationKey: string; config?: object };
    permissions?: { principal: PrincipalKey; access: IdProviderAccess }[];
}): IdProvider;

getIdProviders(): IdProvider[];
```

`IdProvider` matches the shape returned by other auth-lib functions (`key`, `displayName`, `description`, `idProviderConfig`).

Permissions on the underlying id provider node are still enforced by `SecurityServiceImpl` — callers without admin privileges get the same `NodeAccessException` they would from the REST endpoint.

## Implementation notes

- `CreateIdProviderHandler` / `GetIdProvidersHandler` in `lib-auth` delegate to `SecurityService` (no logic duplicated).
- New `IdProviderMapper` in `lib-common` mirrors the existing `PrincipalMapper`.
- Examples under `examples/auth/createIdProvider.js` and `examples/auth/getIdProviders.js`.

## Tests

- `CreateIdProviderHandlerTest` — full-fields, minimal-fields, missing-key, missing-displayName, duplicate-key.
- `GetIdProvidersHandlerTest` — populated and empty result.
- `./gradlew :lib:lib-auth:test :lib:lib-common:test :lib:lib-auth:check :lib:lib-common:check` all green.

## Verification

End-to-end verification via companion PR in `enonic/app-simple-idprovider`: enonic/app-simple-idprovider#72.

### DIFF
--- a/modules/lib/lib-auth/src/main/java/com/enonic/xp/lib/auth/CreateIdProviderHandler.java
+++ b/modules/lib/lib-auth/src/main/java/com/enonic/xp/lib/auth/CreateIdProviderHandler.java
@@ -5,15 +5,12 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.function.Supplier;
 
-import com.enonic.xp.app.ApplicationKey;
-import com.enonic.xp.data.PropertyTree;
 import com.enonic.xp.lib.common.IdProviderMapper;
 import com.enonic.xp.script.ScriptValue;
 import com.enonic.xp.script.bean.BeanContext;
 import com.enonic.xp.script.bean.ScriptBean;
 import com.enonic.xp.security.CreateIdProviderParams;
 import com.enonic.xp.security.IdProvider;
-import com.enonic.xp.security.IdProviderConfig;
 import com.enonic.xp.security.IdProviderKey;
 import com.enonic.xp.security.PrincipalKey;
 import com.enonic.xp.security.SecurityService;
@@ -32,8 +29,6 @@ public final class CreateIdProviderHandler
 
     private String description;
 
-    private ScriptValue idProviderConfig;
-
     private ScriptValue permissions;
 
     public void setKey( final String key )
@@ -51,11 +46,6 @@ public final class CreateIdProviderHandler
         this.description = description;
     }
 
-    public void setIdProviderConfig( final ScriptValue idProviderConfig )
-    {
-        this.idProviderConfig = idProviderConfig;
-    }
-
     public void setPermissions( final ScriptValue permissions )
     {
         this.permissions = permissions;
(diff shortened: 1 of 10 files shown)


modules/lib/lib-auth/src/main/java/com/enonic/xp/lib/auth/CreateIdProviderHandler.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    private IdProviderAccessControlList buildPermissions()
+    {
+        if ( this.permissions == null )
+        {
+            return null;
+        }
+        final List<ScriptValue> entries = Optional.ofNullable( this.permissions.getArray() ).orElse( List.of() );
+        if ( entries.isEmpty() )
```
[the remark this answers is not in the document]
rymsha: @copilot fix

modules/lib/lib-auth/src/test/resources/test/getIdProviders-test.js
```diff
@@ -0,0 +1,34 @@
...
+exports.getIdProviders = function () {
+
+    var result = auth.getIdProviders();
+
+    var expectedJson = [
+        {
+            'key': 'idProviderTestKey',
+            'displayName': 'Id Provider test',
+            'description': 'Id Provider used for testing',
+            'idProviderConfig': {
```
anatol-sialitski:
```suggestion
            'config': {
```

rymsha:
@copilot idPrividerConfig does not seem right - it is not for create/update via api. It is used buy some id providers to provision idps. Remove from js api

modules/lib/lib-auth/src/test/resources/test/getIdProviders-test.js
rymsha: @copilot remove idPrividerConfig entirely from mappings

anatol-sialitski (approved):

Issue #11991, opened 2026-05-02
Audit lib JSDoc, TS types, Java handlers


rymsha:
For every lib under `modules/lib/`, verify:
- JSDoc `@param` / `@return` types match the corresponding
  TypeScript declaration's types.
- JSDoc descriptions accurately reflect what the underlying Java
  handler actually does (param semantics, optionality, edge cases,
  side effects).
- TypeScript type declarations match the Java handler's actual
  runtime signature.

The Java handler is treated as the runtime truth: when JSDoc / TS
disagrees with Java, the docs are updated to match (not the other
way around — code semantics are not changed by this audit).

## Libs in scope (one PR per lib if discrepancies are found)

- [ ] lib-admin
- [ ] lib-app
- [ ] lib-auditlog
- [ ] lib-auth
- [ ] lib-cluster
- [ ] lib-common
- [ ] lib-content
- [ ] lib-context
- [ ] lib-event
- [ ] lib-export
- [ ] lib-grid
- [ ] lib-i18n
- [ ] lib-io
- [ ] lib-mail
- [ ] lib-node
- [ ] lib-portal
- [ ] lib-project
- [ ] lib-repo
- [ ] lib-scheduler
- [ ] lib-schema
- [ ] lib-sse
- [ ] lib-task
- [ ] lib-value
- [ ] lib-vhost
- [ ] lib-websocket

Per-lib audit progress is tracked as comments on this issue. PRs
that fix discrepancies reference this issue with `Part of #<n>` (not
`Closes`, since the umbrella stays open until all libs are audited).

rymsha:
lib-admin audited — fixes in #11992. Three discrepancies in admin.ts: widgetUrl/extensionUrl JSDoc `[params]` (optional) didn't match required TS signature; GetHomeToolUrlParams.type was required in TS but optional in JSDoc and Java handler.

rymsha:
lib-app audited — fixes in #11993. Dropped dead `hasVirtual()` (Java handler was deleted in #9764), widened `get()` and `getDescriptor()` return types to allow null, fixed `Application.started`/`system` to be non-nullable, tightened a few JSDoc `@returns` types, and removed stale `hasVirtual`/`hasReal` mentions from the README.

rymsha:
lib-auditlog audited — fixes in #11994. Two JSDoc discrepancies fixed: `get()` return type now reflects nullable, and `find()` count description corrected from a copy-pasted "Number of contents to fetch".

[...]

PR #11992, opened 2026-05-02
lib-admin JSDoc/TS fixes #11991

rymsha:
Part of #11991.

Audit findings for lib-admin (`modules/lib/lib-admin/src/main/resources/lib/xp/admin.ts`):

- `widgetUrl`: JSDoc marked `[params]` optional, but the TS signature has `params: WidgetUrlParams` (required) and the implementation calls `checkRequired(params, 'application')` which throws if missing. Removed the brackets so JSDoc matches TS / runtime behavior.
- `extensionUrl`: same issue as `widgetUrl`. Removed the brackets on `[params]`.
- `GetHomeToolUrlParams`: TS interface declared `type: HomeToolUrlType` (required), but the JSDoc says `[params.type=server]` (optional) and the Java handler accepts a null URL type via `__.nullOrValue(params?.type)`. Made the field optional (`type?: HomeToolUrlType`).

### DIFF
--- a/modules/lib/lib-admin/src/main/resources/lib/xp/admin.ts
+++ b/modules/lib/lib-admin/src/main/resources/lib/xp/admin.ts
@@ -72,7 +72,7 @@ export function getHomeToolUrl(params?: GetHomeToolUrlParams): string {
 }
 
 export interface GetHomeToolUrlParams {
-    type: HomeToolUrlType;
+    type?: HomeToolUrlType;
 }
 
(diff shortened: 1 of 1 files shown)


rymsha:
Reviewed Copilot feedback: applied 0, skipped 0. Copilot's review (id 4215170031) is a top-level descriptive overview of the PR's intent with no inline comments and no actionable suggestions, so no changes were needed.

anatol-sialitski (approved):

PR #11994, opened 2026-05-02
lib-auditlog JSDoc/TS fixes #11991

rymsha:
Part of #11991.

Audit findings for `lib-auditlog`:

- `get()` JSDoc: `@returns {object}` did not match the Java handler. `GetAuditLogHandler.doExecute()` returns `null` when no audit log exists for the given id, and the TS signature already says `AuditLog | null`. JSDoc updated to `{object|null}` with a description note.
- `find()` JSDoc: `@param {number} [params.count=10] Number of contents to fetch.` is a stale copy-paste from content-oriented libs — `lib-auditlog` returns audit logs, not contents. Description corrected to "Number of audit logs to fetch."

Java handler behavior is unchanged — only JSDoc was updated to match what the code actually does. TS type declarations were already correct and untouched.

Verified with `./gradlew :lib:lib-auditlog:test` — green.

### DIFF
--- a/modules/lib/lib-auditlog/src/main/resources/lib/xp/auditlog.ts
+++ b/modules/lib/lib-auditlog/src/main/resources/lib/xp/auditlog.ts
@@ -111,7 +111,7 @@ interface GetAuditLogHandler {
  * @param {object} params     JSON with the parameters.
  * @param {string} params.id  Id of the audit log.
  *
- * @returns {object} Audit log as JSON.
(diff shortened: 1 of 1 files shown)


rymsha:
Reviewed Copilot feedback: applied 0, skipped 0. Copilot's only review was a top-level descriptive overview of the PR's changes (no inline comments, no suggestions, no objections), so there was nothing to apply or reject.

anatol-sialitski (approved):

PR #11995, opened 2026-05-02
Fix lib-auth JSDoc #11991

rymsha:
Part of #11991.

Audited `modules/lib/lib-auth/` against the Java handlers and TypeScript types. The Java handler is the runtime truth; this PR only updates JSDoc and JSDoc-only wording in `auth.ts`. No Java behavior or TS signatures are changed.

## Findings fixed

- **`getUser`** — said "returns *undefined*" when not logged in; Java returns `null`. Also updated `@returns {User}` to `@returns {User | null}` (TS signature is already `User | null`).
- **`createUser`** — was missing `@returns {User}`.
- **`createGroup`** — was missing `@returns {Group}`; the description for `params.description` was the broken "`as principal description .`" — replaced with `Group description.`.
- **`createRole`** — was missing `@param {object} params JSON parameters.`, missing `@returns {Role}`, and had the same broken `params.description` text — replaced with `Role description.`.
- **`modifyUser`** — `@returns {User}` had a stray double-space typo and dropped the `null` case; tightened to `@returns {User | null}` and cleaned the wording.
- **`modifyGroup` / `modifyRole`** — same null-case fix applied to `@returns`.
- **`addMembers`** — description said "Adds members to a principal (user or role)"; you don't add members to a user. The `principalKey` parameter is typed `GroupKey | RoleKey` in TS — corrected to "(group or role)".
- **`removeMembers`** — same "user or role" wording fix as `addMembers`. Also `@param {string} members` was wrong: `members` is an array (`string[]` / `(UserKey | GroupKey)[]`).
- **`findUsers`** — `params.count` comment said "Number of contents to fetch" (stale copy from `lib-content`) — corrected to "Number of users to fetch".

## Verification

- `./gradlew :lib:lib-auth:test` — green (TypeScript declarations regenerate as part of the build).
- No `integrationTest` source set in `lib-auth` and no `itest-lib-auth` module.

### DIFF
--- a/modules/lib/lib-auth/src/main/resources/lib/xp/auth.ts
+++ b/modules/lib/lib-auth/src/main/resources/lib/xp/auth.ts
@@ -160,14 +160,14 @@ export function getUser<Profile extends Record<string, unknown> = Record<string,
     includeProfile: true
 }): UserWithProfile<Profile> | null;
 /**
- * Returns the logged-in user. If not logged-in, this will return *undefined*.
+ * Returns the logged-in user. If not logged-in, this will return *null*.
  *
  * @example-ref examples/auth/getUser.js
  *
  * @param {object} [params] JSON parameters.
  * @param {boolean} [params.includeProfile=false] Include profile.
  *
- * @returns {User} Information for logged-in user.
+ * @returns {User | null} Information for the logged-in user, or null if not logged in.
  */
 export function getUser(params?: GetUserParams): User | null {
     const {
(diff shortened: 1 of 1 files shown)


rymsha:
Reviewed Copilot feedback: applied 0, skipped 0. Copilot's review (id 4215169495) is a high-level overview only — no inline comments and no concrete suggestions to act on. PR left unchanged.

anatol-sialitski (approved):

PR #11997, opened 2026-05-02
lib-context JSDoc/TS fixes #11991

rymsha:
Part of #11991.

Audit findings for lib-context — JSDoc and TS types disagreed with what the Java handler (`ContextHandlerBean` / `ContextMapper`) actually does. Fixed:

- `run` `@returns {object}` → `{*}` — the function returns whatever the callback returns (generic `T`), not necessarily an object. Tests demonstrate string returns.
- `Context.branch` / `Context.repository` typed as required `string` → made optional. `ContextMapper.serialize` only emits these keys when `getBranch() != null` / `getRepositoryId() != null`.
- `AuthInfo.user?: User | null` and `principals?: PrincipalKey[] | null` → dropped the `| null`. `ContextMapper.serializeUser` and `serializePrincipals` skip the field when null and never emit a literal `null`, so the runtime shape is "present or absent", not "present or null".

`./gradlew :lib:lib-context:test` is green locally.

### DIFF
--- a/modules/lib/lib-context/src/main/resources/lib/xp/context.ts
+++ b/modules/lib/lib-context/src/main/resources/lib/xp/context.ts
@@ -18,15 +18,15 @@ import type {PrincipalKey, ScriptValue, User} from '@enonic-types/core';
 export type {PrincipalKey, UserKey, Principal, ScriptValue, User} from '@enonic-types/core';
 
 export interface AuthInfo {
-    user?: User | null;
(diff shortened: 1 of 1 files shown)


rymsha:
Reviewed Copilot feedback: 1 review (high-level overview only, no inline comments or concrete suggestions). Applied 0, skipped 0 — nothing actionable to act on.

anatol-sialitski (approved):

PR #11998, opened 2026-05-02
lib-event JSDoc/TS fixes #11991

rymsha:
Part of #11991.

Audit findings for `lib-event`:

- `listener` — `params.localOnly` is optional in the TS interface (`localOnly?: boolean`) but the JSDoc didn't use the `[name]` brackets convention. Updated to `[params.localOnly=false]`, matching the project convention used in `lib-auth` / `lib-context` and the `setLocalOnly(params.localOnly === true)` runtime default.
- `send` — `event.distributed` is optional in the TS interface (`distributed?: boolean`); the Java `EventSenderHelper.initialize` sets the default to `false`. Updated JSDoc to `[event.distributed=false]`.
- `send` — `event.data` is optional in the TS interface (`data?: EventData`); the Java handler only sets data when the value is a non-null object. Updated JSDoc to `[event.data]`.

Java handler behavior is unchanged — only JSDoc annotations on `event.ts` were updated to match the TypeScript declarations and the Java runtime defaults.


rymsha:
Reviewed Copilot feedback: applied 0, skipped 0. Copilot's review was an overview-only summary (no inline comments, no suggested changes), so there was nothing actionable to apply. CI remains green.

anatol-sialitski (approved):

PR #12000, opened 2026-05-02
lib-grid JSDoc/TS fixes #11991

rymsha:
Part of #11991.

Audit findings for lib-grid (`modules/lib/lib-grid`):

- `modify()` declared a non-nullable `Map[Key]` return type, but the Java `SharedMap.modify` (see `LocalSharedMap.modify`) returns `null` when the modifier callback returns `null` and the entry is removed. The JSDoc `@returns` text already mentioned the null case — only the TS type was wrong. Fixed in `JavaSharedMap.modify`, `SharedMap.modify`, and `SharedMapImpl.modify`.
- `modify()` JSDoc was missing a `@returns {string|number|boolean|JSON|null}` type annotation (the other methods include it). Added.
- `get()` JSDoc had a duplicate word: `@param {string} key - key the key whose...`. Cleaned up to match `delete()`'s style.
- The class-level WARNING listed "Strings, numbers, and pure JSON objects" as supported, but `SharedMapValueType` and the `set()` JSDoc both list `boolean` explicitly. Added "booleans" to the WARNING.

Java handler behavior is unchanged — only JSDoc and TS type declarations were updated to match what the Java code does.

Tests: `./gradlew :lib:lib-grid:test` is green. No integration test source set in this module.

### DIFF
--- a/modules/lib/lib-grid/src/main/resources/lib/xp/grid.ts
+++ b/modules/lib/lib-grid/src/main/resources/lib/xp/grid.ts
@@ -29,7 +29,7 @@ interface JavaSharedMap<Map extends GridMap> {
 
     set<Key extends keyof Map>(key: Key, value: ConvertedType, ttlSeconds?: number | null): void;
 
-    modify<Key extends keyof Map>(key: Key, modifier: SharedMapModifierFn<Map, Key>, ttlSeconds?: number | null): Map[Key];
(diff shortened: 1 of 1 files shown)


rymsha:
Reviewed Copilot feedback: 1 review found, 0 inline comments. The single review was a high-level PR overview summarizing the changes — no actionable critique or suggestions. Applied 0, skipped 0.

anatol-sialitski (approved):

PR #12003, opened 2026-05-02
lib-node JSDoc/TS fixes #11991

rymsha:
Part of #11991.

Audit findings for `lib-node` — JSDoc and TypeScript declarations were
checked against the Java handler runtime truth (under
`modules/lib/lib-node/src/main/java/com/enonic/xp/lib/node/`). Where
they disagreed, the JS-side declarations were updated. **Java behavior
is unchanged.**

### JSDoc-only

- `applyPermissions`: scope value typo `'SINGE'` → `'SINGLE'`
  (Java enum `ApplyPermissionsScope` is `SINGLE | TREE | SUBTREE`).
- `findChildren`: `parentKey` typed `{number}` → `{string}` (it's a
  path or id).
- `connect`, `multiRepoConnect`: `repoId` and `branch` typed
  `{object}` → `{string}`.
- `push`: `key` and `keys` are individually optional (the runtime
  requires at least one) — JSDoc now uses `[params.key]` /
  `[params.keys]`.
- `getActiveVersion`: return description "Active content versions
  per branch" → "Active node version, or null if not found"
  (`GetActiveVersionHandler` returns one `NodeVersion` or `null`).
- `commit`: drop the misleading `...` varargs notation; document
  the `params` object; `keys` is `string | string[]` (matches
  `CommitParams.keys`).
- `getCommit`: `@example-ref` was pointing at `commit.js`; switched
  to the dedicated `getCommit.js` example.
- `patch`: `@example-ref` was pointing at `update.js`; switched to
  `patch-1.js`. Documented the `branches` parameter.

### TypeScript-only

- `PatchNodeResult`: dropped the non-existent `nodeId` field; renamed
  `branchResult` (singular) → `branchResults` (plural) — that's what
  `PatchNodeResultMapper` actually emits.
- `PatchNodeParams.branches`: now optional, matching the JS impl
  (which defaults it to `[]`) and `PatchNodeParams` Java builder
  (which defaults to `Branches.empty()`).
- `PushNodesResult`: dropped the non-existent `deleted` field
  (`PushNodesResultMapper` only emits `success` and `failed`).
- `SortNodeResult.reorderedNodes`: shape is `{ node: Node }[]` (not
  `Node[]`), and always emitted — matches `SortNodeResultMapper`.
- `ApplyPermissionsResult` / `BranchResult`: rewrote to match
  `ApplyPermissionsResultMapper`. The actual shape is
  `Record<string, { branchResults: { branch: string; permissions: { _permissions?: AccessControlEntry[] } }[] }>`,
  not the previous `Record<string, { branch: string; node: Node }[]>`.

### TS + JSDoc

- `move`: returns the moved/renamed `Node`, not `boolean`. The Java
  handler chain (`MoveNodeHandler` → `NodeMapper`) returns a node
  object, and the existing `move-1.js` / `move-2.js` examples
  already access `_path` on the result. JSDoc and the
  `RepoConnection.move` signature both updated accordingly.

## Validation

- `./gradlew :lib:lib-node:test` — green.
- The `:lib:typescript` task (which runs `tsc --declaration`) also
  succeeded on the updated declarations, so the TS changes type-check
  cleanly.
- No `integrationTest` source set is present for `lib:lib-node`.

### DIFF
--- a/modules/lib/lib-node/src/main/resources/lib/xp/node.ts
+++ b/modules/lib/lib-node/src/main/resources/lib/xp/node.ts
@@ -247,7 +247,7 @@ interface NodeHandler {
 
     diff(params: DiffBranchesHandlerParams): DiffBranchesResult;
 
-    move(source: string, target: string): boolean;
+    move<NodeData>(source: string, target: string): Node<NodeData>;
 
     query<
         AggregationInput extends Aggregations = never
@@ -317,7 +317,6 @@ export interface PushNodesResult {
         id: string;
         reason: string;
     }[];
-    deleted: string[];
 }
 
 interface PushNodeHandlerParams {
@@ -337,12 +336,11 @@ interface PushNodeHandlerParams {
 export interface PatchNodeParams {
     key: string;
     editor: (node: Node) => PatchedNode;
-    branches: string[];
+    branches?: string[];
 }
 
 export interface PatchNodeResult {
-    nodeId: string;
-    branchResult: {
+    branchResults: {
         branch: string;
         node: PatchedNode;
     }[];
@@ -394,7 +392,9 @@ export interface SortNodeParams {
 
 export interface SortNodeResult<NodeData> {
     node: Node<NodeData>;
-    reorderedNodes?: Node<NodeData>[];
+    reorderedNodes: {
+        node: Node<NodeData>;
+    }[];
 }
 
 export interface QueryNodeParams<AggregationInput extends Aggregations = never> {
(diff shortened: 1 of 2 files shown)


modules/lib/lib-node/src/main/resources/lib/xp/node.ts
```diff
@@ -525,11 +525,15 @@ export interface ApplyPermissionsParams {
...
-export type ApplyPermissionsResult = Record<string, BranchResult[]>;
+export type ApplyPermissionsResult = Record<string, {
+    branchResults: BranchResult[];
+}>;
 
 export interface BranchResult {
     branch: string;
-    node: Node;
+    permissions: {
+        _permissions?: AccessControlEntry[];
```
rymsha: @copilot this is a bug. There should be no extra level `_permisions`. Fix

modules/lib/lib-node/src/main/resources/lib/xp/node.ts
```diff
@@ -394,7 +392,9 @@ export interface SortNodeParams {
 
 export interface SortNodeResult<NodeData> {
     node: Node<NodeData>;
-    reorderedNodes?: Node<NodeData>[];
+    reorderedNodes: {
+        node: Node<NodeData>;
```
rymsha, line 396: @copilot This is a bug. The type is correct. Implementation isn't. Fix

anatol-sialitski (approved):

PR #12005, opened 2026-05-02
lib-project JSDoc/TS fixes #11991

rymsha:
Part of #11991.

Audit of `modules/lib/lib-project` comparing JSDoc, TS type declarations, and Java handlers. The Java handler is treated as the runtime truth; only JSDoc and TS were updated.

### TS type bugs
- `Project.readAccess` was typed as `ProjectPermission`, but the Java mapper emits `{ public: boolean }` — fixed to `ProjectReadAccess`.
- `ProjectPermission` was `Record<ProjectRole, string[]>`, claiming all five role keys are present, but the mapper omits roles with empty principal lists and the input side accepts any subset — fixed to `Partial<Record<...>>`.
- `Project.description` was non-null required, but Java emits null when description is unset — marked optional.
- `Project.siteConfig` was required, but the mapper only emits the key when non-empty — marked optional.
- `ModifyProjectParams.displayName` was required in TS, but the Java handler falls back to the existing display name when null (and JSDoc already marked it optional) — marked optional.
- `ModifyProjectReadAccessParams.readAccess` was optional in TS, but the Java handler throws `IllegalArgumentException` when `readAccess` or its `public` member is missing — marked required.

### JSDoc bugs
- `create()`: `params.readAccess` was bracketed (optional), but Java throws when missing — removed brackets.
- `create()` / `modify()`: `params.siteConfig` was typed `{object}`; it's an array — fixed to `{Object[]}`.
- `create()` / `addPermissions()` / `removePermissions()`: described bogus `params.permissions.role` and `params.permissions.principals` fields — the permissions object is a record keyed by role, not a shape with `role` / `principals` fields. Removed the misleading lines and consolidated the description on the parent param tag.
- `addPermissions()` / `removePermissions()`: `params.permissions` was unbracketed (required), but TS marks it optional and the Java handler treats null as a no-op (returns current permissions) — marked optional in JSDoc.
- `getAvailableApplications()`: `@param {GetProjectParams}` referenced the wrong type — fixed to `GetAvailableApplicationsParams`.

## Verification

`./gradlew :lib:lib-project:test` — green. TypeScript compilation (`tsc --declaration --project tsconfig.build.json`) — green. No `integrationTest` source set exists for this module.

### DIFF
--- a/modules/lib/lib-project/src/main/resources/lib/xp/project.ts
+++ b/modules/lib/lib-project/src/main/resources/lib/xp/project.ts
@@ -49,7 +49,6 @@ export interface CreateProjectParams<Config extends Record<string, unknown>> {
     displayName: string;
     description?: string;
     language?: string;
-    parent?: string;
     parents?: string[];
     siteConfig?: SiteConfig<Config>[];
     permissions?: ProjectPermission;
@@ -60,7 +59,6 @@ export interface Project<Config extends Record<string, unknown> = Record<string,
     id: string;
     displayName: string;
     description?: string;
-    parent?: string;
     parents: string[];
     siteConfig?: SiteConfig<Config>[];
     language?: string;
@@ -100,7 +98,6 @@ interface CreateProjectHandler<Config extends Record<string, unknown>> {
  * @param {string} params.displayName Project's display name.
  * @param {string} [params.description] Project description.
  * @param {string} [params.language] Default project language.
- * @param {string} [params.parent] Deprecated: use 'parents' param. Parent project id.
  * @param {string[]} [params.parents] Parent project ids.
  * @param {Object[]} [params.siteConfig] Connected applications config.
  * @param {Object.<string, string[]>} [params.permissions] Project permissions. 1 to 5 properties where key is role id and value is an array of principals.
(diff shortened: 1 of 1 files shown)


rymsha:
Reviewed Copilot feedback: applied 2, skipped 0.

- 3177113589 — `ProjectPermissions.permissions` made required: confirmed by `ProjectPermissionsMapper.doSerialize` (always emits the outer `permissions` key for non-null values) and `ModifyProjectPermissionsHandler` (always passes a non-null `ProjectPermissions`). `ModifyProjectPermissionsHandlerTest.js` expects `{permissions: {}}` even on null-input no-op, confirming the key is always present.
- 3177113598 — `Project.readAccess` made required: `ProjectMapper` is the only producer of `Project`, and every call site (Create/Get/List/Modify) calls `setIsPublic(...)` with a non-null `Boolean`, so `ProjectReadAccessMapper` always emits the `readAccess` map. `GetProjectHandlerTest.js` expects `readAccess: {public: true}` in every case.

Force-pushed to keep the PR at a single commit. New head: 511dfd356f.

modules/lib/lib-project/src/main/resources/lib/xp/project.ts
```diff
@@ -59,13 +59,13 @@ export interface CreateProjectParams<Config extends Record<string, unknown>> {
 export interface Project<Config extends Record<string, unknown> = Record<string, unknown>> {
     id: string;
     displayName: string;
-    description: string;
+    description?: string;
     parent?: string;
```
rymsha: @copilot Parent was deprecated and now must be removed - use parents

anatol-sialitski (approved):

PR #12012, opened 2026-05-02
lib-websocket JSDoc/TS fixes #11991

rymsha:
Part of #11991.

Audit findings for lib-websocket:
- `getGroupSize` was missing a JSDoc `@returns` tag. The Java handler (`WebSocketManagerBean.getGroupSize`) returns `int` and the TS signature returns `number`, but the JSDoc gave no clue about the return value. Added `@returns {number} Number of sockets in the group.`

All other functions (`addToGroup`, `removeFromGroup`, `send`, `sendToGroup`) — JSDoc, TypeScript types, and Java handler signatures already agree.

Java handler behavior is unchanged — only JSDoc was updated to match what the code does.

### DIFF
--- a/modules/lib/lib-websocket/src/main/resources/lib/xp/websocket.ts
+++ b/modules/lib/lib-websocket/src/main/resources/lib/xp/websocket.ts
@@ -83,6 +83,8 @@ export function sendToGroup(group: string, message: string): void {
  * @example-ref examples/websocket/addToGroup.js
  *
  * @param {string} group Group name.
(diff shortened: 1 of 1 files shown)


rymsha:
Reviewed Copilot feedback: applied 0, skipped 0. Copilot left only a high-level descriptive review of the PR (overview/summary), with no inline comments and no actionable suggestions. Nothing to apply.

anatol-sialitski (approved):

PR #12020, opened 2026-05-04
Expose sha512 and textContent on attachments

rymsha:
## Summary

The lib-content [`AttachmentsMapper`](https://github.com/enonic/xp/blob/master/modules/lib/lib-content/src/main/java/com/enonic/xp/lib/content/mapper/AttachmentsMapper.java) only emitted `name`, `label`, `size`, `mimeType` — even though the underlying `Attachment` model has carried `sha512` and `textContent` for some time. As a result, those two fields were invisible to anything reading attachments via `/lib/xp/content` (e.g. `contentLib.get(...)` / `contentLib.getAttachments(...)`), including Content Studio's JSON preview, which renders the lib-content view rather than the app-contentstudio-side `AttachmentJson`.

This PR surfaces both fields:

- `AttachmentsMapper#serializeAttachment` now emits `sha512` and `textContent` (`MapGenerator.value` handles `null` fine; on the script side the existing `ScriptMapGenerator` filters nulls out, so existing JS callers that don't expect these keys are unaffected).
- The `Attachment` interface in `@enonic-types/core` (`modules/lib/core/index.d.ts`) gains optional `sha512?: string | null` and `textContent?: string | null`.
- New `AttachmentsMapperTest` covers both the all-fields and null cases against `JsonMapGenerator`.

## Context

Follow-up to [app-contentstudio#10388 (comment by @arod1ique)](https://github.com/enonic/app-contentstudio/pull/10388#issuecomment-4371089923):

> To see the result in JSON preview we need to update modules/lib/lib-content/src/main/java/com/enonic/xp/lib/content/mapper/AttachmentsMapper.java

PR #10388 in app-contentstudio already plumbed `sha512` / `textContent` through the REST response and the TS model. The JSON preview path is independent and lives here in xp's lib-content.

## Test plan

- [x] `./gradlew :lib:lib-content:test` is green locally (new `AttachmentsMapperTest` runs 2 cases; existing `GetAttachmentsHandlerTest` JS spec still passes since null-valued fields are dropped by `ScriptMapGenerator`).
- [ ] CI green.
- [ ] Spot-check on a local XP: read an attachment via `contentLib.get(...)` / `contentLib.getAttachments(...)` and verify `sha512` / `textContent` are present when set.

### DIFF
--- a/modules/lib/core/index.d.ts
+++ b/modules/lib/core/index.d.ts
@@ -563,6 +563,8 @@ export interface Attachment {
     label?: string;
     size: number;
     mimeType: string;
+    sha512?: string | null;
+    textContent?: string | null;
 }
 
 export interface DataValidationError
--- a/modules/lib/lib-content/src/main/java/com/enonic/xp/lib/content/mapper/AttachmentsMapper.java
+++ b/modules/lib/lib-content/src/main/java/com/enonic/xp/lib/content/mapper/AttachmentsMapper.java
@@ -33,5 +33,7 @@ public final class AttachmentsMapper
         gen.value( "label", attachment.getLabel() );
         gen.value( "size", attachment.getSize() );
         gen.value( "mimeType", attachment.getMimeType() );
+        gen.value( "sha512", attachment.getSha512() );
+        gen.value( "textContent", attachment.getTextContent() );
     }
 }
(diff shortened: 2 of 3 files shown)


ashklianko (commented):
I can see sha515 now, but tried couple .txt files and still can't get non-empty "textContent"

ashklianko (approved):
Works, but spawns a question: What is 'textContent' property and what it used for?

rymsha:
Heads up — the `textContent` half of this PR was relocated. Architectural argument: `textContent` is the Tika-extracted text from a media binary, so it's meaningful only for media content types, not every attachment. `sha512` was correct as introduced here.

Follow-up: #12021 moves `textContent` off `Attachment` and onto the media content's own data at `data.media.text`. The `sha512` half from this PR stays exactly as merged.

Paired contentstudio correction: https://github.com/enonic/app-contentstudio/pull/10392

PR #12021, opened 2026-05-04
Move attachment textContent to media data

### LINKED ISSUE #12022: Move attachment textContent to media data
text is only relevant for media

rymsha:
## Summary

Follow-up to PR #12020. PR #12020 exposed `sha512` and `textContent` on the `Attachment` JS/TS surface (and added `textContent` storage to `Attachment` itself). The `sha512` half is correct — it's a binary-level concern. The `textContent` half is misplaced: it's the Tika-extracted text from a media binary, so it's meaningful **only for media content types** (PDFs, docs, …), not for every attachment.

Non-media attachments (page banners, article source attachments, project icons, …) always carried a null `textContent`, which was misleading both in the JS API and in Content Studio's JSON preview. Conceptually it belongs on the media content's own data, not on every attachment.

## Changes

- `sha512` stays on `Attachment` (no change to the #12020 fix).
- `textContent` is removed from:
  - `Attachment` (in-memory model, equals/hashCode/toString)
  - `CreateAttachment` (input DTO; `text(...)` builder method dropped)
  - `AttachmentsMapper` (lib-content JS serializer)
  - `AttachmentsDeserializer` (lib-content JS deserializer)
  - `Attachment` interface in `modules/lib/core/index.d.ts`
  - `ModifyAttachmentParam` and `CreateAttachmentParam` in `lib-content/src/main/resources/lib/xp/content.ts`
  - All upstream code that piped `getTextContent()` through (`UpdateContentCommand`, `PatchContentCommand`, `Created/UpdatedEventSyncCommand`, `ProjectServiceImpl`, `ContentAuditLogSupportImpl`)
- `textContent` is now written to the media content's `data.media.text` path:
  - New `MEDIA_TEXT = "text"` and `MEDIA_TEXT_COMPONENT = "media.text"` constants in `ContentPropertyNames` (replacing the removed `ATTACHMENT_TEXT` and `ATTACHMENT_TEXT_COMPONENT`).
  - `MediaFormDataBuilder` accepts `text(...)` and writes `media.text` only when `type.isTextualMedia()` and the value is non-null.
  - `CreateMediaCommand` and `UpdateMediaCommand` route `mediaInfo.getTextContent()` into the form-data builder instead of the attachment.
- `AttachmentConfigProcessor` now configures fulltext indexing for `media.text` instead of `attachment.text`, and only emits the rule for textual-media types (the prior else-branch added a no-op rule on a path nothing wrote to).

## Migration

Migration is handled by `AttachmentTextToMediaUpgrader` (registered in the `model8to9` dump-upgrade bundle, right after `AttachmentSha512Upgrader`). On dump import the upgrader copies the legacy attachment-level `text` value onto `data.media.text` for textual-media content types (`ContentTypeName.isTextualMedia()`) and clears the legacy `attachment[].text` field. The upgrader is gated by project repo prefix and content node type, and is a no-op when there is no legacy `text` to move or when `data.media.text` is already populated.

For in-place upgrades (no dump round-trip), new media saves write to the new path; the legacy `data.attachment[].text` becomes harmless dead data on un-resaved nodes and stops being read.

**Breaking change** (small window): code reading `attachment.textContent` from `/lib/xp/content` since #12020 will now see it absent. Downstream Content Studio gets the same correction in https://github.com/enonic/app-contentstudio (PR follow-up).

## Test plan

- [x] `./gradlew test` green (excluding the pre-existing, unrelated `core-export:ZipExportWriterTest.testPathWithSpecialCharacters` failure on the runner's filesystem)
- [x] `./gradlew :itest:itest-core-content:integrationTest` green for `*ContentServiceImplTest_media*`, `*ContentServiceImplTest_patch*`, and `*ParentContentSynchronizer*`
- [x] `AttachmentsMapperTest` updated: `textContent` no longer in JSON output
- [x] `AttachmentTest` toString assertion updated
- [x] `AttachmentConfigProcessorTest` updated for the new path and the dropped non-textual-media branch
- [x] `PatchContentHandlerTest` (Java + JS fixture) updated
- [x] `ContentServiceImplTest_patch` and `ParentContentSynchronizerTest` updated
- [x] `AttachmentTextToMediaUpgraderTest` covers: textual-media migration, non-media skip, already-migrated no-op, non-project-repo skip, non-content-node skip, non-textual-media skip, no-attachment-text no-op, and the existing-`media.text` preservation case
- [x] `:itest:itest-core:integrationTest --tests *DumpUpgradeIntegrationTest*` green (existing fixture has no legacy `text` so the new upgrader is a no-op there)

### DIFF
--- a/modules/core/core-api/src/main/java/com/enonic/xp/attachment/Attachment.java
+++ b/modules/core/core-api/src/main/java/com/enonic/xp/attachment/Attachment.java
@@ -22,8 +22,6 @@ public final class Attachment
 
     private final String sha512;
 
-    private final String textContent;
-
     public Attachment( final Builder builder )
     {
         this.mimeType = requireNonNull( builder.mimeType, "mimeType is mandatory for an Attachment" );
@@ -31,7 +29,6 @@ public final class Attachment
         this.sha512 = builder.sha512;
         this.size = builder.size;
         this.label = builder.label;
-        this.textContent = builder.textContent;
     }
 
     public String getMimeType()
@@ -74,11 +71,6 @@ public final class Attachment
         return sha512;
     }
 
-    public String getTextContent()
-    {
-        return textContent;
-    }
-
     @Override
     public boolean equals( final Object o )
     {
@@ -92,14 +84,13 @@ public final class Attachment
         }
         final Attachment that = (Attachment) o;
         return Objects.equals( this.name, that.name ) && Objects.equals( this.mimeType, that.mimeType ) &&
-            Objects.equals( this.label, that.label ) && this.size == that.size && Objects.equals( this.sha512, that.sha512 ) &&
-            Objects.equals( this.textContent, that.textContent );
+            Objects.equals( this.label, that.label ) && this.size == that.size && Objects.equals( this.sha512, that.sha512 );
     }
 
     @Override
     public int hashCode()
     {
-        return Objects.hash( name, mimeType, label, size, sha512, textContent );
+        return Objects.hash( name, mimeType, label, size, sha512 );
     }
 
     @Override
@@ -111,7 +102,6 @@ public final class Attachment
         s.add( "label", label );
         s.add( "size", size );
         s.add( "sha512", sha512 );
-        s.add( "textContent", textContent );
         return s.toString();
     }
 
@@ -137,8 +127,6 @@ public final class Attachment
 
         private long size;
 
-        private String textContent;
-
         private Builder()
         {
 
@@ -150,7 +138,6 @@ public final class Attachment
             this.name = attachment.name;
             this.label = attachment.label;
             this.size = attachment.size;
-            this.textContent = attachment.textContent;
             this.sha512 = attachment.sha512;
         }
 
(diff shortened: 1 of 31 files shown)


rymsha:
Added a dump upgrader so legacy `data.attachment[].text` data from PR #12020-era builds is migrated on dump import instead of being lost on next save.

**Commit:** e29f3d923e — _Add dump upgrader for attachment text relocation_

- New `AttachmentTextToMediaUpgrader` registered in `model8to9` right after `AttachmentSha512Upgrader`. Gated by project-repo prefix, content node type, and `ContentTypeName.isTextualMedia()`. Idempotent: returns null when there's nothing to migrate or when `data.media.text` is already populated.
- New `AttachmentTextToMediaUpgraderTest` mirroring the style of the other upgrader tests (`ImageUpgraderTest`, `ReferenceLowercaseUpgraderTest`).
- The PR body's _Migration_ section has been rewritten to describe the upgrader instead of the prior backfill-on-save plan.

**Local CI run:** `./gradlew :core:core-repo:test` and `./gradlew :itest:itest-core:integrationTest --tests *DumpUpgradeIntegrationTest*` both green. Watching the PR's CI status now.

ashklianko (approved):

PR #12030, opened 2026-05-09
Block stop of system apps #12029

### LINKED ISSUE #12029: Prevent stop of system applications via ApplicationService
## Problem

`ApplicationService.stopApplication(ApplicationKey)` has no guard against being called on a system application. The implementation goes straight through `ApplicationServiceImpl#stopApplication` → `doStopApplication` → `ApplicationRegistryImpl#stop` → `bundle.stop()`. There is no check on `Application#isSystem()` anywhere along the path.

System applications carry `X-Bundle-Type: system` in their bundle headers (see `ApplicationBundleUtils#isSystemApplication`) and provide core platform functionality. Stopping one at runtime through the service API leaves the platform in a broken state until the JVM is restarted.

## Scope: stop only

Only `stopApplication` needs the guard. The other lifecycle methods are intentionally left unguarded:

| Method | Guard? | Reason |
|---|---|---|
| `stopApplication(ApplicationKey)` | **YES** | Stopping a system app breaks the platform |
| `startApplication(ApplicationKey)` | no | Starting an already-`ACTIVE` system app is a no-op; starting a non-`ACTIVE` one is legitimate recovery |
| `uninstallApplication(ApplicationKey)` | no | The global-uninstall path won't normally find a system app (system apps are local), and operators who need to remove an extra system app must be able to |
| `uninstallLocalApplication(ApplicationKey)` | no | Called by `DeployDirectoryWatcher` when an operator removes a system-app JAR from `$XP_HOME/deploy/` — that is the legitimate removal path for an extra system app |

`installLocalApplication` is also unguarded — it is the path through which extra system apps arrive (operators can drop a JAR with `X-Bundle-Type: system` into `$XP_HOME/deploy/`, which `DeployDirectoryWatcher` picks up).

The existing guard in `ApplicationServiceImpl#doInstallGlobalApplication` (which rejects system bundles via `appInfo.system`) stays as-is and is unrelated to this change.

The bundle-state guard in `ApplicationRegistryImpl#stop` (lines 227–260) only short-circuits when the bundle isn't `Bundle.ACTIVE` — system apps are normally `ACTIVE` after boot, so the call goes through cleanly and `bundle.stop()` is invoked.

## Reproduction

```java
@Reference
private ApplicationService applicationService;

ApplicationKey key = ApplicationKey.from( "com.enonic.app.contentstudio" ); // any system app
applicationService.stopApplication( key );  // succeeds — bundle stops, platform breaks
```

## Proposed fix

Add an `isSystem()` guard at the top of `ApplicationServiceImpl#stopApplication` only. Sketch:

```java
@Override
public void stopApplication( final ApplicationKey key )
{
    final Application app = registry.get( key );
    if ( app != null && app.isSystem() )
    {
        throw new IllegalArgumentException( "Cannot stop system application: " + key );
    }
    // ... existing logic
}
```

## Tests

- `ApplicationServiceImplTest`: replace the existing `stop_system_application` test (which asserted the bug — that the bundle went `ACTIVE` → `RESOLVED`) with one that asserts `IllegalArgumentException` and that the bundle remains `ACTIVE`. Leave `start_system_application_ignored` as-is.
- Follow-up: add an integration test (`itest-core`) that installs a `X-Bundle-Type: system` test bundle via the deploy path and verifies the stop guard fires in a real OSGi runtime — see follow-up task.

## Notes

- `lib/xp/app` (server-side JS) does not currently expose `stopApplication`, so JS callers are unaffected. The risk is purely from in-process Java callers (other XP modules, third-party app bundles).
- Throwing (rather than silently no-op'ing) surfaces programmer errors clearly. Stopping a system app is never legitimate.

rymsha:
## Summary

Adds a single guard at the top of `ApplicationServiceImpl#stopApplication`: if the application resolved through the registry is a system app (`X-Bundle-Type: system`), throw `IllegalArgumentException`. Stopping a system app at runtime leaves the platform in a broken state until JVM restart, so the call is rejected.

The scope is intentionally narrow: `startApplication`, `uninstallApplication`, and `uninstallLocalApplication` remain unguarded. Starting an already-`ACTIVE` system app is a no-op; uninstalling is a legitimate operator action — particularly via `DeployDirectoryWatcher` when an "extra system app" JAR is removed from `$XP_HOME/deploy/`. The existing `appInfo.system` rejection in `doInstallGlobalApplication` is unrelated and stays as-is.

`ApplicationServiceImplTest`: the previous `stop_system_application` test asserted the bug (bundle went `ACTIVE` → `RESOLVED`); it's replaced with `stopApplication_systemApp_throws`, which asserts `IllegalArgumentException` and that the bundle remains `ACTIVE`. The pre-existing `start_system_application_ignored` test is left intact.

Closes #12029

## Test plan
- [x] `./gradlew :core:core-app:test --tests '*ApplicationServiceImplTest*'` (passes)
- [x] `./gradlew :core:core-app:test` (passes)

### DIFF
--- a/modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationServiceImpl.java
+++ b/modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationServiceImpl.java
@@ -122,7 +122,6 @@ public final class ApplicationServiceImpl
     @Override
     public void startApplication( final ApplicationKey key )
     {
-        requireNotSystemApp( key, "start" );
         final boolean global = !localApplicationSet.contains( key );
         ApplicationHelper.runWithContext( () -> {
             if ( global )
@@ -144,7 +143,11 @@ public final class ApplicationServiceImpl
     @Override
     public void stopApplication( final ApplicationKey key )
     {
-        requireNotSystemApp( key, "stop" );
+        final Application app = registry.get( key );
+        if ( app != null && app.isSystem() )
+        {
+            throw new IllegalArgumentException( "Cannot stop system application: " + key );
+        }
         final boolean global = !localApplicationSet.contains( key );
         ApplicationHelper.runWithContext( () -> {
             if ( global )
@@ -188,7 +191,6 @@ public final class ApplicationServiceImpl
     @Override
     public void uninstallApplication( final ApplicationKey key )
     {
-        requireNotSystemApp( key, "uninstall" );
         if ( localApplicationSet.contains( key ) )
         {
             throw new ApplicationBundleException(
@@ -209,7 +211,6 @@ public final class ApplicationServiceImpl
     @Override
     public void uninstallLocalApplication( final ApplicationKey key )
     {
-        requireNotSystemApp( key, "uninstall" );
         final boolean removed = localApplicationSet.remove( key );
         if ( !removed )
         {
@@ -217,15 +218,6 @@ public final class ApplicationServiceImpl
         }
         doUninstallApplication( key );
         ApplicationHelper.runWithContext( () -> doReinstallStoredApplication( key ) );
-    }
-
-    private void requireNotSystemApp( final ApplicationKey key, final String op )
-    {
-        final Application app = registry.get( key );
-        if ( app != null && app.isSystem() )
-        {
-            throw new IllegalArgumentException( "Cannot " + op + " system application: " + key );
-        }
     }
 
     @Override
--- a/modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationServiceImpl.java
+++ b/modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationServiceImpl.java
@@ -122,6 +122,7 @@ public final class ApplicationServiceImpl
     @Override
     public void startApplication( final ApplicationKey key )
     {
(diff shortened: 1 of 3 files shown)


modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationServiceImpl.java
```diff
@@ -206,6 +209,7 @@ public void uninstallApplication( final ApplicationKey key )
     @Override
     public void uninstallLocalApplication( final ApplicationKey key )
     {
+        requireNotSystemApp( key, "uninstall" );
```
rymsha:
```suggestion
```
it is fully possible to uninstall local system app

modules/core/core-app/src/main/java/com/enonic/xp/core/impl/app/ApplicationServiceImpl.java
```diff
@@ -122,6 +122,7 @@ public boolean isLocalApplication( final ApplicationKey key )
     @Override
     public void startApplication( final ApplicationKey key )
     {
+        requireNotSystemApp( key, "start" );
```
rymsha:
```suggestion
```
it is fully possible to start system app

ashklianko (approved):

PR #12039, opened 2026-05-14
Fix grayscale color shift and improve scaling quality #7688

### LINKED ISSUE #7688: Non-RGB JPEG images rendered with wrong color profile
If a JPEG image that is not RGB color (e.g. 256 grays) is uploaded into Enonic XP, the color is crudely converted when rendering the image both inside Content Studio and on a page. It does not help to force JPEG format in the imageUrl() call. However, the original image is kept as-is in the repo, so a workaround for editors at the moment is to download the original image, convert it to RGB with an image editor, and re-upload. With the new RGB image, the colors are shown as they should be.

This issue appears both on Enonic XP 6 and 7.

trygve-skaug-1920x1080
64493aa1-8554-4311-b121-767b5a97385d

rymsha:
## Summary

Issue #7688 reports a visible color shift in Content Studio thumbnails of grayscale JPEGs. Diagnosing this carefully shows the root cause is **not** the destination `ColorModel` of the scale step — it is `Image.getScaledInstance(SCALE_SMOOTH)` itself, which routes through Java's color management and applies a tone curve that shifts raw grayscale samples by ~+55 levels. The same call also degenerates to nearest-neighbor on upscale, producing pixelated thumbnails when a user requests a scale larger than the source.

Measurements on the repo's `effect/grayscale.jpg`:

| Path | Raw gray sample after scale | Faithful? |
|---|---|---|
| `Image.getScaledInstance` + drawImage → sRGB target (pre-PR) | 180.7 | no |
| `Image.getScaledInstance` + drawImage → gray target (first iteration of this PR) | 180.7 | no |
| `Graphics2D.drawImage(src, 0, 0, w, h, null)` bilinear → any target | 125.2 | yes |
| Source raw gray sample mean | 125.0 | — |

So preserving the source `ColorModel` alone does not fix the shift; the shifted pixels just end up in a gray container instead of an sRGB one.

## What changed

`ImageHelper.getScaledInstance` now uses **progressive bilinear halving plus a bicubic final / upscale step**:

- While either dimension is more than 2× the target, halve via `Graphics2D` bilinear.
- Final pass at exact target dimensions via `Graphics2D` bicubic. Upscales naturally fall through to a single bicubic pass.

This is the standard "Chris Campbell perils-of-getScaledInstance" idiom. Across the fixtures:

- Grayscale source: raw sample mean preserved (125 → 125 in the new path; was 125 → 180 before).
- Downscale quality: matches area-average baseline within RMS ≈2 on adversarial high-frequency content.
- Upscale quality: smooth interpolation instead of blocky nearest-neighbor.
- Performance: comparable or faster than the old `SCALE_SMOOTH` path.

### Destination color model

The output `BufferedImage` type is now chosen by source color space:

- **Grayscale** → `TYPE_BYTE_GRAY`. Smaller files, no chroma noise, universally rendered by all browsers.
- **Alpha** → `TYPE_INT_ARGB`. Alpha preserved through scaling.
- **Everything else (sRGB, CMYK, exotic ICC profiles)** → `TYPE_INT_RGB`. CMYK in particular is flattened to sRGB because browser support for CMYK JPEGs is unreliable (Firefox in particular has long-standing issues; Chrome/Edge render approximations; only Safari is solid).

## Notes

- No new dependency. Pure JDK.
- The `processed.jpg` test reference was regenerated because the new algorithm produces different (better-quality, color-faithful) pixels. Deployed XP instances will lazily regenerate any cached thumbnails on next request — worth mentioning in release notes.
- `ImageHelper.createCompatibleImage` (added in the first iteration of this PR) was removed; it has no callers outside `ImageHelper` and is no longer needed.

## Test plan

- [x] `./gradlew test` — full suite green across 541 tasks
- [x] `ImageHelperTest.getScaledInstance_preserves_grayscale_brightness` — asserts raw gray sample mean survives scaling within ±2 (legacy path shifted +55)
- [x] `ImageHelperTest.getScaledInstance_preserves_alpha_for_argb_source`
- [x] `ImageHelperTest.getScaledInstance_upscale_interpolates_not_nearest_neighbor`
- [x] `ImageServiceImplTest.readImage_cmyk_is_flattened_to_srgb` — asserts CMYK input produces an sRGB JPEG
- [x] `ImageServiceImplTest.readImage_grayscale_jpeg_preserves_grayscale_after_scaling` — end-to-end verification through the read → scale → encode pipeline
- [ ] Manual verification: upload a grayscale JPEG and confirm the rendered Content Studio thumbnail keeps the source's brightness instead of being ~55 levels too bright

Closes #7688


anatol-sialitski (approved):

rymsha:
Addressed both Copilot review comments in c6372531af.

**Comment 1 — [r3243891535](https://github.com/enonic/xp/pull/12039#discussion_r3243891535): `chooseDestinationType` dropped alpha for grayscale+alpha sources.**

Reordered the method so the transparency check runs first. Grayscale-with-alpha images (e.g. grayscale PNGs) now route to `TYPE_INT_ARGB` and keep their transparent pixels; opaque grayscale still maps to `TYPE_BYTE_GRAY`. New test: `ImageHelperTest.getScaledInstance_preserves_alpha_for_grayscale_alpha_source` — builds a `ComponentColorModel(CS_GRAY, …, hasAlpha=true)` source with a transparent half, scales it, and asserts both `hasAlpha()` on the destination and that source-transparent pixels stay transparent. The opposite path (`removeAlphaChannel` background fill) is unchanged and remains covered by the existing `TYPE_INT_ARGB` test.

**Comment 2 — [r3243891562](https://github.com/enonic/xp/pull/12039#discussion_r3243891562): `-1` aspect-ratio sentinel regression.**

Picked **branch (b) — reject explicitly**. `getScaledInstance` now throws `IllegalArgumentException` for `width <= 0 || height <= 0` with a message that documents the API change.

Rationale from grep survey across `/home/agent/workspace/` (xp + all `app-*`, `lib-*`, `starter-*`):
- 11 call sites total.
- **No** caller passes `-1` deliberately or computes a value that could be `-1` from a known-positive input path.
- The only paths where a negative value can reach the method are HTTP query parameters on `ContentImageResource`/`ContentIconResource` (`@QueryParam("size") int`) routed through `BaseImageHelper.readIconImage` / `scaleSquare`. These are user-controllable and **would already have failed under the pre-PR implementation**: the old code immediately called `createImage(width, height, …)` → `new BufferedImage(-1, …, …)`, which itself throws. So this PR does not introduce a regression for the sentinel; it just turns the failure mode into a clearer, earlier `IllegalArgumentException` with a message instead of relying on `BufferedImage`'s internal validation.

The behavior change (rejection rather than implicit failure) is now mentioned in the PR description for release notes.

New test: `ImageHelperTest.getScaledInstance_rejects_non_positive_dimensions` — asserts the exception is thrown for `(-1, 20)`, `(20, -1)`, `(0, 20)`, and `(20, 0)` via AssertJ `assertThatThrownBy(...)`.

**Build:** `./gradlew :core:core-api:test` and `:core:core-image:test` and `:core:core-project:test` all green locally. CI run for the new commit: https://github.com/enonic/xp/actions/runs/25900634615

PR #12047, opened 2026-05-16
Add portal.getRegions() helper

rymsha:
## Summary

Adds a small convenience helper to `lib/xp/portal`:

```ts
portal.getRegions(): Region[]
```

Returns the regions of the current page or layout component as an **array**, in declaration order. Convenience wrapper around `Object.values(getComponent().regions)` that also handles the edge cases (no current component, no regions) by returning `[]` instead of throwing.

### Why

`portal.getComponent().regions` returns a *map keyed by region name*, but practically every caller wants an *array* — to render regions in declaration order, count them, iterate them. Apps work around this with the same 4-line snippet pasted everywhere:

```js
var component = portalLib.getComponent();
var regions = component.regions;
var keys = Object.keys(regions);
var result = [];
for (var i = 0; i < keys.length; i++) { result.push(regions[keys[i]]); }
```

`lib-util` exposed exactly this as `util.region.get()`. With `lib-util` being phased out across the IdeaProjects tree, the platform should provide it directly. Real-world case: [enonic/app-wireframe#578](https://github.com/enonic/app-wireframe/pull/578) had to inline the snippet in four layout files just to drop the `lib-util` dependency.

### Behavior

- On a page/layout with regions: returns each region in declaration order (`Object.keys` order from the descriptor).
- No current component (e.g. called outside a layout/part controller): returns `[]`.
- Current component is fragment/part/text (no regions): returns `[]`.

`getComponent().regions` keeps working — both paths are supported.

## Test plan

- [x] `./gradlew :lib:lib-portal:test` — all green (78 tests, including 3 new for `getRegions`)
- [x] `./gradlew :lib:lint` — clean
- [x] `./gradlew :lib:jsdoc` — `getRegions` appears in generated `module-portal.html` with the example
- [x] New WDIO/E2E coverage: N/A — this is a server-side lib helper, exercised by the Nashorn `GetCurrentComponentScriptTest`.

### DIFF
--- a/modules/lib/lib-portal/src/main/resources/lib/xp/examples/portal/getRegions.js
+++ b/modules/lib/lib-portal/src/main/resources/lib/xp/examples/portal/getRegions.js
@@ -1,0 +1,28 @@
+var portalLib = require('/lib/xp/portal');
+var assert = require('/lib/xp/testing');
+
+// BEGIN
+var regions = portalLib.getRegions();
+log.info('Number of regions = %s', regions.length);
+// END
+
+// BEGIN
+// Array of regions returned, in declaration order.
+var expected = [
+    {
+        'components': [
+            {
+                'path': '/main/0/bottom/0',
+                'type': 'part',
+                'descriptor': 'myapplication:mypart',
+                'config': {
+                    'a': '1'
+                }
+            }
+        ],
+        'name': 'bottom'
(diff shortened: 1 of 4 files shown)


rymsha:
@copilot isn't it just Object.values(component.regions) ?

PR #12085, opened 2026-05-28
CSP support #12084

### LINKED ISSUE #12084: Native CSP support
Similar to how WebResponse supports cookies XP must allow to build CSP header
- add Sha
- get nonce (lazy set internally when first read)
- add directives

The "build" is dynamic. If directives conflict during "add" they are merged with - so they become more "weakening"

There should be a "set" method, too - that overrides.

rymsha:
Fixes #12084

Adds a first-class `ContentSecurityPolicy` model (`com.enonic.xp.web.csp`) carried on `WebRequest` — one instance per servlet request, shared by reference when handlers wrap requests — plus a `lib-portal` JavaScript surface (`portal.csp()`) so site / page controllers, parts, layouts, and widgets can contribute directives, SHA hashes, and a request nonce alongside Java contributors.

The policy is treated as a fancy header: `ResponseSerializer` serializes it into the servlet response at the same place other headers are written, so late additions during rendering still land. A non-empty policy overrides a `Content-Security-Policy` header set directly on the response. A CSP header a script controller sets directly is folded into the request policy — `resetAll()` plus the header's own rules (`resetTo`) — so later contributions still apply on top and the composed value is what gets emitted. Since the enforcing and report-only headers can legitimately coexist, the report-only rules are a second, independent rule set (`csp.reportOnly()`, Java only); both share the request nonce.

## JavaScript API (`lib-portal`)

```js
const portal = require('/lib/xp/portal');
const csp = portal.csp(); // request-scoped; same instance for the whole request

// Deny-all baseline -- start restrictive, then open up:
csp.strict(); // default-src 'none'; base-uri 'none'; frame-ancestors 'none'

// Typed source-list directives -- variadic CspSource and/or raw strings
csp.scriptSrc(portal.CspSource.SELF, 'https://cdn.example.com');
csp.styleSrc(portal.CspSource.SELF);
csp.imgSrc(portal.CspSource.SELF, portal.CspSource.DATA);
// also: defaultSrc, fontSrc, connectSrc, mediaSrc, objectSrc, frameSrc, workerSrc, manifestSrc, childSrc

// Restrictive directives
csp.frameAncestors(portal.CspSource.NONE);
csp.baseUri(portal.CspSource.SELF);
csp.formAction(portal.CspSource.SELF);

// Boolean directive (add with no sources) + sandbox (unquoted tokens)
csp.add('upgrade-insecure-requests');
csp.sandbox(portal.SandboxFlag.ALLOW_SCRIPTS, portal.SandboxFlag.ALLOW_SAME_ORIGIN);

// Hash sources (script-src / style-src only) -- digest content, or a precomputed digest
csp.shaScriptSrc({content: 'window.foo = 42;'});                       // sha256
csp.shaStyleSrc({content: 'body { color: red; }', algo: 'sha384'});    // choose algorithm
csp.shaScriptSrc({hash: 'AbCdEf0123...', algo: 'sha384'});             // precomputed base64

// Nonce -- valid only on script-src / style-src; returns the value to stamp on the tag
const nonce = csp.nonceScriptSrc(); // -> script-src 'nonce-...'
csp.nonceStyleSrc();                // -> style-src  'nonce-...' (same value)

// Granular / reporting / trusted-types directives
csp.scriptSrcElem(portal.CspSource.SELF);   // also scriptSrcAttr, styleSrcElem, styleSrcAttr
csp.reportTo('csp-endpoint');               // + a Reporting-Endpoints response header
csp.requireTrustedTypesForScript();         // require-trusted-types-for 'script'
csp.trustedTypes('my-policy');

// Generic escape hatches -- variadic add (one, several, or none = boolean directive)
csp.add('report-uri', '/csp-report');       // report-uri is deprecated; reach it via add
csp.add('script-src', "'self'", 'https://cdn.example.com');
csp.override('script-src', portal.CspSource.SELF);
csp.resetAll();                             // clean slate (policy-level)
csp.resetTo("default-src 'self'");          // replace with a raw header's rules (null/'' = resetAll)
```

All directive methods are chainable. `CspSource` keyword constants are emitted single-quoted (`'self'`, `'none'`, `'unsafe-inline'`, `'unsafe-eval'`, `'strict-dynamic'`, `'unsafe-hashes'`, `'wasm-unsafe-eval'`, `'report-sample'`); scheme constants are emitted verbatim (`data:`, `blob:`). `SandboxFlag` constants are emitted unquoted (`allow-scripts`, `allow-same-origin`, ...). `TrustedTypesKeyword` constants (`ALLOW_DUPLICATES`, `NONE`, `WILDCARD`) cover the `trusted-types` magic values; policy names stay plain strings.

The report-only rule set is deliberately not exposed to JavaScript: apps must not be able to flip the response to non-enforcing with one call.

## Java API

`WebRequest.getContentSecurityPolicy()` returns the mutable, request-scoped `ContentSecurityPolicy` (same instance for the lifetime of the servlet request; request wrappers share it; no separate `Builder`). It mirrors the JS surface:

```java
ContentSecurityPolicy csp = req.getContentSecurityPolicy();

csp.strict();
csp.scriptSrc( CspSource.SELF, CspSource.STRICT_DYNAMIC );
csp.imgSrc( CspSource.SELF, CspSource.DATA );
csp.frameAncestors( CspSource.NONE );
csp.add( "upgrade-insecure-requests" );
csp.sandbox( SandboxFlag.ALLOW_SCRIPTS );
csp.shaScriptSrc( inlineScriptBytes );

String nonce = csp.nonceScriptSrc(); // or nonceStyleSrc() for style-src

csp.reportOnly().scriptSrc( CspSource.NONE ); // independent report-only rule set, same nonce

csp.add( "require-trusted-types-for", "'script'" );
csp.override( "script-src", "'self'" );
```

All directive methods return `ContentSecurityPolicy` (chainable); `build()` renders the header value.

## Preset

- `strict()` -- deny-all baseline: `base-uri 'none'; default-src 'none'; frame-ancestors 'none'`. Call it first, then open up only what you need.

## Semantics

The policy aggregates directives from independent contributors (platform, site app, custom apps, widgets, controllers) by **plain union**: each directive's source list is the union of every contributor's sources, emitted as written. The API does not arbitrate between sources -- it is a builder, not a browser.

- **Merge:** *contributor* methods are additive -- the directive methods (`scriptSrc`, `styleSrc`, ...) and `add` only union sources (deduped). *Policy-level* methods change the whole policy and are not for parts: `override(d, ...)` replaces a directive's sources (no freeze -- a later `add` still extends), and `strict()` seeds the deny-all baseline.
- **Removal:** `reset(d, ...)` removes the named directives (with no argument it removes nothing); `resetAll()` is the clean slate (the request nonce is preserved); `resetTo(headerValue)` replaces the policy with the rules parsed from a raw header value (lenient, browser-like parsing; a `null` or empty value is effectively `resetAll()` -- if nothing is added afterwards, no header is emitted). There is no per-source removal -- e.g. `reset('upgrade-insecure-requests')` is how a boolean directive is unset.
- **Validation:** `add`/`override` (and every typed directive method) reject sources containing whitespace, control characters, `;` or `,` -- tokens that would smuggle extra directives into the emitted header -- and directive names must match `[a-zA-Z][a-zA-Z0-9-]*`. So `csp.scriptSrc(userInput)` cannot be used to widen the policy.
- **Serialization:** the platform writes the policy into the servlet response next to the other headers. A non-empty rule set overrides a same-named header set directly on the response; an empty rule set emits nothing and leaves a directly-set header as is. A CSP header a script controller sets directly never travels as a plain header -- it is folded into the request policy (`resetTo`), so it overrides earlier contributions while later ones still apply on top.
- **No source resolution:** the builder emits `'unsafe-inline'` alongside a `'nonce-...'`, hash, or `'strict-dynamic'` exactly as declared and lets the **browser** apply its precedence (a nonce/hash/`'strict-dynamic'` makes the browser ignore `'unsafe-inline'`). That precedence is the secure default, so the API does not second-guess it. A contributor that genuinely needs the looser source to win -- e.g. an editor allowing inline styles over a strict `style-src` -- uses `override(d, ...)` to replace the directive, which drops the nonce/hash as a plain side-effect of replacement. Relaxing is thus an explicit, greppable act by the contributor that needs it, not emergent magic applied to everyone.
- **Redundant `'none'`:** `'none'` is the union identity (it matches nothing), so when a directive carries `'none'` plus real sources, the now-meaningless `'none'` is omitted from the output. Alone, `'none'` is emitted as-is.
- **Nonce:** lazily generated, >= 128 bits, cached for the request and shared between the enforcing and report-only rule sets. A `nonce-` source is valid only for `script-src` and `style-src`; `nonceScriptSrc()` and `nonceStyleSrc()` both return the same value and wire their target directive. If no `nonce*` method is called, no `nonce-` entry is emitted.
- **Output:** directives render in alphabetical order (deterministic); sources within a directive keep insertion order.
- **Report-only:** `csp.reportOnly()` (Java) reaches the independent rule set emitted as `Content-Security-Policy-Report-Only`; both headers can coexist (e.g. enforce a settled policy while trialling a stricter one). While left empty, no report-only header is emitted.

## Test plan

- [x] `./gradlew :web:web-api:test :web:web-impl:test :portal:portal-api:test :portal:portal-impl:test :lib:lib-portal:test` -- green.
- [x] `ContentSecurityPolicyTest` (model: directives, validation, reset/resetAll/resetTo, report-only rule set, nonce sharing), `ResponseSerializerTest` (serialization, override of direct headers, header coexistence), `ControllerScriptImpl_cspTest` (fold of controller-set headers), `CspHandlerScriptTest` (script bridge).

Generated with Claude Code

### DIFF
--- a/modules/lib/lib-portal/src/main/java/com/enonic/xp/lib/portal/csp/CspHandler.java
+++ b/modules/lib/lib-portal/src/main/java/com/enonic/xp/lib/portal/csp/CspHandler.java
@@ -110,11 +110,6 @@ public final class CspHandler
         policy().formAction( sources );
     }
 
-    public void upgradeInsecureRequests()
-    {
-        policy().upgradeInsecureRequests();
-    }
-
     public void scriptSrcElem( final String[] sources )
     {
         policy().scriptSrcElem( sources );
--- a/modules/lib/lib-portal/src/main/java/com/enonic/xp/lib/portal/csp/CspHandler.java
+++ b/modules/lib/lib-portal/src/main/java/com/enonic/xp/lib/portal/csp/CspHandler.java
@@ -28,6 +28,16 @@ public final class CspHandler
         policy().set( directive, toSources( sources ) );
     }
 
+    public void strict()
+    {
+        policy().strict();
+    }
+
+    public void strictDynamic()
+    {
+        policy().strictDynamic();
+    }
+
     public void addShaContent( final String directive, final String content )
     {
         requireNonNull( content, "content is required" );
@@ -146,9 +156,19 @@ public final class CspHandler
         policy().addStyleSrcSha( parseAlgo( algo ), base64 );
     }
 
-    public String getNonce()
+    public String nonce()
     {
         return policy().nonce();
+    }
+
+    public String nonceScriptSrc()
+    {
+        return policy().nonceScriptSrc();
+    }
+
+    public String nonceStyleSrc()
+    {
+        return policy().nonceStyleSrc();
     }
 
     public String build()
--- a/modules/lib/lib-portal/src/main/java/com/enonic/xp/lib/portal/csp/CspHandler.java
+++ b/modules/lib/lib-portal/src/main/java/com/enonic/xp/lib/portal/csp/CspHandler.java
@@ -1,0 +1,104 @@
+package com.enonic.xp.lib.portal.csp;
+
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.Locale;
+
+import com.enonic.xp.portal.PortalRequest;
+import com.enonic.xp.portal.csp.ContentSecurityPolicy;
+import com.enonic.xp.portal.csp.HashAlgo;
+import com.enonic.xp.script.ScriptValue;
+import com.enonic.xp.script.bean.BeanContext;
+import com.enonic.xp.script.bean.ScriptBean;
+
+import static java.util.Objects.requireNonNull;
+
+public final class CspHandler
+    implements ScriptBean
+{
+    private PortalRequest request;
+
+    public void add( final String directive, final ScriptValue sources )
+    {
+        policy().add( directive, toSources( sources ) );
+    }
+
+    public void set( final String directive, final ScriptValue sources )
+    {
+        policy().set( directive, toSources( sources ) );
+    }
+
+    public void addShaContent( final String directive, final String content )
+    {
+        requireNonNull( content, "content is required" );
+        policy().addSha( directive, content.getBytes( StandardCharsets.UTF_8 ) );
+    }
+
+    public void addShaDigest( final String directive, final String base64, final String algo )
+    {
+        policy().addSha( directive, parseAlgo( algo ), base64 );
+    }
+
+    public String getNonce()
+    {
+        return policy().nonce();
+    }
+
+    public void applyNonceTo( final ScriptValue directives )
+    {
+        policy().applyNonceTo( toSources( directives ) );
+    }
+
+    public String build()
+    {
+        return policy().build();
+    }
+
+    private ContentSecurityPolicy policy()
+    {
+        return this.request.getContentSecurityPolicy();
+    }
+
+    private static String[] toSources( final ScriptValue value )
+    {
+        if ( value == null )
+        {
+            return new String[0];
+        }
+        if ( value.isArray() )
+        {
+            final List<String> list = value.getArray( String.class );
+            return list.toArray( new String[0] );
+        }
+        if ( value.isValue() )
+        {
+            return new String[]{value.getValue( String.class )};
+        }
+        throw new IllegalArgumentException( "Expected a string or an array of strings" );
+    }
+
+    private static HashAlgo parseAlgo( final String algo )
+    {
+        if ( algo == null )
+        {
+            return HashAlgo.SHA256;
+        }
+        switch ( algo.toLowerCase( Locale.ROOT ) )
+        {
+            case "sha256":
+                return HashAlgo.SHA256;
+            case "sha384":
+                return HashAlgo.SHA384;
+            case "sha512":
+                return HashAlgo.SHA512;
+            default:
+                throw new IllegalArgumentException( "Unsupported hash algorithm: " + algo + ". Expected one of: sha256, sha384, sha512" );
+        }
+    }
+
+    @Override
+    public void initialize( final BeanContext context )
+    {
+        this.request = requireNonNull( context.getBinding( PortalRequest.class ).get(), "no request bound" );
+    }
+}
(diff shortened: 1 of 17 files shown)


rymsha:
Pushed `bcc73e9406` — typed CSP API on top of the existing branch (no force-push, no rebase).

**What changed**
- New `CspSource` / `SandboxFlag` enums in `portal-api` (single-quoted vs. unquoted tokens, respectively).
- `ContentSecurityPolicy` grew typed per-directive methods (`defaultSrc` / `scriptSrc` / `styleSrc` / `imgSrc` / `fontSrc` / `connectSrc` / `mediaSrc` / `objectSrc` / `frameSrc` / `workerSrc` / `manifestSrc` / `childSrc` / `frameAncestors` / `baseUri` / `formAction` / `upgradeInsecureRequests` / `sandbox`) in `CspSource...` and `String...` overload pairs, plus `addScriptSrcSha` / `addStyleSrcSha`. Generic `add` / `set` / `addSha` remain as escape hatches.
- `CspHandler` forwarders + `lib/xp/portal.ts` extensions, plus exported `CspSource` / `SandboxFlag` `as const` constants (importable as both value and type).
- `examples/portal/getCsp.js` rewritten against the typed API.
- Tests: `ContentSecurityPolicyTest` 22 → 43; `CspHandlerScriptTest` 9 → 20.

**Local tests**
- `./gradlew :portal:portal-api:test :portal:portal-impl:test :lib:lib-portal:test` — green
- `./gradlew :portal:portal-api:check :lib:lib-portal:check` — green
- `./gradlew :lib:lint` — green

**CI on `bcc73e9406`**
All required checks pass: build (11m24s), Analyze (java-kotlin / javascript-typescript / actions), CodeQL, Codacy. See https://github.com/enonic/xp/pull/12085/checks.

PR #12099, opened 2026-06-01
No way to restrict urls in installUrl #12086

### LINKED ISSUE #12086: No way to restrict urls in installUrl
app-applications has a feature to install applications from a url. this is insecure and must be prohibited without additional configuration. configuration should allow installations from specific paths (wildcard)

Additionally we should add a config to force sha512 for installUrl. by default no installation allowed without sha512 verification.

rymsha:


modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/ApplicationLoader.java
```diff
@@ -41,11 +62,6 @@ public ByteSource load( final String urlString, final String sha512Hex, final Co
 
     public ByteSource load( final URL url, final byte[] sha512Checksum, final Consumer<Event> eventConsumer )
     {
-        if ( !ALLOWED_PROTOCOLS.contains( url.getProtocol() ) )
-        {
-            throw WebException.badRequest( "Unsupported protocol: " + url.getProtocol() );
-        }
-
         try
         {
```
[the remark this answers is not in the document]
rymsha: it was a band-aid fix, this PR brings more sophisticated filtering that just protocol

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/ApplicationResourceService.java
```diff
@@ -47,6 +48,13 @@ public ApplicationResourceService( final @Reference ApplicationService applicati
         this.eventPublisher = eventPublisher;
     }
 
+    @Activate
+    @Modified
+    public void activate( final AppManagementConfig config )
+    {
+        this.applicationLoader = new ApplicationLoader( config.installUrl_allowedUrls(), config.installUrl_checksumRequired() );
```
anatol-sialitski, line 55:
```suggestion
        this.applicationLoader = new ApplicationLoader( config );
```
Maybe it is better to pass the entire config object, so it is easier to maintain in the future.

anatol-sialitski (approved):

modules/server/server-rest/src/main/java/com/enonic/xp/impl/server/rest/ApplicationResourceService.java
rymsha: we will cross that bridge when we get there. YAGNI

PR #12190, opened 2026-07-11
Annotation-based tracing with OSGi load-time weaving #12197

### LINKED ISSUE #12197: Annotation-based tracing with OSGi load-time weaving
Internal tracing wraps code in `Tracer.trace(...)` lambdas, which obscures business logic and couples every call site to the tracing API.

Replace it with a declarative model:

- `@Traced("name")` annotation on the method that is the unit of work; classes are instrumented at OSGi class-load time by a ByteBuddy(ASM) `WeavingHook` (same mechanism as the existing Nashorn weaver), so third-party apps can use the annotation too.
- `Tracer` propagates the current trace with `ScopedValue` instead of `ThreadLocal` (Java 25).
- The public `Tracer` API stays backwards compatible; exceptions propagate unchanged (tracing never alters what a caller catches).
- Trace attributes are restricted to an exporter-safe value model (String | boolean | long | double | List&lt;String&gt;) with typed writers, preparing for OpenTelemetry export (e.g. via vector.dev) while remaining compatible with app-livetrace.
- All existing internal trace points are converted, and trace delivery to listeners is covered end-to-end.

rymsha:
Closes #12197

Internal tracing no longer wraps code in `Tracer.trace(...)` lambdas. Methods are annotated with the new `@Traced` annotation (core-api) and rewritten at class-load time by a ByteBuddy(ASM)-based OSGi WeavingHook (`TraceWeaver` in launcher-impl, registered alongside `NashornWeaver`). The weaver moves each annotated method body to a synthetic method and replaces the method with a wrapper that runs the body through the ordinary public `Tracer.trace(String, TraceRunnable)` / `Tracer.trace(String, Runnable)` API via an invokedynamic-created lambda, so no auxiliary classes need to be injected into bundles. When tracing is disabled the wrapper invokes the body directly with no allocation. Weaving applies only to bundles wired to `com.enonic.xp.trace`; woven bundles get a dynamic import of the package, and third-party apps can use `@Traced` too. Classes that fail to weave load unmodified. The annotation always sits on the method that is the unit of work (worker `execute()` methods, handler `doHandle`, `sendMessage`/`onEvent`) - no hand-written wrapper methods.

`Tracer` now propagates the current trace with a `ScopedValue` instead of a `ThreadLocal` (Java 25). The public `Tracer` API is unchanged and external usages keep working; exceptions propagate unchanged (the old `RuntimeException` wrapping in `trace(Trace, TraceRunnable)` is intentionally gone - tracing never alters what a caller catches).

**Attribute model (OTel-ready).** `Trace` gains typed, chainable `attribute(String, String | long | double | boolean | List<String>)` writers plus matching `Tracer.attribute(...)` statics for single-attribute enrichment on the current trace (multi-attribute blocks and expensive values use `Tracer.withCurrent`); attributes are append/overwrite-only, null values are ignored, and the unsafe `Map` mutators (`put`/`putAll`/`remove`/`clear`) are deprecated - the `Map` inheritance itself is documented as deprecated for removal. `TraceImpl` stores attributes in a `ConcurrentHashMap` (events are dispatched to listeners while attributes are still being recorded) and normalizes legacy `put` values to the closed model. Attribute changes vs master: values that used to be domain objects (`NodeId`, `ContentPath`, query expressions, ...) are recorded as strings/lists of strings; the live `httpRequest`/`httpResponse`/`context` objects are gone (response data was already carried by `status`/`type`/`size`); the `websocket` attribute is the sub-protocol list instead of a `Map`; `node.findByQueryMulti` records one composed `["repo:branch", ...]` list under `searchTargets`. app-livetrace is unaffected: it reads only `script`/`app` (as strings) and `path`/`rawpath`/`type`/`parentId`/`websocket`-truthiness in the UI.

Trace start events are dispatched before the body records attributes; attributes are guaranteed complete at the end event (documented on `@Traced`). `TraceLocation` points at the annotated method with the original first line number preserved by the weaver. The tracing code is `@NullMarked` with explicit `@Nullable` (JSpecify).

**Tests:** core-api unit tests for the ScopedValue `Tracer` (scoping, shielding, exception transparency) and the typed/static attribute writers; server-trace tests for attribute normalization; launcher-impl weaver tests covering primitive/void/static/synchronized/generic and interface-default methods, checked-exception transparency, nesting, idempotency, wiring gate and dynamic imports; a `TestTrace` fixture (core-api test fixtures) under which every `withCurrent` enrichment lambda in the conversion executes in some suite (node/content itests, task, portal handlers/workers/scripts/websocket, web filter, admin tool); and an end-to-end test proving traces from woven classes are delivered to `TraceListener`s through the real `TraceService` and async event dispatcher, including backwards compatibility of the pre-existing `Tracer` API.

https://claude.ai/code/session_016ESKFVmv1S29HL3h2ASWxC


anatol-sialitski (approved):

rymsha:
@copilot resolve the merge conflicts in this pull request

PR #12191, opened 2026-07-12
Pass custom data from controllers to response filters #6122

### LINKED ISSUE #6122: Add javascript object as property on response object, which can be used in response filters
The way the current response object is works is very coupled with HTML and thymeleaf.
Components are not rendered inline so that what page returns is a finished HTML, but rather is contains component comments. I'm guessing the components are rendered async and put together in postProcess sorta like promises.

BTW are the order of pageContributions in the same order as the components on a page?

Working with render-js I autogenerate css and html from a javascript object.

Example render-js javascript object:
https://github.com/ComLock/render-js/blob/master/test/obj.test.es#L95-L140

Currently the css is put inside a html string as a style element and then added to pageContributions.headEnd.

Since various components use the same autogenerated classes I built a filter to try to make it unique:
- https://github.com/ComLock/enonic-xp-lib-autoprefixer/blob/master/src/main/resources/site/filters/uniqCss.es
- https://github.com/ComLock/enonic-xp-lib-autoprefixer/blob/master/src/main/resources/lib/enonic/autoprefixer/concat.es

If I could pass a javascript object to the filters, it would open a lot more possibilities and make it possible to make much smarter implementations.

This may also be relevant in cases where you want to return something else than HTML...

I have previously discovered that you cannot add your own properties on the response object:
https://discuss.enonic.com/t/passing-data-from-page-controller-to-responsefilter-via-http-response/609

Ideas, thoughts?

Documentation:
http://xp.readthedocs.io/en/stable/developer/ssjs/http-response.html
http://xp.readthedocs.io/en/stable/developer/site/filters.html

rymsha:
Fixes #6122

`contextLib.setCustomLocalAttribute(name, value)` stores a JSON-like value in the local scope of the current context. Values are serialized into `GenericValue` and stored under the `custom.` prefix; read them via `contextLib.get().attributes['custom.<name>']`. Passing `null`/`undefined` removes the attribute. Stored values are visible to everything running in the same execution — controllers, response processors and filters, tasks — including nested `contextLib.run` calls.

Groundwork and related fixes:
- `ContextAccessor` and `TaskProgressReporterContext` migrated from `ThreadLocal` to `ScopedValue`. A legacy `ThreadLocal` remains only behind `ContextAccessorSupport` to keep the test support backward compatible (same pattern as `PortalRequestAccessor`).
- Script event listeners and async main execution now run in a bound context, so ambient state stays within one dispatch.
- `ScriptRunnerSupport` no longer removes the test context before dynamic tests execute.

https://claude.ai/code/session_01PZjHraoiaNaP7vJmM6vpg6


anatol-sialitski (approved):

rymsha:
@copilot resolve the merge conflicts in this pull request

PR #12246, opened 2026-08-06
Type context attributes as JSON-like on the read side #6122

### LINKED ISSUE #6122: Add javascript object as property on response object, which can be used in response filters

rymsha:
Closes the TypeScript gap left by #6122 (`setCustomLocalAttribute`), the sibling of #12245.

## Problem

`setCustomLocalAttribute` accepts JSON-like values, and `get()` returns them in full — `ContextMapper.serializeAttributes` unwraps `GenericValue` through `toRawJs()`, so arrays and nested objects come back exactly as they went in. `context-test.js` asserts this for `['a', 'b']`.

But `Context.attributes` was typed `Record<string, number | string | boolean | Record<string, unknown>>`, which cannot hold an array:

```
✖  Property 'custom.myList' is incompatible with index signature.
     Type string[] is not assignable to type string | number | boolean | Record<string, unknown>.
       Type string[] is not assignable to type Record<string, unknown>.
```

So reading back an array attribute required a cast.

## Change

The read and write sides are genuinely different sets, so they now have different types:

- **`ContextAttributeValue`** — the JSON-like value shape. `Context.attributes` is a record of these.
- **`CustomAttributeValue`** becomes an alias of it. Same structural type, so `setCustomLocalAttribute`'s signature and any import of the name keep working.
- **`ContextAttributes`** is untouched and now documents the write side only. The asymmetry is worth stating in the type: an object passed through `run()` *is* stored on the context and reaches Java consumers, but `ContextMapper.canBeSerialized` admits only `Number`, `String`, `Boolean` and `MapSerializable`, so it never appears in `get()` output.

## One consequence to weigh

Passing `get().attributes` straight back into `run({attributes})` no longer compiles, because the read type admits arrays and the write type does not. That round trip was already lossy — an array would be dropped on the next read — so failing at compile time looks like the honest outcome. But if you would rather keep it assignable, the alternative is widening `ContextAttributes` to include arrays too, at the cost of implying `run()` can round-trip them.

## Verification

```
$ npm run test:types      # in modules/lib
✔ clean

$ npm run lint
✔ clean
```

Negative control — with the `context.ts` change stashed and the declarations rebuilt, `Context.test-d.ts` fails with exactly the error quoted above, so the new case exercises the fix rather than passing vacuously.

New `Context.test-d.ts` covers scalar, array and nested-object reads, and rejects a function and a `Date` (both of which `ContextHandlerBean.toGenericValue` throws on at runtime).

Docs for the feature are in enonic/doc-code#70.

### DIFF
--- a/modules/lib/lib-context/src/main/resources/lib/xp/context.ts
+++ b/modules/lib/lib-context/src/main/resources/lib/xp/context.ts
@@ -22,13 +22,30 @@ export interface AuthInfo {
     principals?: PrincipalKey[];
 }
 
+/**
+ * A JSON-like value: the shape context attributes have when read back with `get()`.
+ */
+export type ContextAttributeValue = string | number | boolean | ContextAttributeValue[] | {
+    [key: string]: ContextAttributeValue;
+};
+
+/**
+ * Attributes accepted by `run`. Note the asymmetry with {@link Context.attributes}: an object value is
+ * stored on the context and reaches Java consumers, but is not part of what `get()` returns - only
+ * scalars survive that round trip. Values stored with `setCustomLocalAttribute` do survive it in full.
+ */
 export type ContextAttributes = Record<string, number | string | boolean | Record<string, unknown>>;
 
 export interface Context {
     branch?: string;
     repository?: string;
     authInfo?: AuthInfo;
-    attributes: ContextAttributes;
+    /**
+     * Attributes visible in the current context: those set on it, those stored in its local scope with
+     * `setCustomLocalAttribute` (keyed `custom.<name>`), and any session attributes, merged in that
+     * order of precedence.
+     */
+    attributes: Record<string, ContextAttributeValue>;
 }
 
 export interface ContextUserParams {
@@ -60,9 +77,10 @@ interface ContextRunParams {
     setCallback<T>(fn: () => T): void;
 }
 
-export type CustomAttributeValue = string | number | boolean | CustomAttributeValue[] | {
-    [key: string]: CustomAttributeValue;
-};
+/**
+ * Value accepted by {@link setCustomLocalAttribute}.
+ */
+export type CustomAttributeValue = ContextAttributeValue;
 
 interface ContextHandler {
     get(): Context;
--- a/modules/lib/lib-context/src/main/resources/lib/xp/context.ts
+++ b/modules/lib/lib-context/src/main/resources/lib/xp/context.ts
@@ -23,18 +23,21 @@ export interface AuthInfo {
 }
 
 /**
- * A JSON-like value: the shape context attributes have when read back with `get()`.
+ * A JSON-like context attribute value.
  */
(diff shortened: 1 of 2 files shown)


rymsha:
Switched to your version — you are right, and for a sharper reason than simplicity.

The split I had drawn the line where the runtime does not. Its write type kept `Record<string, unknown>`, so it **allowed objects while forbidding arrays** — but neither survives a round trip: `ContextMapper.canBeSerialized` admits only `Number`, `String`, `Boolean` and `MapSerializable`, so an object passed to `run()` is just as absent from `get()` output as an array is. Two types, and still no protection against the surprise the split was supposed to encode.

One JSON-like type is coherent, keeps `run({attributes: get().attributes})` assignable, and is actually *stricter* than the old write type, which let anything through inside `Record<string, unknown>`. The round-trip asymmetry is a runtime rule, so it now lives in the JSDoc on `ContextAttributes` rather than in a second type.

Pushed as `7e31ee2004`: `ContextAttributes = Record<string, ContextAttributeValue>`, `Context.attributes` back to referencing it, and a tsd case carrying attributes from a read context into `ContextParams` so the reason for the single type does not get split apart again. `tsd` and `eslint` clean; negative control still fails on the array case without the widening.

One consequence worth naming: a variable typed `Record<string, unknown>` is no longer assignable as an attribute value. That is a narrowing, but the runtime cannot do anything useful with arbitrary values anyway.

doc-code#70 is updated to match.

anatol-sialitski (approved):

Issue #12216, opened 2026-07-30
Node storage commands must not read from search index


rymsha:
Some NodeService commands with pure storage semantics use the search index for reading — bulk permission-filtered child lookups and child/subtree enumeration — even though the search index is only eventually consistent. To compensate, these commands call search index refresh frequently (in `MoveNodeCommand` once per moved node), which is expensive and still fragile.

Storage commands may read the branch storage index (and real-time GETs) and of course keep writing to the search index when storage is modified, but they must never read from the search index. Search index reads should remain only in the query API (`findByQuery`, `findByParent`).

Known readers: `MoveNodeCommand`, `DuplicateNodeCommand`, `ApplyNodePermissionsCommand`, `PushNodesCommand` (children of moved nodes), `SortNodeCommand`, `ResolveInsertOrderValueCommand`, `FindNodesDependenciesCommand`.

PR #12247, opened 2026-08-07
Query by parent, deprecate findByParent #12065

### LINKED ISSUE #4772: Add support for selecting fields when doing queries
As a developer, I should be able to select fields to be returned from queries. 
E.g, I query all products with price > 100, and I will display product-name and price.

Proposed syntax:

```
var result = repo.query({
    start: 0,
    count: 2,
    query: "data.price > 100,0",
    sort: "duration DESC",
    fields: [ "data.productName", "data.price" ]
});

```

Return value suggestion:

```
{
    "total": 12902,
    "count": 2,
    "hits": [
        {
            "id": "b186d24f-ac38-42ca-a6db-1c1bda6c6c26",
            "score": 1.2300000190734863,
            "fields": [
                 "data.productName" : "Pine plywood 122X144",
                 "data.price" : 789,0
            ]
        },
        {
            "id": "350ba4a6-589c-498b-8af0-f183850e1120",
            "score": 1.399999976158142,
            "fields": [
                 "data.productName" : "Oak plywood 122X144",
                 "data.price" : 1999,0
            ]
        }
    ]
}
```

### LINKED ISSUE #12065: Deprecate content.findByParent
findContentByParent is a wrapper on top of findContentIdsByParent that "hides" expensive calls to getById for each content found
Deprecate and advise to use other methods.

on JS level it is called getChildren. and the most difficult part - query method does not provide decent way of filtering by parent.

rymsha:
Closes #12065.

## Why

`content.findByParent` is `findIdsByParent` plus a `getByIds` of every hit, so a caller who only needs ids, paths or a count pays for full content resolution behind what looks like a search. Deprecating it needed an alternative first, and the alternative was missing: `query` had no decent way to filter by parent — you had to hand-write a filter on the node-level `_parentPath` field and know that a content path gets prefixed with the content root.

Rather than add one parameter, this reshapes the surface into **two API families**, at node and content level, in Java and both libs.

## 1. Rich query — the return shape becomes a choice

Previously the shape was fixed per method: `find()` gave ids, `findPaths()` was a hard-wired one-field variant, lib-content `query()` always fetched full contents, lib-node `query()` gave ids only, and index fields were unreachable.

- **Filter by parent**: `ContentQuery.parentPath()` / `parentId()` (mutually exclusive), `NodeQuery.parent()` gains `recursive()` for descendants at any depth. Non-recursive uses the `_parentPath` term filter; recursive folds a `_path` prefix into the query expression.
- **Inherit child order — for children, not for a subtree**: a query for the *direct children* of a parent that brings no order of its own comes back in the parent's child order, the same order `findIdsByParent` used, including language-aware display-name collation. A **recursive** query does not: a child order orders siblings, so applied to a subtree it sorts levels against each other by a value only siblings can be compared on, and for a manually ordered parent `_manualordervalue` is assigned per sibling set, making the result an arbitrary interleaving rather than any arrangement an editor made. Recursive keeps the order it asks for. Explicit sort always wins, and the parent is not read at all when it is named by path and its order is not wanted — recursive, explicit sort, or size 0.
- **Pick what comes back**: `returnFields()` on both queries; values arrive per hit in the new `FieldValues` type. In JS, `content.query({returns})` takes `'contents'` (default, unchanged), `'ids'`, or a field list; `node.query({returns})` takes a field list. Every shape but the default skips reading objects entirely.
- **Parent by key in JS**: `parent` accepts the same path-or-id key `getChildren` took. An unknown parent matches nothing rather than failing, and `recursive` without `parent` is an error rather than a silently non-recursive query.

There are only two named shapes because only two behaviours differ. `'ids'` is not sugar for a field list — it is the shape that requests *no* fields, which is what lets the search skip `_source` entirely. Anything narrower than a whole content is a field list: a path is `returns: ['_path']`, already translated to a content path. An empty field list is refused rather than silently meaning `'ids'`, so each outcome has exactly one spelling and a list built at runtime cannot change the shape of the answer on the day it comes out empty.

**The two APIs treat return fields differently, because they are different kinds of API.**

| | Return fields |
|---|---|
| `ContentQuery` | `_name` `_path` `displayName` `type` `creator` `modifier` `createdTime` `modifiedTime` `owner` `language`, **enforced** — anything else is an `IllegalArgumentException` at builder time, in Java and in JS |
| `NodeQuery` | `_name` `_path` `_nodeType` `_versionKey` `_ts` are **documented and guaranteed**, but any indexed field may be requested |

The content API curates a surface over nodes, so its set is closed: the index layout stays private, and the set can widen later without breaking callers. The node API is the low-level one, whose callers already name index paths in filters and order expressions, so refusing to return a field they can already query by would be a limitation with nothing behind it. Fields beyond the documented five are index layout rather than API — their names, presence and value form may change between versions, and values arrive in indexed rather than exposed form, so a number comes back as the string the index holds. That is stated in the javadoc and the JSDoc, and asserted by an itest against real Elasticsearch.

Index keys are lowercase, but that does not surface: **a JS hit shows each field under the name it was requested by**, because the mappers walk the request rather than the answer. That also means the content set is a plain `Set` — it says which fields are allowed and publishes no naming table. A field the hit has no value for is left out rather than answered as an empty list.

## 2. Enumerator — listing is not searching

`NodeService.list(ListNodesParams)` enumerates from storage rather than from the search index:

- consistent with the latest writes — no search-refresh window, which is the same motivation as #12216
- reads no node data; entries are `{id, path, timestamp}`, ordered by path, direct children or the whole subtree
- permission-filtered per entry, with an admin short-circuit
- correspondingly limited: no filters, no paging, no choice of order, and **publish times are not evaluated**, so on `master` it lists what exists rather than what is currently published
- a read, so it refreshes nothing: writes decide when they become visible, and every write through the content API already refreshes storage, as do the security, application, virtual-app and scheduler commands whose trees the callers below list

**It stays at the node level.** Those limits make it a sharp tool that asks the caller to hold a lot of XP internals to use correctly, which is not what the content API is for — so there is no `ContentService.list` and no `content.list()` in JS. Layers sync is the one content-side caller that wants an enumeration, and it already has its own internal API for it in `LayersContentService.findAllChildren` / `findAllByParent`.

Four in-repo "list the children of a fixed system path, then getByIds" call sites moved onto it — id providers, installed apps, virtual apps, scheduled jobs — which also removes their dependency on search-index freshness.

**And four content commands that were walking the tree through search now enumerate.** #12216 stopped the *node* storage commands reading the search index, but the *content* commands above them still resolved their work-set through it first:

| | Was | Why it mattered |
|---|---|---|
| `DeleteContentCommand` | search for descendants | they get unpublished before deletion — a child written moments earlier could be missed and left published |
| `ArchiveContentCommand` | search for descendants | same unpublish path |
| `LayersContentService.findAllChildren` | search on `_parentPath` | sync would skip a content the index had not caught up with |
| `LayersContentService.findAllByParent` | search on `_path` prefix | same, for the subtree |

Sync gains twice over, since enumeration neither orders nor filters — which is all it ever wanted. Note this also means the sync walks no longer apply the master publish-window filter a search adds: syncing should mirror what exists, not only what is currently published.

Every other `find` reached from the content API stays a search, because it filters on indexed data rather than tree structure: inbound references via `_references` (dependency resolution and two sync commands), and page templates by type plus `data.supports`.

## 2b. Two listings were silently losing data

Both build a complete list out of a query that never called `size()`, so both took the default of **10** and dropped everything past the tenth hit without a trace:

| | Was | Now |
|---|---|---|
| `NodeResourceApplicationUrlResolver.findFiles` | `_path` matched against `/{app}/cms/*/*/*`, unsized — a virtual application with more than ten schemas and components resolved only some of its resources | lists the `cms` subtree and keeps the entries three levels down: no size to forget, no index to be stale, and a path order rather than an unordered scored one feeding its `LinkedHashSet`. The resolver had no test; it has one now |
| `ProjectResource.fetchSites` | `type = portal:site`, unsized — a project with more than ten sites reported ten in the REST response | `size(-1)` |

These came out of a sweep for the same shape across the repo — every `NodeQuery` / `ContentQuery` / `*Params` builder in production code that reaches `build()` without a `size()`. The rest were clean: the query translators all propagate the caller's size, the vacuum and audit commands set batch sizes, the two `SecurityServiceImpl` uniqueness checks only need to know whether more than one hit exists, and the JS handlers carry their documented `count` defaults.

## 3. Deprecations — retrofitted, not duplicated

Every deprecated method is now a call into its replacement, so the old paths are deleted rather than kept running alongside the new ones:

| Deprecated | Replacement | What that deleted |
|---|---|---|
| `NodeService.findByParent` | `findByQuery(NodeQuery)` + `parent`/`recursive`/`returnFields`; a count is a query of size 0 | `FindNodeIdsByParentCommand` — 215 lines; `findByParent` builds the `NodeQuery` it always described and hands it to `findByQuery` |
| `ContentService.findIdsByParent` | `find(ContentQuery)` + `parentPath`/`parentId`, `recursive()` | `FindContentIdsByParentCommand` down to a `ContentQuery` handed to `FindContentIdsByQueryCommand`, keeping only the parent read that gives a missing parent its `ContentNotFoundException` |
| `ContentService.findByParent` | same, then `getByIds` | — it was already `findIdsByParent` plus a fetch |
| `ContentService.findPaths` | `find(ContentQuery)` + `returnFields` naming `_path` | `FindContentPathsByQueryCommand` — 91 lines, a near-clone of the ids command with its own copy of the parent resolution and the node-to-content translation |
| `NodeQuery.withPath` | `returnFields(NodeIndexPath.PATH)` — `NodeHit.getNodePath()` still populated | |
| JS `content.getChildren` | `content.query({parent, …})` | |
| JS `node.findChildren` | `node.query({parent, …})` | |

Each retrofit keeps its test suite **unchanged and green** — 9 node tests, 27 content tests for the parent methods, 4 for `findPaths`, exception cases included — which is the evidence that the query path is a true superset of what these methods did, not a near-enough replacement.

**One deliberate behaviour change comes with the retrofit**: the old commands resolved the parent's child order whether or not `recursive` was set, so `findByParent(recursive)` sorted a subtree by an order that only orders siblings. It does not any more, for the reason given in section 1. No test pinned it — every recursive case in both the new and the deprecated suites asserts counts or set membership.

`findPaths` is worth a word: `find` already translates `_path` to a content path relative to the context root, so the two answered with the same strings for the same query. All it added was the typing, and it dropped the ids — a caller wanting both had to search twice.

That exhausts the pattern. Of the `find*` methods left, `ContentService.findNearestSiteByPath` walks ancestors testing a predicate rather than searching, and `NodeService.findVersions` / `findCommits` search `StaticStorageType.VERSION` and `COMMIT` — different data, not a return shape of `NodeQuery`.

## 4. Searches that need no fields no longer fetch `_source`

Hits are read from `_source` exclusively through the requested return fields, so a search requesting none — every ids-only query, i.e. the overwhelming majority — now runs with `setFetchSource(false)`. Elasticsearch skips loading and shipping stored source per hit instead of transferring it for the client to discard. This touches every search path in the system, hence the full-suite gate below.

## Notes for review

- **A content path means whatever the content root of the context says it means.** The same content API serves the archive, so `/mysite/foo` names a different content depending on the context. Parent paths and the `_path` return field resolve through the context root, and a parent *id* below a different root is refused rather than reached across — asserted in both directions by archive-context itests.
- **Parent resolution runs with elevated privileges.** Nothing the parent holds reaches the caller — the resolved path only builds the filter, the child order only sorts — so a caller who may read a child but not its parent gets that child either way, and one who may read neither still gets nothing. This differs from `findIdsByParent`, which throws `ContentNotFoundException` for an unreadable parent; here a missing or unreadable parent is an empty result, so path-form and id-form queries behave identically.
- **The content API borrows nothing from the node API to describe itself.** `FieldValues` lives in `com.enonic.xp.index`, keyed by `IndexPath` and saying nothing about nodes, so a content result does not import the node API to describe itself; and the content javadoc explains its return fields on its own terms, naming no node field.
- **`PageTemplateServiceImpl.getBySite`** previously inherited `findIdsByParent`'s 500-item default and silently truncated; it now requests all templates.
- **Non-recursive listing** excludes the level below the children in the query itself, so it no longer hauls a whole subtree back to discard it (raised by Copilot; the same comment also claimed descendants were permission-checked before being discarded — they were not, the direct-children filter already ran first).
- New value objects — `NodeListEntry`, `FieldValues` — are records, matching `DeleteNodeResult.Result` and `ApplyNodePermissionsResult.BranchResult`. `*Params` and `*Result` keep builders, as every sibling in the API does, and every new class is `@NullMarked` with `@Nullable` where a null is part of the contract.
- Stored values keep their original casing. Index normalization applies to field keys, matching terms and `._orderby` variants, not to values — asserted against real Elasticsearch.

## Verification

Full `itest-core` + `itest-core-content` integration suites, all unit suites of touched modules, eslint and `tsc` over `modules/lib`. `itest-core-content` is 408 tests green; `itest-core` is 694 with a single failure, `NodeImporterIntegrationTest > import_special_characters`, which reproduces identically on a clean checkout — a container locale issue, and CI passes it. The suites covering the enumeration switch — `ParentContentSynchronizerTest` (37), `ProjectContentEventListenerTest` (35), delete (10), archive (11), node list (8) — all pass. CI runs both integration suites on every push, since `check` depends on `integrationTest` in both itest modules.

## Follow-ups left open

- `RepositoryEntryServiceImpl.findRepositoryEntryIds` lists the children of a fixed system path through the search index. It is correctly sized and its writes refresh every system-repo index immediately, so there is no bug — but it is enumeration-shaped, and moving it would let that refresh go. It sits below `NodeService`, so it would have to reach for the branch-entry command directly; left alone here.
- `NodeExporter` walks a subtree through a `_path` prefix match and has the same staleness exposure, but it pages deliberately over what can be a whole repository where `list` returns everything at once. That is a memory question, not a mechanical swap.
- Typed access to **data** fields on the *content* API is intentionally not part of `returnFields` — exposing it there would have meant exposing the `._number`/`._datetime` variant scheme, and the plain field would hand back a string where the content hands back a number. The node API now answers for those fields with that caveat documented; a typed content-level equivalent deserves its own design.
- `CreateIssueCommand` picks the next issue index from a search, so a lagging index can hand two issues the same one. Same class of problem as the four commands above, but enumeration does not fix it — it needs a storage-backed counter. Untouched here.

https://claude.ai/code/session_01PT3t5hDXj91Sk6idKRQwBs


modules/core/core-content/src/main/java/com/enonic/xp/core/impl/content/ContentQueryParent.java
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * The parent a {@link ContentQuery} is restricted to, resolved into the node path used to filter children and the child order used to
+ * sort them. The child order is {@code null} where the query supplies its own order expressions and the order of the parent must not be
+ * applied.
+ */
+@NullMarked
+record ContentQueryParent( NodePath nodePath, @Nullable ChildOrder childOrder )
+{
+    static boolean isSpecifiedIn( final ContentQuery query )
```
anatol-sialitski: It’s not entirely clear from the method name what it does.

rymsha: Agreed — as a static on `ContentQueryParent` it read as a claim about the type rather than about the query, and no rename really fixed that. It had one caller and the body needs no name, so I inlined it (`c114153`):

```java
if ( this.query.getParentPath() != null || this.query.getParentId() != null )
```

`resolve` now states the precondition in its javadoc instead.

---

modules/core/core-api/src/main/java/com/enonic/xp/node/NodeQuery.java
```diff
@@ -1,60 +1,175 @@
...
+         * <li>{@link NodeIndexPath#NAME} - {@code _name}</li>
+         * <li>{@link NodeIndexPath#PATH} - {@code _path}</li>
+         * <li>{@link NodeIndexPath#NODE_TYPE} - {@code _nodeType}</li>
+         * <li>{@link NodeIndexPath#VERSION} - {@code _versionKey}</li>
+         * <li>{@link NodeIndexPath#TIMESTAMP} - {@code _ts}, in ISO-8601 form</li>
+         * </ul>
+         * Any other indexed field may also be requested, including data fields and the typed variants the index maintains for its own
+         * purposes, such as {@code ._number} and {@code ._orderby}. Those are the layout of the index rather than API: their names,
+         * presence and value form are not guaranteed to survive an upgrade, and the value returned is the indexed one, which for a
+         * number or a date is not the form the node exposes.
```
rymsha: simplify to short sentence. Other fields are not supported

rymsha: Done in `96ecb7b` — the paragraph is now one sentence, "Other fields are not supported.", in the javadoc and in the lib-node JSDoc for `returns`.

The builder still doesn't reject them, so I also renamed the itest covering that from `fields_beyond_the_documented_ones` to `unsupported_field_is_not_rejected`, so it reads as the absence of a rejection rather than as a capability.

---

anatol-sialitski (approved):

PR #12282, opened 2026-08-20
Benchmark the whole-tree node operations

rymsha:
Adds `NodeTreeOperationsBenchmark` to `ptest-content`: `node.move`, `node.duplicate`, `node.applyPermissions` and `node.delete`, each as a single shot over the simplest big shape — one parent with 10 000 direct children, one branch, no cross-branch work.

The same benchmark (byte-identical files) sits on #12281's branch, so a Performance Test workflow run on `master` and one on `claude/xp-issue-12065-alternative-pur3l4` measure the difference the walker rework makes on these four operations.

## What is measured

Time alone would not show what the rework did, so each operation is measured three ways:

| Benchmark | Mode | Score | Unit | Throughput | Time/node | Alloc/op | Alloc/node | Peak live/op |
|---|---|---:|---|---:|---:|---:|---:|---:|
| NodeTreeOperationsBenchmark.applyPermissions | ss | 2436.01 | ms/op | 821 nodes/s | 1217.4 us | 235.2 MiB | 120.4 KiB | 7.2 MiB |

(one local run at `-Dptest.tree.children=2000`, to show the shape of the report — not a result)

- **Throughput in nodes, not operations.** One operation rewrites the whole tree, so its reciprocal says nothing: it is one operation however many nodes it covered. Each measurement reports the nodes it touched, and the report divides by that.
- **Allocation** is churn — every node an operation rewrites is read, rebuilt and re-indexed, and that traffic is the same however the walk is organized. Expect this column to match between the two runs.
- **Peak live** is what the operation *holds*: how far the surviving heap grows while it runs. This is the column the rework should move, since it replaced whole-subtree collections with bounded strides.

Both memory figures are measured close to the operation rather than through JMH's GC profiler: the embedded search server allocates on its own threads throughout, and the corpus a benchmark rebuilds between invocations allocates too. The whole-process figure came out at ~800 MiB for every operation alike, drowning out the differences between them. Allocation is read from the counter of the thread that runs the operation; the live set is sampled from the heap pools' post-collection usage while it runs.

**On reading the counters:** JMH sums an auxiliary event counter over the iterations of a run rather than averaging it, so every counter here is a total. Each measurement therefore counts its own operations, and the report divides by that — without it, a five-iteration run reads as one operation that allocated five times as much and covered five times as many nodes.

## Design notes

- **Single shot, milliseconds.** Every operation is one full walk of the tree, so `Mode.SingleShotTime` with 1 warmup and 5 measurement iterations (3 for the corpus-consuming pair).
- **Only the operation is timed.** The index is settled with an untimed refresh before every invocation, so a measurement never pays for the ground the previous one left behind.
- **Move** alternates the tree parent between two targets — every child follows on each shot.
- **Duplicate** copies the whole tree; each invocation leaves its copy behind, as production would.
- **applyPermissions** alternates between two permission lists over the whole tree (`scope=TREE`), so every invocation writes a real change; the operating role keeps full access in both variants.
- **Delete** consumes its tree, so each invocation gets a fresh one, built and settled outside the measurement.
- The child count bends to `-Dptest.tree.children`, and a single benchmark can be run with `-Pptest.include=...`, for quick local runs.

https://claude.ai/code/session_01PT3t5hDXj91Sk6idKRQwBs

### DIFF
--- a/modules/ptest/ptest-content/build.gradle
+++ b/modules/ptest/ptest-content/build.gradle
@@ -67,6 +67,16 @@ tasks.register( 'performanceTest', JavaExec ) {
 
     systemProperty 'java.awt.headless', 'true'
 
+    // -Pptest.tree.children=<n> sizes the tree the whole-tree node benchmarks operate on; -Pptest.include=<regex> narrows the run.
+    if ( project.hasProperty( 'ptest.tree.children' ) )
+    {
+        systemProperty 'ptest.tree.children', project.property( 'ptest.tree.children' )
+    }
+    if ( project.hasProperty( 'ptest.include' ) )
+    {
+        args project.property( 'ptest.include' )
+    }
+
     workingDir = layout.buildDirectory.dir( 'perftest-report' ).get().asFile
     doFirst {
         workingDir.mkdirs()
--- a/modules/ptest/ptest-content/src/main/java/com/enonic/xp/perftest/content/BenchmarkMain.java
+++ b/modules/ptest/ptest-content/src/main/java/com/enonic/xp/perftest/content/BenchmarkMain.java
@@ -10,7 +10,6 @@ public final class BenchmarkMain
     public static void main( final String[] args )
         throws Exception
     {
-        // an argument narrows the run to the benchmarks whose name matches it, for measuring one operation at a time
         final String include =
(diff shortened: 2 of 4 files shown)


rymsha:
included in #12281

PR #12305, opened 2026-08-29
app.config conversion and read-only app on GraalJS #12304 #12306

### LINKED ISSUE #12304: app.config is a host Java map on GraalJS
On GraalJS, `app.config` reaches scripts as a `java.util.HashMap` instead of a JavaScript object. Direct reads still work — `config['key']` returns the value, and `for...in` walks the real keys — but `Object.keys`, `in` and `hasOwnProperty` resolve against the Java object's members, which are its methods: `get`, `put`, `size`. The working set and the broken set never intersect, so the ordinary guarded loop over config returns nothing and throws nothing. In production this silently emptied a config-driven allowlist.

The cause is one line. Both engines build `app` from `ApplicationInfoBuilder.buildMap`, which uses the supplier it is handed for the outer map *and* for the nested `config` map. Nashorn passes `javascriptHelper::newJsObject`, whose `Bindings` return type happens to be a `Map`, so both levels become real JS objects. Graal's `newJsObject` returns `Value`, which is not a `Map`, so the same call does not compile there — the Graal executor built a plain `HashMap` and wrapped it in `ProxyObject.fromMap`, which wraps only the outer map and does not recurse. That is why `app` behaves correctly and `app.config` does not.

The fix routes the same map through `GraalObjectConverter.toJs`, which already recurses into nested maps and already backs callback-result conversion. One line and a dropped import. Nashorn is untouched.

## Measured

With `app.config` holding `idProvider.name = enonic-cloud`:

| idiom | Nashorn | GraalJS |
| --- | --- | --- |
| `config['idProvider.name']` | `"enonic-cloud"` | `"enonic-cloud"` |
| `for (k in config)` | the config keys | the config keys |
| `Object.keys(config)` | the config keys | `clear,clone,compute,containsKey,get,put,size,…` |
| `hasOwnProperty.call(config, 'idProvider.name')` | `true` | `false` |
| `'idProvider.name' in config` | `true` | `false` |

The two rows that work and the three that do not never intersect, so the standard defensive loop finds nothing:

```js
for (const key in config) {
    if (Object.prototype.hasOwnProperty.call(config, key)) { /* never reached */ }
}
```

## Test

`application-test.js` gets a non-empty `app.config` and pins `Object.keys`, `hasOwnProperty`, `in`, bracket access, `JSON.stringify`, and the `for...in` + `hasOwnProperty` guard. `gradle/js-tests.gradle` already runs that file on both engines, so the assertions hold Nashorn and GraalJS to the same contract.

## Scope

Not covered here, and not a defect: values returned from Java beans still need `__.toNativeObject`, the documented conversion API, used 1134 times across XP's own libraries. Nashorn's `in` was lenient on unconverted host objects and GraalJS's is not, so bean reads that skipped the conversion worked by accident before. That is application code, not XP.

## Verified

`:script:script-impl:test` and `:script:script-impl:testGraalJs` both pass, as does `testGraalJs` for portal-impl, core-task, app-system, tools:testing and all 25 `lib-*` modules. No XP instance was started — verification is unit-level.

The binding has had this shape since `9f5fcec533` (7.11.0), so any release where an application selects GraalJS is affected.

### LINKED ISSUE #12306: app and app.config are mutable on GraalJS
The `app` object is documented as application information and nothing is meant to write to it, but nothing stops a script from doing so. `app.name`, `app.version` and `app.config` are all writable on both engines, and on GraalJS the global binding itself can be replaced.

That matters more on GraalJS than it looks, because contexts are pooled. A `ContextSlot` holds one `Context` for the application's lifetime and `app` is bound once per slot, so a write from one request is visible to every later request served by that slot — and only that slot, so the same request produces different results depending on which context the pool hands out.

## Measured

Both engines run strict mode (`NashornHelper` passes `-strict`, `GraalJSContextFactory` sets `js.strict=true`).

| idiom | Nashorn | GraalJS |
| --- | --- | --- |
| `app = 5` | `ReferenceError: "app" is not defined` | succeeds; the next module in the same context sees `typeof app === 'number'` |
| `app.name = 'x'` | succeeds | succeeds |
| `app.config['key'] = 'x'` | succeeds | succeeds |
| `app.config['newKey'] = 'x'` | succeeds | succeeds |
| `delete app.config['key']` | succeeds | succeeds |

Nashorn already rejects the reassignment, and by accident rather than design: its `app` lives in `GLOBAL_SCOPE` bindings, which are readable but are not an assignable binding, so strict mode has nothing to assign to. GraalJS binds through `context.getBindings("js").putMember(...)`, which creates a `writable: true, configurable: true` property of `globalThis` — the polyglot API has no way to express property attributes, so the property cannot be declared read-only from Java at all.

## The error messages are their own problem

Making the properties non-writable is not enough on its own, because GraalJS serialises the target object into the `TypeError` it raises. With a two-key configuration:

```
app = 1                     -> 1783 characters, the entire global object: Object, Function, Math, JSON, ...
app.name = 'x'              -> 139 characters, inlining name, version and every config value
app.config['key'] = 'x'     -> Cannot assign to read only property 'key' of {key: "value", dotted.key: "dotted value"}
```

Application `.cfg` files hold passwords and API keys. An ordinary application bug would therefore write the deployment's secrets into the log, an error page, or an exception report. Any fix has to control the message, not only block the write.

One further detail worth recording: `js.strict = true` does not reach `Function`-constructed code. Under a plain non-writable property, `Function("app = 1")()` — reachable through bundled third-party code — fails silently instead of throwing.

## Scope

**GraalJS only.** Nashorn is being removed and is not worth changing; it already refuses the reassignment, and its remaining looseness disappears with the engine. The TypeScript declarations express the contract for both, which is what steers application code toward what survives the removal.

**Breaking on GraalJS.** An application that writes to `app` or `app.config` currently succeeds and will start throwing a `TypeError`. GraalJS is opt-in per application and was not encouraged before 8.1, which is still in beta.

**The test harness owes a replacement.** `ScriptTestSupport` hardcodes an empty application config with no way to supply one, so an application author testing configuration-dependent code could only seed it by writing to `app.config` from the test script. That workaround stops working, so the harness needs a supported hook in the same change.

## Test

`application-test.js` runs under both engines and pins that reassigning `app` throws on either. The GraalJS-only half — `Object.isFrozen`, the exact error messages, and that writing a key, adding a key, deleting a key and redefining a property each throw — sits behind the `typeof Graal !== 'undefined'` marker that `engine-test.js` already establishes, since the contract deliberately holds on one engine.

The types get a `tsd` test in `modules/lib/test`, pinning that writes to `app.name`, `app.version`, `app.config` and `app.config['key']` are compile errors.

edloidas:
Two defects on the same GraalJS binding, one commit each.

`ProxyObject.fromMap` wraps only the map it is given, so the nested `config` map reached scripts as a host `java.util.HashMap` — `app` behaved as a JavaScript object and `app.config` did not. `GraalObjectConverter.toJs` already recurses into nested maps and already backs callback-result conversion, so routing the same map through it introduces no new conversion logic.

`app` was also writable, at both levels. `putMember` cannot express property attributes and the object behind it was never sealed, so a stray write stayed visible to every later request on the same pooled context. Nashorn already rejects `app = ...` because its binding lives in `GLOBAL_SCOPE` and is not assignable; making the GraalJS binding read-only closes that divergence rather than creating one.

## Changes

- `GraalScriptExecutor` binds `app` through `objectConverter().toJs(...)` instead of `ProxyObject.fromMap(...)`
- `GraalScriptExecutor` freezes `app` and `app.config`, wraps both in a `Proxy` whose `set`, `defineProperty` and `deleteProperty` traps throw, and redefines the global as a non-configurable accessor whose setter throws
- `App` declares `name`, `version` and `config` `readonly`, with `config` typed as a new exported `AppConfig`; the jsdoc `app` namespace gains the missing `config` member and `@readonly` on all three
- `ScriptTestSupport` gains `createAppConfig()`, replacing the `app.config` write that test scripts used to seed configuration
- `application-test.js` pins the config contract on both engines and the immutability contract on GraalJS; `App.test-d.ts` pins the type contract; `AppConfigScriptTest` pins the harness hook

Nashorn is untouched.

## Error messages

GraalJS serialises the target object into its own `TypeError`, and `.cfg` files hold credentials. Measured with a two-key config:

| write | before | after |
| --- | --- | --- |
| `app = 1` | 1783 chars, the entire global object | `Cannot assign to read only property 'app' of global` |
| `app.name = 'x'` | 139 chars, inlining every config value | `Cannot assign to read only property 'name' of app` |
| `app.config['key'] = 'x'` | `... of {key: "value", dotted.key: "dotted value"}` | `Cannot assign to read only property 'key' of app.config` |

The global is an accessor rather than a non-writable data property because `js.strict` does not reach `Function`-constructed code: under a data property, `Function("app = 1")()` fails silently. Cost is about 5ns per read of `app` (20M reads: 1047.6ms against 947.2ms).

## Breaking

On GraalJS, applications calling Java `Map` methods on `app.config` — `config.get('x')`, `config.containsKey('x')` — now get a `TypeError`, as do applications writing to `app` or `app.config`. GraalJS is opt-in per application and was not encouraged before 8.1, which is still in beta.

## Verified

`:script:script-impl:test` and `:testGraalJs` (180 each), `:tools:testing:test` and `:testGraalJs`, `testGraalJs` for portal-impl, core-task, app-system and all 25 `lib-*`, `:lib:typescript`, `:lib:jsdoc` and `tsd`. Each new assertion was mutation-checked by removing the code it guards and confirming it fails. Verification is unit-level; no XP instance was started.

Closes #12304
Closes #12306

### DIFF
--- a/modules/lib/core/index.d.ts
+++ b/modules/lib/core/index.d.ts
@@ -16,19 +16,21 @@ declare global {
     type XpMixin = Record<string, Record<string, Record<string, unknown>>>;
 }
 
+export type AppConfig = Readonly<Record<string, string | undefined>>;
+
 export interface App {
     /**
      * The name of the application.
      *
      * @type string
      */
-    name: string;
+    readonly name: string;
     /**
      * Version of the application.
      *
      * @type string
      */
-    version: string;
+    readonly version: string;
     /**
      * Values from the application’s configuration file.
      * This can be set using $XP_HOME/config/<app.name>.cfg.
@@ -36,7 +38,7 @@ export interface App {
      *
      * @type Object
      */
-    config: Record<string, string | undefined>;
+    readonly config: AppConfig;
 }
 
 export interface DoubleUnderscore {
--- a/modules/lib/jsdoc/global.js
+++ b/modules/lib/jsdoc/global.js
@@ -54,6 +54,7 @@ var app = {
      * Name of application.
      *
      * @type string
+     * @readonly
      */
     name: '',
 
@@ -61,6 +62,7 @@ var app = {
      * Application version.
      *
      * @type string
+     * @readonly
      */
     version: '',
 
@@ -69,7 +71,9 @@ var app = {
      * $XP_HOME/config/<app-name>.cfg.
      *
      * @type Object
+     * @readonly
      */
+    config: {}
 };
 
 /**
--- a/modules/lib/test/App.test-d.ts
+++ b/modules/lib/test/App.test-d.ts
@@ -1,0 +1,21 @@
+import type {App} from '../core/index';
+
+import {
+    expectError,
+    expectType,
+} from 'tsd';
+
+declare global {
+    // Ignore this error in code editor, it's needed when running the tests.
+    const app: App;
+}
+
+// The reads come first: assigning to app.config['someKey'] below narrows it to string.
+expectType<string>(app.name);
+expectType<string>(app.version);
+expectType<string | undefined>(app.config['someKey']);
+
+expectError(app.name = 'com.enonic.app.name');
+expectError(app.version = '1.0.0');
+expectError(app.config = {});
+expectError(app.config['someKey'] = 'someValue');
--- a/modules/script/script-impl/src/main/java/com/enonic/xp/script/graal/executor/GraalScriptExecutor.java
+++ b/modules/script/script-impl/src/main/java/com/enonic/xp/script/graal/executor/GraalScriptExecutor.java
@@ -59,6 +59,26 @@ public class GraalScriptExecutor
 
     private static final long SLOT_WAIT_SECONDS = 300;
 
+    private static final Source READ_ONLY_APP_SCRIPT = Source.newBuilder( "js", "(function() {" +
+        "const readOnly = function( target, label ) {" +
+        "const deny = function( property ) {" +
+        "throw new TypeError( \"Cannot assign to read only property '\" + String( property ) + \"' of \" + label );" +
+        "};" +
+        "return new Proxy( Object.freeze( target ), {" +
+        "set: function( target, property ) { deny( property ); }," +
+        "defineProperty: function( target, property ) { deny( property ); }," +
+        "deleteProperty: function( target, property ) { deny( property ); } } );" +
+        "};" +
+        "const source = globalThis.app;" +
+        "source.config = readOnly( source.config, 'app.config' );" +
+        "const app = readOnly( source, 'app' );" +
+        "Object.defineProperty( globalThis, 'app', {" +
+        "get: function() { return app; }," +
+        "set: function() { throw new TypeError( \"Cannot assign to read only property 'app' of global\" ); }," +
+        "enumerable: true," +
+        "configurable: false } );" +
+        "})();", "read-only-app" ).buildLiteral();
+
     private final ScriptSettings scriptSettings;
 
     private final ClassLoader classLoader;
@@ -872,7 +892,7 @@ public class GraalScriptExecutor
             final Map<String, Object> globalVariables = new HashMap<>( scriptSettings.getGlobalVariables() );
             globalVariables.put( "app", this.javascriptHelper.objectConverter().toJs( application.buildMap( HashMap::new ) ) );
             globalVariables.forEach( ( key, value ) -> this.context.getBindings( "js" ).putMember( key, value ) );
-            this.context.eval( "js", "Object.defineProperty( globalThis, 'app', { writable: false, configurable: false } );" );
+            this.context.eval( READ_ONLY_APP_SCRIPT );
         }
     }
 }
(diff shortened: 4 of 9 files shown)


rymsha (changes requested):
Generally good. But the bug is a bit wider
1. Need to deep freeze app object
2. Need to declare fool app object as read-only (don't modify Nashorn branch - declaration is enough and shows the intention)

edloidas:
Agreed on the second part. Implemented it in a new commit with tests. GraalJS only, Nashorn is not touched.

The deep freeze would be nice, but it has a larger blast radius and I don't think it belongs in this task. Mutability of `app` and `app.config` predates this PR and is identical on both engines today, so freezing GraalJS alone would make the two diverge. But if we doing it for both, it would still require the App type to become `readonly`, and there is one `.d.ts` for both engines, so it would declare readonly for Nashorn as well. It may break some apps, which is fine since it is incorrect behaviour anyway, but it should be its own issue.

rymsha (approved):

Issue #12297, opened 2026-08-26
Add `type` property to `enonic.yaml`


anatol-sialitski:
Support `Static` and `Bundle` values for the `type` property, with `Bundle` as the default.

PR #12313, opened 2026-09-01
Persisted schema Feature

anatol-sialitski:
Add `type` property to `enonic.yaml` #12297

### DIFF
--- a/modules/core/core-jsonschema/src/main/schema/templates/application.schema.json
+++ b/modules/core/core-jsonschema/src/main/schema/templates/application.schema.json
@@ -12,6 +12,15 @@
       "type": "string",
       "const": "Application"
     },
+    "type": {
+      "type": "string",
+      "enum": [
+        "Static",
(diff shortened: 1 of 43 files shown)


modules/core/core-jsonschema/src/main/schema/templates/application.schema.json
```diff
@@ -12,6 +12,12 @@
       "type": "string",
       "const": "Application"
     },
+    "name": {
+      "type": "string",
+      "pattern": "^\\w+(?:\\.\\w+)*$",
+      "maxLength": 63,
+      "description": "Application name, e.g. com.example.myapp. Optional for gradle-built applications, where it must match the Bundle-SymbolicName of the jar. Required for applications without a build (schema only), packaged by the Enonic CLI."
```
rymsha: Gradle-built is implementation detail. Don't mention it
Btw Gradle plugin should verify that values match

PR #12336, opened 2026-09-05
Reject cross-origin state-changing universal API calls from session users

rymsha:
## Summary

Security audit finding L9. State-changing calls to a universal API relied on the session cookie's `SameSite=Lax` attribute alone. A single configuration change to `None` or an empty value made every mounted `POST`, `PUT`, `PATCH` or `DELETE` API forgeable from another site, and with an API mounted on the management connector that reaches management operations.

## Changes

`SlashApiHandler` now checks the browser-sent `Origin` header before dispatching a non-safe method, when the request is authenticated through a session. The origin's scheme, host and port must equal the request's own, with default ports normalised and case-insensitive comparison. A mismatch, an opaque `null` origin or a malformed value is answered with 403 `Origin [...] is not allowed`.

Untouched, so non-browser clients and existing integrations keep working:

- `GET`, `HEAD`, `OPTIONS` and `TRACE`
- requests without an `Origin` header
- anonymous requests
- authenticated requests that carry no session, such as token or Basic authentication through an id provider that logs in with `session: false`

The comparison mirrors the same-origin check the WebSocket upgrade already applies in `web-jetty`, kept local to `portal-impl` as `SameOriginCheck`.

## Deployment notes

- The request side of the comparison is the scheme, host and port Jetty reports for the request. Behind a TLS-terminating proxy those come from the `Forwarded` or `X-Forwarded-*` headers, exactly as they do for every absolute URL and redirect XP already generates. A proxy that does not pass the protocol would make browser calls fail here, but it already breaks redirects and absolute URLs on such an installation.
- An application that deliberately serves a credentialed cross-origin API (CORS with `Access-Control-Allow-Credentials`) to session users on another origin will now receive 403 for its state-changing calls. Such an API has to use a token instead of the session cookie.

## Tests

- `SlashApiHandlerTest`: cross-origin `POST` from a session user is rejected; same origin, safe method, no session, no `Origin` header, no raw request and anonymous variants pass.
- `SameOriginCheckTest`: default ports, case, subdomains, scheme and port mismatches, origins without a host, requests without scheme or host, schemes without a default port, opaque and malformed origins.

https://claude.ai/code/session_01NF5h4qbEHLSwEBTxQKSoMV

### DIFF
--- a/modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/SameOriginCheck.java
+++ b/modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/SameOriginCheck.java
@@ -1,0 +1,55 @@
+package com.enonic.xp.portal.impl.handler;
+
+import java.net.URI;
+import java.util.Locale;
+
+import com.enonic.xp.web.WebRequest;
+
+final class SameOriginCheck
+{
+    private SameOriginCheck()
+    {
+    }
+
+    static boolean isSameOrigin( final String origin, final WebRequest request )
+    {
+        if ( origin == null )
+        {
+            return true;
+        }
+
+        final URI originUri;
+        try
+        {
+            originUri = URI.create( origin );
+        }
+        catch ( IllegalArgumentException e )
+        {
+            return false;
+        }
+
+        final String scheme = originUri.getScheme();
+        final String host = originUri.getHost();
+        if ( scheme == null || host == null || request.getScheme() == null || request.getHost() == null )
+        {
+            return false;
+        }
+
+        return scheme.equalsIgnoreCase( request.getScheme() ) && host.equalsIgnoreCase( request.getHost() ) &&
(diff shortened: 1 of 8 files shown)


rymsha:
@copilot cross origin prohibition is not global. The idea is that api developers decide.
What should be done instead - all `server:` apis must prevent cross-origin 

Method name verifySameOrigin should be called verifyOrigin - as it allows missing value (no browser) and it allows safe methods.

rymsha:
@copilot read my comment again

rymsha:
@copilot the build fails after recent changes

anatol-sialitski (approved):

rymsha:
Closing without merging, by decision of the maintainer.

A global origin check in the universal API dispatcher is the wrong layer for this. CSRF protection for state-changing calls is the responsibility of the individual API, and it only matters for management APIs. The session cookie's `SameSite=Lax` default already provides a solid baseline for everything else, so a second, platform-wide check adds friction without a matching gain.

---

